PR #317, opened 2026-03-19
fix(scripts): support OSMO control-plane deploys with in-cluster Redis

### LINKED ISSUE #316: fix(scripts): OSMO control-plane deploy fails when using in-cluster Redis without managed Redis
### Component

OSMO Control Plane (Helm/Kubernetes)

### Bug Description

Deploying the OSMO control plane with `--use-incluster-redis` fails when the infrastructure was provisioned with managed Redis disabled. The failure is not caused by Terraform provisioning itself. The setup flow assumes external-Redis and workload-identity secret behavior even when the deployment explicitly requests in-cluster Redis.

During repro, the setup path failed across multiple stages:

- the shared helper resolved the SecretProviderClass manifest from an invalid path
- the workload-identity SecretProviderClass still requested the Key Vault secret for managed Redis
- the Helm chart expected `redis-secret` even for the in-cluster Redis path
- the chart rendered an empty Redis `storageClassName`, leaving the PVC unbound on the target cluster
- the post-deploy dev-user bootstrap step failed even though `osmo login` itself worked

### Steps to Reproduce

1. Provision infrastructure with a sanitized configuration equivalent to the snippet below, where PostgreSQL is enabled and managed Redis is disabled.
2. Run the setup deploy for the OSMO control plane with the in-cluster Redis option.
3. Observe the setup failure and the control-plane pods not reaching a healthy state.

Commands Used:

```bash
cd infrastructure/terraform
terraform init
terraform apply -var-file=<sanitized>.tfvars

cd ../setup
./03-deploy-osmo-control-plane.sh --use-incluster-redis
```

### Expected Behavior

The deploy script should complete successfully when Terraform does not provision managed Redis and the operator explicitly requests `--use-incluster-redis`. The control-plane should install with in-cluster Redis, all pods should become `Running`, and the post-deploy configuration step should not fail on optional dev-user management behavior.

### Environment

_No response_

### Logs / Error Output

```shell
/workspace/.../infrastructure/setup/manifests/aks-secret-provider-class.yaml: No such file or directory
error: no objects passed to apply

failed to get objectType:secret, objectName:redis-primary-key
RESPONSE 404: 404 Not Found
ERROR CODE: SecretNotFound

Error: secret "redis-secret" not found

FailedBinding: no persistent volumes available for this claim and no storage class is set

Server responded with status code 404
Error message: {"detail":"Not Found"}


Relevant Configuration:

Sanitized `terraform.tfvars` excerpt used for reproduction:


environment     = "<env>"
location        = "<azure-region>"
resource_prefix = "<resource-prefix>"
instance        = "<instance>"

should_create_resource_group = true

system_node_pool_vm_size    = "<system-node-vm-size>"
system_node_pool_node_count = 1

node_pools = {
  gpu = {
    vm_size                 = "<gpu-vm-size>"
    subnet_address_prefixes = ["<gpu-subnet-cidr>"]
    node_taints             = ["nvidia.com/gpu:NoSchedule", "kubernetes.azure.com/scalesetpriority=spot:NoSchedule"]
    gpu_driver              = "Install"
    node_labels = {
      "kubernetes.azure.com/scalesetpriority" = "spot"
    }
    priority                   = "Spot"
    should_enable_auto_scaling = true
    enable_auto_scaling        = true
    min_count                  = 0
    max_count                  = 1
    zones                      = []
    eviction_policy            = "Delete"
  }
}

should_deploy_postgresql = true
should_deploy_redis      = false

should_enable_private_endpoint    = true
should_enable_private_aks_cluster = true

should_enable_public_network_access     = false
should_add_current_user_key_vault_admin = true
should_enable_microsoft_defender        = true

should_deploy_grafana = false

osmo_config = {
  should_enable_identity   = true
  should_federate_identity = true
  control_plane_namespace  = "osmo-control-plane"
  operator_namespace       = "osmo-operator"
  workflows_namespace      = "osmo-workflows"
}
```

### Checklist

- [x] I have searched existing issues for duplicates
- [x] I have read the [documentation](README.md)

fbeltrao:
## Description

Fix the OSMO control-plane setup path so `./03-deploy-osmo-control-plane.sh --use-incluster-redis` works when Terraform is applied with managed Redis disabled.

This change:

- fixes the shared manifest path resolution for the Key Vault SecretProviderClass helper
- splits SecretProviderClass rendering so managed-Redis secrets are only requested when the deployment is actually using external Redis
- creates the `redis-secret` object expected by the `osmo/service` chart in the in-cluster Redis path
- detects a default Kubernetes `StorageClass` and passes it to the chart so the Redis PVC binds instead of staying pending
- disables Redis TLS for the in-cluster Redis chart deployment to match the chart’s bundled Redis container
- makes the dev-user bootstrap step best-effort when the OSMO user-management endpoint returns `404`, while still requiring a successful `osmo login` and applying the service configuration

This PR is intentionally scoped to deployment automation and setup manifests. It does not change Terraform resource provisioning or the upstream OSMO Helm chart.

Validation performed:

- `bash -n infrastructure/setup/03-deploy-osmo-control-plane.sh shared/lib/common.sh`
- rendered `osmo/service` locally with sanitized values to verify `services.redis.storageClassName`, `services.redis.tlsEnabled`, and `redis-secret` handling
- re-ran the sanitized deployment flow:

```bash
cd infrastructure/terraform
terraform init
terraform apply -var-file=<sanitized>.tfvars

cd ../setup
./03-deploy-osmo-control-plane.sh --use-incluster-redis
kubectl get pods -n osmo-control-plane
helm list -n osmo-control-plane
```

- confirmed `service`, `router`, and `ui` Helm releases reached `deployed`
- confirmed `osmo-agent`, `osmo-delayed-job-monitor`, `osmo-logger`, `osmo-router`, `osmo-service`, `osmo-ui`, `osmo-worker`, and `redis` were `Running`

Side-effect review:

- `shared/lib/common.sh` is shared through symlinks into both `infrastructure/setup` and `data-management/setup`
- the new SecretProviderClass argument only affects the OSMO control-plane deploy caller
- `osmo_login_and_setup` is also used by the OSMO backend deploy script; the change only converts optional dev-user setup into a warning when the endpoint is unavailable
- the uninstall script already cleans up `redis-secret`, so the new in-cluster secret does not add cleanup drift

Closes #316 

## Type Of Change

- [ ] 📚 Documentation update
- [x] ♻️ Refactoring (no functional changes outside the deploy fix)

## Component(s) Affected

- [ ] `docs/` - Documentation

## Testing Performed

- [x] Terraform `apply` tested in a sanitized dev configuration

## Documentation Impact

## Bug Fix Checklist

- [x] Justification for no regression test: this repository’s deploy automation is validated through script checks, chart rendering, and live environment deploy verification rather than an automated integration test for AKS + Helm + workload identity + OSMO.

## Checklist

- [ ] Commit messages follow conventional commit format
- [x] No new linting warnings introduced that are attributable to this change

### DIFF
--- a/infrastructure/setup/03-deploy-osmo-control-plane.sh
+++ b/infrastructure/setup/03-deploy-osmo-control-plane.sh
@@ -60,6 +60,7 @@ use_local_osmo=false
 config_preview=false
 chart_version_set=false
 image_version_set=false
+redis_storage_class=""
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
@@ -206,7 +207,7 @@ fi
 
 if [[ -n "$osmo_identity_client_id" ]]; then
   info "Applying SecretProviderClass for Azure Key Vault CSI driver..."
-  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id"
+  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id" "$([[ "$use_incluster_redis" == "false" ]] && echo true || echo false)"
 else
   info "Workload identity not configured; retrieving secrets from Key Vault..."
   pg_password=$(az keyvault secret show --vault-name "$keyvault" --name "psql-admin-password" --query value -o tsv)
@@ -225,6 +226,15 @@ else
       --from-literal=redis-password="$redis_key" \
       --dry-run=client -o yaml | kubectl apply -f -
   fi
+fi
+
+if [[ "$use_incluster_redis" == "true" ]]; then
+  redis_storage_class=$(detect_default_storage_class)
+  info "Creating Redis secret for in-cluster Redis..."
+  kubectl create secret generic "$SECRET_REDIS" \
+    --namespace="$NS_OSMO_CONTROL_PLANE" \
+    --from-literal=redis-password="" \
+    --dry-run=client -o yaml | kubectl apply -f -
 fi
 
 # Apply internal LB ingress if present
@@ -277,7 +287,15 @@ base_helm_args=(
 # Deploy service
 info "Deploying osmo/service..."
 helm_args=("${base_helm_args[@]}" -f "$service_values" --set "services.postgres.serviceName=$pg_fqdn" --set "services.postgres.user=$pg_user")
-[[ "$use_incluster_redis" == "false" ]] && helm_args+=(--set "services.redis.serviceName=$redis_hostname" --set "services.redis.port=$redis_port")
+if [[ "$use_incluster_redis" == "true" ]]; then
+  helm_args+=(
+    --set "services.redis.enabled=true"
+    --set "services.redis.tlsEnabled=false"
+    --set-string "services.redis.storageClassName=$redis_storage_class"
+  )
+else
+  helm_args+=(--set "services.redis.serviceName=$redis_hostname" --set "services.redis.port=$redis_port")
+fi
 [[ -n "$osmo_identity_client_id" ]] && helm_args+=(-f "$service_identity_values" --set "serviceAccount.annotations.azure\.workload\.identity/client-id=$osmo_identity_client_id")
 
 if [[ "$use_acr" == "true" ]]; then
--- a/infrastructure/setup/03-deploy-osmo-control-plane.sh
+++ b/infrastructure/setup/03-deploy-osmo-control-plane.sh
@@ -27,6 +27,7 @@ OPTIONS:
     --use-acr               Pull images from ACR deployed by 001-iac
     --acr-name NAME         Pull images from specified ACR
     --use-incluster-redis   Use in-cluster Redis instead of Azure Managed Redis
+                            (unauthenticated, non-TLS; suitable for development)
     --skip-mek              Skip MEK configuration
     --force-mek             Replace existing MEK (data loss warning)
     --mek-config-file PATH  Use existing MEK config file
@@ -60,7 +61,7 @@ use_local_osmo=false
 config_preview=false
 chart_version_set=false
 image_version_set=false
-redis_storage_class=""
+redis_storage_class="" # Will be resolved if --use-incluster-redis is set using detect_default_storage_class (requires cluster connection)
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
@@ -207,7 +208,9 @@ fi
 
 if [[ -n "$osmo_identity_client_id" ]]; then
   info "Applying SecretProviderClass for Azure Key Vault CSI driver..."
-  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id" "$([[ "$use_incluster_redis" == "false" ]] && echo true || echo false)"
+  include_redis_secret=true
+  [[ "$use_incluster_redis" == "true" ]] && include_redis_secret=false
+  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id" "$include_redis_secret"
 else
   info "Workload identity not configured; retrieving secrets from Key Vault..."
   pg_password=$(az keyvault secret show --vault-name "$keyvault" --name "psql-admin-password" --query value -o tsv)
@@ -230,6 +233,7 @@ fi
 
 if [[ "$use_incluster_redis" == "true" ]]; then
   redis_storage_class=$(detect_default_storage_class)
+  warn "In-cluster Redis runs without TLS or authentication. Use it only for development environments."
   info "Creating Redis secret for in-cluster Redis..."
   kubectl create secret generic "$SECRET_REDIS" \
     --namespace="$NS_OSMO_CONTROL_PLANE" \
--- a/infrastructure/setup/manifests/aks-secret-provider-class-external-redis.yaml
+++ b/infrastructure/setup/manifests/aks-secret-provider-class-external-redis.yaml
@@ -1,8 +1,8 @@
 # SecretProviderClass for Azure Key Vault secrets sync via CSI driver
-# Syncs psql-admin-password from Key Vault to a Kubernetes secret.
+# Syncs psql-admin-password and redis-primary-key from Key Vault to Kubernetes secrets.
 #
 # Deployed by: 03-deploy-osmo-control-plane.sh (via common.sh apply_secret_provider_class)
-# Creates: db-secret (db-password key)
+# Creates: db-secret (db-password key), redis-secret (redis-password key)
 #
 # Secrets are created when pods mount the CSI volume (configured in osmo-control-plane-identity.yaml).
 #
@@ -26,9 +26,17 @@ spec:
         - |
           objectName: psql-admin-password
           objectType: secret
+        - |
+          objectName: redis-primary-key
+          objectType: secret
   secretObjects:
     - secretName: db-secret
       type: Opaque
       data:
         - objectName: psql-admin-password
           key: db-password
+    - secretName: redis-secret
+      type: Opaque
+      data:
+        - objectName: redis-primary-key
+          key: redis-password
--- a/shared/lib/common.sh
+++ b/shared/lib/common.sh
@@ -248,13 +248,21 @@ osmo_login_and_setup() {
   local service_url="${1:?service URL required}"
   local username="${2:-guest}"
   local roles="${3:-osmo-backend}"
+  local lookup_output=""
+  local lookup_status=0
+
   info "Logging into OSMO at ${service_url}..."
   osmo login "${service_url}/" --method dev --username "$username"
   info "Ensuring dev user '$username' exists with $roles role..."
-  if ! osmo user get "$username" &>/dev/null; then
-    osmo user create "$username" --roles "$roles" >/dev/null 2>&1 || warn "OSMO user management endpoint unavailable; skipping dev user setup"
+  lookup_output=$(osmo user get "$username" 2>&1) || lookup_status=$?
+
+  if [[ "$lookup_status" -eq 0 ]]; then
+    osmo user update "$username" --add-roles "$roles" >/dev/null || warn "Unable to update OSMO dev user roles; continuing"
+  elif [[ "$lookup_output" == *"404"* || "$lookup_output" == *"not found"* ]]; then
+    osmo user create "$username" --roles "$roles" >/dev/null || warn "OSMO user management endpoint unavailable; skipping dev user setup"
   else
-    osmo user update "$username" --add-roles "$roles" >/dev/null 2>&1 || warn "Unable to update OSMO dev user roles; continuing"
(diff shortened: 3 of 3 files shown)


shared/lib/common.sh
```diff
@@ -239,28 +252,32 @@ osmo_login_and_setup() {
   osmo login "${service_url}/" --method dev --username "$username"
   info "Ensuring dev user '$username' exists with $roles role..."
   if ! osmo user get "$username" &>/dev/null; then
-    osmo user create "$username" --roles "$roles"
+    osmo user create "$username" --roles "$roles" >/dev/null 2>&1 || warn "OSMO user management endpoint unavailable; skipping dev user setup"
   else
-    osmo user update "$username" --add-roles "$roles"
+    osmo user update "$username" --add-roles "$roles" >/dev/null 2>&1 || warn "Unable to update OSMO dev user roles; continuing"
   fi
```
katriendg, line 266: The user-management commands now swallow all failures (`>/dev/null 2>&1 || warn ...`), not just the 404 that motivated the change. This also affects `04-deploy-osmo-backend.sh` which calls the same function. Network errors, auth failures, and misconfiguration will all be silently downgraded to warnings with no diagnostic output.

**Suggested change:**

```bash
if ! osmo user get "$username" &>/dev/null; then
  osmo user create "$username" --roles "$roles" 2>&1 || warn "OSMO user management endpoint unavailable; skipping dev user setup"
else
  osmo user update "$username" --add-roles "$roles" 2>&1 || warn "Unable to update OSMO dev user roles; continuing"
fi
```

At minimum, keep stderr visible so the operator can see what failed before the warning. Ideally, check for a specific 404 status rather than swallowing all errors.

shared/lib/common.sh
```diff
@@ -239,28 +252,32 @@ osmo_login_and_setup() {
...
 apply_secret_provider_class() {
   local namespace="${1:?namespace required}"
   local keyvault="${2:?keyvault name required}"
   local client_id="${3:?client_id required}"
   local tenant_id="${4:?tenant_id required}"
+  local include_redis_secret="${5:-true}"
 
   local manifest_dir
-  manifest_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)/infrastructure/setup/manifests"
+  manifest_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)/manifests"
```
katriendg, line 280: 💡 The path change from `../../` to `..` fixes the broken resolution when `BASH_SOURCE[0]` follows the symlink path (`infrastructure/setup/lib/common.sh` → `shared/lib/common.sh`). This works because bash consistently preserves the symlink path in `BASH_SOURCE`. Worth a brief inline comment noting this expectation, since the function lives in `shared/lib/` but resolves paths relative to its symlinked location.

shared/lib/common.sh
```diff
@@ -163,6 +163,19 @@ detect_service_url() {
...
+
+  storage_class=$(kubectl get storageclass -o jsonpath='{range .items[?(@.metadata.annotations.storageclass\.kubernetes\.io/is-default-class=="true")]}{.metadata.name}{"\n"}{end}' 2>/dev/null | head -n 1)
+
+  if [[ -z "$storage_class" ]]; then
+    storage_class=$(kubectl get storageclass default -o jsonpath='{.metadata.name}' 2>/dev/null || true)
+  fi
+
+  [[ -n "$storage_class" ]] || fatal "No default StorageClass detected for in-cluster Redis"
+  echo "$storage_class"
+}
```
katriendg, line 177: 💡 The annotation-based lookup is the correct primary approach. The fallback to a literal `storageclass default` name is unlikely to trigger on AKS (default classes are named `managed-csi` or `managed-csi-premium`), but it's a harmless safety net. The `fatal` exit on complete failure is appropriate.

infrastructure/setup/03-deploy-osmo-control-plane.sh
```diff
@@ -206,7 +207,7 @@ fi
 
 if [[ -n "$osmo_identity_client_id" ]]; then
   info "Applying SecretProviderClass for Azure Key Vault CSI driver..."
-  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id"
+  apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id" "$([[ "$use_incluster_redis" == "false" ]] && echo true || echo false)"
```
katriendg: 💡 The double-negation ("if NOT in-cluster, then include redis is true") is hard to parse at a glance. Consider extracting to a variable:

```bash
include_redis_secret=$([[ "$use_incluster_redis" == "false" ]] && echo true || echo false)
apply_secret_provider_class "$NS_OSMO_CONTROL_PLANE" "$keyvault" "$osmo_identity_client_id" "$tenant_id" "$include_redis_secret"
```

infrastructure/setup/03-deploy-osmo-control-plane.sh
```diff
@@ -227,6 +228,15 @@ else
   fi
 fi
 
+if [[ "$use_incluster_redis" == "true" ]]; then
+  redis_storage_class=$(detect_default_storage_class)
+  info "Creating Redis secret for in-cluster Redis..."
+  kubectl create secret generic "$SECRET_REDIS" \
+    --namespace="$NS_OSMO_CONTROL_PLANE" \
+    --from-literal=redis-password="" \
+    --dry-run=client -o yaml | kubectl apply -f -
```
katriendg, line 241: The in-cluster Redis path creates a secret with an empty password and disables TLS. This is expected for the chart's bundled Redis container, but it means the Redis instance has no authentication or encryption in transit. For dev/test this is fine. If `--use-incluster-redis` is ever used in production, this becomes a security gap.

Consider adding a brief note in the `--help` text for `--use-incluster-redis` (e.g., "Uses an unauthenticated, non-TLS Redis instance suitable for development environments").

infrastructure/setup/03-deploy-osmo-control-plane.sh
```diff
@@ -60,6 +60,7 @@ use_local_osmo=false
 config_preview=false
 chart_version_set=false
 image_version_set=false
+redis_storage_class=""
```
katriendg: 💡 `redis_storage_class` isn't populated until after config-preview exits. The preview shows "in-cluster" for Redis mode but can't confirm a suitable StorageClass exists. Moving the `detect_default_storage_class` call before the preview block (when `use_incluster_redis` is true) would close this gap, or the preview can note "storage class: resolved at deploy time".

infrastructure/setup/manifests/aks-secret-provider-class-external-redis.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - secretName: db-secret
+      type: Opaque
+      data:
+        - objectName: psql-admin-password
+          key: db-password
+    - secretName: redis-secret
+      type: Opaque
+      data:
+        - objectName: redis-primary-key
+          key: redis-password
```
katriendg, line 42: 💡 The new manifest is the old unified manifest verbatim. The base manifest now has the redis section removed. This means ~80% shared content across two files. Any change to the postgres secret sync or CSI driver config needs updating in both.

Not blocking — the current approach is straightforward and reduces script complexity vs. template manipulation. Just noting the maintenance trade-off.

katriendg (changes requested):
Good work — the PR correctly addresses all five failure modes from #316. Code follows codebase conventions (idempotent kubectl, `--dry-run=client` pattern, `set -o errexit -o nounset`), passes `bash -n` and `shellcheck` cleanly, and the live deploy validation is solid.

**One requested change:** Address the error-swallowing in `osmo_login_and_setup` (see inline comment on `shared/lib/common.sh` L255–258). At minimum, remove `>/dev/null` so stderr is visible to operators. The remaining inline comments are non-blocking suggestions.

**What looks good:**
- Idempotent secret creation via `--dry-run=client -o yaml | kubectl apply -f -`
- Correct use of `--set-string` for `storageClassName` (prevents Helm type coercion)
- `tf_get` (soft) for optional redis outputs vs `tf_require` for required values
- Annotation-based StorageClass detection with `fatal` fallback
- No cleanup drift — uninstall script already handles `redis-secret`

shared/lib/common.sh
fbeltrao: Updated `osmo_login_and_setup` so it only creates the dev user when the lookup fails with a 404-style response. Other lookup failures now keep the CLI diagnostics visible and skip user setup instead of being silently downgraded to a warning.



When user does not exists it returns:
`Server responded with status code 400 Error message: User guest1 not found Error code: 400`

shared/lib/common.sh
fbeltrao: Added a brief inline comment in `apply_secret_provider_class` explaining that `BASH_SOURCE[0]` preserves the symlinked `infrastructure/setup/lib` path, so `../manifests` resolves from the setup directory as intended.

shared/lib/common.sh
fbeltrao: Ok if I leave as-is?

infrastructure/setup/03-deploy-osmo-control-plane.sh
fbeltrao: Extracted the `include_redis_secret` flag before calling `apply_secret_provider_class` so the external Redis condition reads directly.

infrastructure/setup/03-deploy-osmo-control-plane.sh
fbeltrao: Added the development-only security note to `--use-incluster-redis` help text and a runtime warning before creating the in-cluster Redis secret.

infrastructure/setup/03-deploy-osmo-control-plane.sh
fbeltrao: We need to select between
a) Displaying the redis_storage_class during a `--config-preview`. Pros: it validates the storage class for in cluster Redis. Cons: complexity; adding a new reasons for the script to fail (not being able to connect to cluster)
b) Do not display `redis_storage_class` during a `--config-preview`. Pros: code is simpler; not fails if cluster connection is not available. Cons: we don't display resolved storage class.

I have implemented a), happy to change if the added complexity is worth it.

infrastructure/setup/manifests/aks-secret-provider-class-external-redis.yaml
fbeltrao: @katriendg, would you suggest us templating the file (or other solution), or leave as is for now?

katriendg: I believe we can leave as-is, not a huge thing.

shared/lib/common.sh
katriendg: Totally

infrastructure/setup/03-deploy-osmo-control-plane.sh
katriendg: LGTM

katriendg (approved):
Thanks for the updates, looks good to merge.

PR #356, opened 2026-03-25
fix(infrastructure): replace hardcoded NAT Gateway availability zones with variable

### LINKED ISSUE #14: NAT Gateway and Public IP fail in regions without availability zone support
## Problem

Deploying to regions without availability zone support (e.g., `westus`) fails with two errors:

```
Error: LocationNotSupportAvailabilityZones: The resource
'Microsoft.Network/publicIPAddresses/pip-ng-rba-dev-01' does not support
availability zones at location 'westus'.

Error: LocationNotSupportAvailabilityZones: The resource
'Microsoft.Network/natGateways/ng-rba-dev-01' does not support availability
zones at location 'westus'.
```

## Root Cause

In `modules/platform/networking.tf`, both `azurerm_public_ip.nat_gateway` and `azurerm_nat_gateway.main` have `zones = ["1"]` hardcoded, which is incompatible with regions that lack availability zone support.

## Fix

Add a configurable `nat_gateway_zones` variable (defaulting to `[]`) and replace the hardcoded `zones = ["1"]` with `var.nat_gateway_zones` on both resources. Users in AZ-supported regions can set `nat_gateway_zones = ["1"]` in their `terraform.tfvars`.

### Files affected

- `deploy/001-iac/variables.tf`
- `deploy/001-iac/modules/platform/variables.tf`
- `deploy/001-iac/main.tf`
- `deploy/001-iac/modules/platform/networking.tf`
- `deploy/001-iac/terraform.tfvars.example`

---
_Migrated from [Azure-Samples/azure-nvidia-robotics-reference-architecture#182](https://github.com/Azure-Samples/azure-nvidia-robotics-reference-architecture/issues/182)_
_Original assignees not transferred: liupeirong_

kgmwang1:
## Summary

Fixes #14 — NAT Gateway and Public IP fail in regions without availability zone support.

## Changes

- Add configurable `nat_gateway_zones` variable (default `[]`) to root and platform module
- Pass `nat_gateway_zones` from root `main.tf` to the platform module
- Replace hardcoded `zones = ["1"]` with `var.nat_gateway_zones` on both `azurerm_public_ip.nat_gateway` and `azurerm_nat_gateway.main`
- Add documented `nat_gateway_zones` example to `terraform.tfvars.example`

## Usage

Regions **without** AZ support (e.g. `westus`) use the default:
```hcl
nat_gateway_zones = []
```

Regions **with** AZ support (e.g. `westus3`, `eastus2`) can pin to a zone:
```hcl
nat_gateway_zones = ["1"]
```

## Validation

- `terraform fmt -check` — passed
- `terraform validate` — passed


infrastructure/terraform/variables.tf
```diff
@@ -275,6 +275,15 @@ variable "should_enable_nat_gateway" {
   default     = true
 }
 
+// WARNING: Changing zones on an existing deployment forces replacement of both the
+// NAT Gateway and its Public IP. This causes a brief outbound connectivity interruption
+// while Azure provisions new resources. Plan changes during a maintenance window.
+variable "nat_gateway_zones" {
```
katriendg, line 281: I guess copilot already flagged this one, your `main.tf` does not yet call and pass in the variable so it will be unused.

infrastructure/terraform/variables.tf
```diff
@@ -275,6 +275,15 @@ variable "should_enable_nat_gateway" {
...
+  type        = list(string)
+  description = "Availability zones for NAT Gateway and its public IP. Set to [\"1\"] in regions with AZ support. Leave empty for regions without AZ support (e.g. westus)"
+  default     = []
```
katriendg: Default here is now set to `[]` which is different from the previous hard-coded `[1]`, should we not pass in the same so we don't explicitly break an existing deployment (force replacement)?

infrastructure/terraform/variables.tf
kgmwang1: I updated variables.tf and main.tf to fix this.  Thanks for catching this miss

katriendg (approved):
Thanks @kgmwang1 I believe it looks good now.

PR #372, opened 2026-03-30
fix(training): validate AzureML and OSMO RL submissions end to end

fbeltrao:
## Description

As reported by #320, RL training is not working in Azure ML. To test it one could ask an AI assistant to run tests and validate metrics, checkpoints, uploaded files. This PR proposes a end-to-end suite that does it automatically. Treat as a conversation starter.

Included in this PR is the fix to RL training job submission in AML and end-to-end tests for Azure ML and OSMO. The end-to-end test checks that metrics are written, that the job completes successfully, that the right code is uploaded to AML. It utilizes command line tools to validate work (az, terraform, osmo).

This is the log of running an end-to-end test: 

```shell
tests/e2e/test_e2e_training.py::test_aml_rl_training_e2e [e2e] [2026-03-30 13:12:15]: Starting AzureML RL e2e test
[e2e] [2026-03-30 13:12:15]: Submitting AzureML training job for task=Isaac-Velocity-Rough-Anymal-C-v0, num_envs=64, max_iterations=10, experiment=rl-training-e2e-aml-1774876335-50fe546b
[e2e] [2026-03-30 13:12:32]: Submitted AzureML job name=silly_glass_wnlrp7sl2z
[e2e] [2026-03-30 13:12:32]: Waiting for AzureML job silly_glass_wnlrp7sl2z to start
[e2e] [2026-03-30 13:12:32]: Waiting for AzureML job silly_glass_wnlrp7sl2z to start for up to 15 minutes (poll every 30s)
[e2e] [2026-03-30 13:12:36]: Observed status=Preparing
[e2e] [2026-03-30 13:12:36]: Reached AzureML job silly_glass_wnlrp7sl2z to start with status=Preparing
[e2e] [2026-03-30 13:12:36]: Waiting for AzureML job silly_glass_wnlrp7sl2z to complete
[e2e] [2026-03-30 13:12:36]: Waiting for AzureML job silly_glass_wnlrp7sl2z to complete for up to 30 minutes (poll every 30s)
[e2e] [2026-03-30 13:12:41]: Completion poll status=Queued
[e2e] [2026-03-30 13:15:03]: Completion poll status=Running
[e2e] [2026-03-30 13:16:48]: Completion poll status=Completed
[e2e] [2026-03-30 13:16:48]: Reached AzureML job silly_glass_wnlrp7sl2z to complete with status=Completed
[e2e] [2026-03-30 13:16:48]: AzureML job silly_glass_wnlrp7sl2z completed successfully
[...]

PR #387, opened 2026-04-03
fix(evaluation): scope SIL AzureML validation code path and script reference

### LINKED ISSUE #377: fix(evaluation): SIL AzureML validation submission has broken code path and script reference
### Component

`evaluation/sil/scripts/submit-azureml-validation.sh`

### Issue Description

The SIL evaluation AzureML submission script has the same class of bugs fixed for RL training in #320 / PR #372:

1. **Entire repo uploaded as code snapshot** — [Line 168](https://github.com/microsoft/physical-ai-toolchain/blob/main/evaluation/sil/scripts/submit-azureml-validation.sh#L168) sets `code_path="$REPO_ROOT"`, uploading the full repository instead of a narrow snapshot.

2. **Command references a non-existent script path** — [Line 271](https://github.com/microsoft/physical-ai-toolchain/blob/main/evaluation/sil/scripts/submit-azureml-validation.sh#L271) uses `command=bash training/scripts/validate.sh $cmd`, but no file exists at that path. The actual validation script is at `evaluation/sil/validate.sh`.

Any attempt to run SIL validation via AzureML will fail with exit code 127 (`No such file or directory`), identical to the RL training failure described in #320.

### Expected Behavior

- The code snapshot should be scoped to only the directories needed for validation
- The command should reference the correct script path relative to the snapshot root

### Discovered During

PR #372 review — the RL training fix applied the same pattern corrections but the evaluation script was not updated.

kgmwang1:
## Summary

Fixes the SIL AzureML validation submission script which was completely broken — any attempt to submit a validation job would fail with exit code 127 (`No such file or directory`).

This applies the same class of fixes from PR #372 (RL training) to the evaluation submission script.

Closes #377

### `evaluation/sil/scripts/submit-azureml-validation.sh`

| Bug | Before | After |
|---|---|---|
| Code snapshot scope | `code_path="$REPO_ROOT"` (entire repo uploaded) | `code_path="$REPO_ROOT/evaluation"` (scoped to evaluation/) |
| Command path | `bash training/scripts/validate.sh` (does not exist) | `bash evaluation/sil/validate.sh` with symlink shim |
| Job template path | `workflows/azureml/validate.yaml` (wrong location) | `evaluation/sil/workflows/azureml/validate.yaml` |
| Directory validation | Checks for `training/` directory | Checks for `sil/` directory |

The symlink shim (`if [ ! -e evaluation ]; then ln -s . evaluation; fi`) mirrors the pattern from #372 — it ensures `validate.sh` relative path resolution (`TRAINING_DIR`, `SRC_DIR`, `PYTHONPATH`) and the `python -m evaluation.sil.policy_evaluation` import both work correctly inside the AML job container.

### `evaluation/.amlignore` (new)

Added `.amlignore` matching the existing `training/.amlignore` pattern to exclude bytecode, caches, virtual environments, and IDE files from the AML code snapshot.


evaluation/sil/scripts/submit-azureml-validation.sh
rezatnoMsirhC: We should probably update the default here to `evaluation/sil/workflows/azureml/validate.yaml`

infrastructure/terraform/terraform.tfvars.example
```diff
@@ -85,6 +85,15 @@ node_pools = {
...
 // }
 
+// NAT Gateway Availability Zones
+// Set to ["1"] in AZ-supported regions (e.g. westus3, eastus2)
+// Leave empty for regions without AZ support (e.g. westus)
+// NOTE: Changing zones on an existing deployment forces replacement of both the NAT Gateway
+// and its Public IP. To preserve existing zonal behavior, set nat_gateway_zones to your current
+// zone (e.g. ["1"]) explicitly before applying. Run `terraform plan` to confirm no unintended
+// NAT/Public IP recreation.
+nat_gateway_zones = []
```
rezatnoMsirhC: It might be safer to use `nat_gateway_zones = ["1"]` (safe default which preserves prior behavior) in this example instead of `nat_gateway_zones = []` because copying this `terraform.tfvars.example` without customizing this line would trigger replacement of both the NAT Gateway and its Public IP on existing AZ deployments, which can cause outbound connectivity interruption.

infrastructure/terraform/variables.tf
```diff
@@ -275,6 +275,15 @@ variable "should_enable_nat_gateway" {
...
 }
 
+// WARNING: Changing zones on an existing deployment forces replacement of both the
+// NAT Gateway and its Public IP. This causes a brief outbound connectivity interruption
+// while Azure provisions new resources. Plan changes during a maintenance window.
+variable "nat_gateway_zones" {
+  type        = list(string)
+  description = "Availability zones for NAT Gateway and its public IP. Set to [\"1\"] in regions with AZ support. Leave empty for regions without AZ support (e.g. westus)"
+  default     = ["1"]
+}
```
rezatnoMsirhC, line 301: [`nat_gateway_zones` in infrastructure/terraform/modules/platform/variables.tf](https://github.com/microsoft/physical-ai-toolchain/blob/main/infrastructure/terraform/modules/platform/variables.tf#L31-L45) includes some validation blocks. Shouldn't we include those here as well?

```suggestion
variable "nat_gateway_zones" {
  type        = list(string)
  description = "Availability zones for NAT Gateway and its public IP. Set to [\"1\"] in regions with AZ support. Leave empty for regions without AZ support (e.g. westus)"
  default     = ["1"]
  
  validation {
    condition     = alltrue([for z in var.nat_gateway_zones : contains(["1", "2", "3"], z)])
    error_message = "Each zone must be \"1\", \"2\", or \"3\""
  }

  validation {
    condition     = length(var.nat_gateway_zones) == length(distinct(var.nat_gateway_zones))
    error_message = "nat_gateway_zones must not contain duplicates"
  }
}
```

rezatnoMsirhC (commented):
Overall LGTM! Leaving a couple of minor comments for your review.

infrastructure/terraform/terraform.tfvars.example
kgmwang1: Great catch! Not sure how this slipped through.  Updated!

rezatnoMsirhC (approved):

nguyena2 (approved):

PR #378, opened 2026-04-01
feat(build): add terraform-docs generation pipeline

WilliamBerryiii:
## Description

This PR adds a complete terraform-docs generation and freshness checking pipeline to the CI system. A new reusable workflow downloads and **SHA256-verifies** the terraform-docs binary, then runs a PowerShell CI wrapper that detects documentation drift across all Terraform modules. The pipeline integrates into both the main branch and PR validation workflows with **soft-fail** enabled during rollout.

> The terraform-docs binary is pinned to v0.21.0 with a hardcoded SHA256 checksum verified at download time, preventing supply-chain substitution.

Closes #327

### CI Pipeline

The new *terraform-docs-check.yml* reusable workflow accepts `soft-fail`, `changed-files-only`, `terraform-docs-version`, and `terraform-docs-sha256` inputs. It downloads the terraform-docs binary, **verifies its SHA256 checksum** via `Get-FileHash`, and delegates to *Invoke-TerraformDocsCheck.ps1* for drift detection. Results are uploaded as a CI artifact.

- Wired **terraform-docs-check** into *main.yml* with `soft-fail: true` and added it to the release-please `needs` array alongside `go-lint`, `go-tests`, and `codeql-analysis`
- Wired **terraform-docs-check** into *pr-validation.yml* with `soft-fail: true` and `changed-files-only: true`
- Added `scripts/` root source-to-test mapping in *pester-tests.yml* so that changes to scripts under `scripts/` trigger their corresponding Pester tests

### CI Wrapper and Documentation Checking

*Invoke-TerraformDocsCheck.ps1* imports `LintingHelpers` and `CIHelpers` modules, runs `npm run docs:tf -- --check`, and parses drifted files from git diff output. It writes structured JSON results to `logs/terraform-docs-check-results.json` and generates a CI step summary with file-level error annotations.

- Supports a **ChangedFilesOnly** mode that uses `Get-ChangedFilesFromGit` to skip checking when no `.tf`, `.tfvars`, or `.terraform-docs.yml` files changed
- Config file changes (`.terraform-docs.yml`) trigger a full check run regardless of individual file changes

### Testing

Both new scripts have comprehensive Pester 5.x test suites using mock-heavy isolation.

- *Invoke-TerraformDocsCheck.Tests.ps1* covers tool availability, clean runs, drift detection with error annotations, `ChangedFilesOnly` skipping, and config file change triggers
[...]

PR #398, opened 2026-04-07
feat(infrastructure): add optional ADLS Gen2 data lake storage account

### LINKED ISSUE #385: feat(infra): Add dedicated ADLS Gen2 storage account
### Component

Infrastructure (Terraform/Azure)

### Problem Statement

The current infrastructure uses a single blob storage account that doubles as the Azure ML workspace default datastore. This creates two problems:

1. **Checkpoint writes fail with orbax/JAX-based training (e.g., OpenPi).** orbax writes checkpoints to a temporary directory then performs an atomic POSIX-style rename to the final path. Standard Azure Blob Storage has a flat namespace, so a rename is a copy-then-delete per blob — non-atomic and incompatible with orbax's semantics. This has caused checkpoint corruption in practice.

2. **The ML workspace storage mixes platform concerns with domain data.** AzureML system data (run metadata, snapshots, logs, file shares) shares the same account as raw data, converted datasets, and model checkpoints. This couples lifecycle policies, RBAC, networking rules, and throughput limits across unrelated workloads.

### Proposed Solution

Add a second storage account with Azure Data Lake Storage Gen2 (hierarchical namespace enabled) dedicated to domain data. The existing ML workspace storage remains untouched.

**New storage account scope:**
- Raw data ingestion (ROS bags, sensor captures)
- Raw datasets (converted and curated datasets as well if desired)
- Base model weights (fine-tuned checkpoints from training run outputs as well if desired)

**Key configuration:**
- `account_kind = "StorageV2"` with `is_hns_enabled = true`
- Private endpoints for both `blob` and `dfs` sub-resources
- Identity-based auth (no shared access keys)
- Registered as a named AzureML datastore (e.g., `data_lake`)
- Gated behind a `should_create_data_lake_storage` variable

**What stays on the existing account:** AzureML workspace default datastore (system metadata, snapshots, logs, file shares).

### Alternatives Considered

- **Enable HNS on the existing storage account.** HNS is irreversible once enabled and risks breaking AzureML internal operations that assume flat namespace behavior. Not viable without recreating the account.
- **Keep a single account, work around orbax rename.** Would require patching orbax or custom checkpoint logic to avoid atomic renames. Fragile and framework-specific — the problem recurs with any tool that expects POSIX semantics (common in the ML/JAX ecosystem).

### Additional Context

- ADLS Gen2 is not a separate Azure service — it is a standard `StorageV2` account with the HNS flag enabled. All existing blob APIs, SDKs, and tools remain compatible.
- The existing lifecycle policies (`delete-raw-bags`, `tier-converted-datasets-to-cool`, `tier-reports-to-cool-then-archive`) are data-domain rules that belong on the new account, not the ML workspace storage.
- A second account provides natural cost attribution (platform vs. data storage) and independent throughput/IOPS limits so data ingestion doesn't compete with AzureML I/O.

jjottar:
# Pull Request

## Description

Add an optional dedicated ADLS Gen2 storage account with hierarchical namespace (HNS) for domain data (datasets, model checkpoints, evaluation reports), separate from the existing AzureML workspace storage. The data lake is gated behind `should_create_data_lake_storage` (default: `false`) and follows existing patterns for naming, networking, RBAC, and lifecycle policies.

Closes #385

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `infrastructure/terraform/` - Terraform infrastructure
- [x] `docs/` - Documentation

## Testing Performed

- [x] Terraform `plan` reviewed (no unexpected changes)
- [x] Terraform `apply` tested in dev environment

### Terraform Plan (with `should_create_data_lake_storage = true`)

Plan: **12 to add, 2 to change, 1 to destroy.**

| Action      | Resource                                                                      |
|-------------|-------------------------------------------------------------------------------|
| **create**  | `module.platform.azurerm_storage_account.data_lake[0]`                        |
| **create**  | `module.platform.azurerm_storage_container.datasets[0]`                       |
| **create**  | `module.platform.azurerm_storage_container.models[0]`                         |
| **create**  | `module.platform.azurerm_storage_container.evaluation[0]`                     |
| **create**  | `module.platform.azurerm_storage_management_policy.data_lake[0]`              |
| **create**  | `module.platform.azurerm_private_dns_zone.core["storage_dfs"]`                |
| **create**  | `module.platform.azurerm_private_dns_zone_virtual_network_link.core["storage_dfs"]` |
| **create**  | `module.platform.azurerm_private_endpoint.data_lake_blob[0]`                  |
| **create**  | `module.platform.azurerm_private_endpoint.data_lake_dfs[0]`                   |
| **create**  | `module.platform.azurerm_role_assignment.user_data_lake_blob[0]`              |
| **create**  | `module.platform.azurerm_role_assignment.ml_data_lake_blob[0]`                |
| **create**  | `module.platform.azurerm_role_assignment.osmo_data_lake_blob[0]`              |
| **update**  | `module.platform.azurerm_key_vault.main` (in-place, pre-existing drift)       |
| **update**  | `module.platform.azurerm_storage_account.main` (in-place, pre-existing drift) |
| **destroy** | `module.platform.azurerm_storage_management_policy.main`                      |

The 2 in-place updates and the destroy are expected:

- **Updates:** Pre-existing drift on Key Vault and ML storage account — not caused by this feature.
- **Destroy:** ML storage lifecycle policy transitions to conditional (`count = 0` when data lake is enabled). Existing deployments without the data lake retain their lifecycle rules.

### Terraform Apply

All 12 data lake resources created successfully in `rg-roboticsch-dev-001` (switzerlandnorth):

| Resource                | Name / ID                                  |
|-------------------------|--------------------------------------------|
| Storage Account         | `stdlroboticschdev001` (HNS enabled)       |
| Container: datasets     | `datasets` (private)                       |
| Container: models       | `models` (private)                         |
| Container: evaluation   | `evaluation` (private)                     |
| Lifecycle Policy        | 3 rules (raw bags delete, datasets cool, evaluation reports cool→archive) |
| DNS Zone                | `privatelink.dfs.core.windows.net`         |
| PE: blob                | `pe-datalake-blob-roboticsch-dev-001`      |
| PE: dfs                 | `pe-datalake-dfs-roboticsch-dev-001`       |
| RBAC: current user      | Storage Blob Data Contributor on `stdl*`   |
| RBAC: ML identity       | Storage Blob Data Contributor on `stdl*`   |
| RBAC: OSMO identity     | Storage Blob Data Contributor on `stdl*`   |

### Terraform Test

```
Total: 169 passed, 0 failed, 0 errors
```

### Lint & Validation

| Check                     | Result   |
|---------------------------|----------|
| `npm run lint:tf`         | 0 issues |
| `npm run lint:tf:validate`| All directories passed |
| `npm run spell-check`     | 0 issues |
| `npm run lint:md`         | 0 errors |

### Platform Module (`infrastructure/terraform/modules/platform/`)

- **`storage.tf`** — New `azurerm_storage_account.data_lake` with `is_hns_enabled = true`, `datasets`, `models`, and `evaluation` containers, data lake lifecycle policy, blob and DFS private endpoints. ML storage lifecycle policy gated with `count = var.should_create_data_lake_storage ? 0 : 1` to avoid regression for existing deployments, and legacy fallback lifecycle prefixes corrected to target `ml-workspace/...` paths when the data lake is disabled.
- **`main.tf`** — Added `storage_dfs = "privatelink.dfs.core.windows.net"` to `base_dns_zones` (7 base zones, up from 6).
- **`variables.tf`** — New `should_create_data_lake_storage` variable (bool, default `false`).
- **`role-assignments.tf`** — Added Storage Blob Data Contributor on data lake for current user, ML identity, and OSMO identity. All gated on the data lake flag.
- **`outputs.tf`** — New `data_lake_storage_account` and `data_lake_storage_account_access` outputs (null when disabled).

### Dataviewer Module (`infrastructure/terraform/modules/dataviewer/`)

- **`variables.deps.tf`** — New optional `data_lake_storage_account` input (nullable) on the reusable dataviewer Terraform module.
- **`role-assignments.tf`** — Conditional Storage Blob Data Contributor on data lake for dataviewer identity when a caller passes the optional data lake dependency.

### Root Module (`infrastructure/terraform/`)

- **`variables.tf`** — New `should_create_data_lake_storage` root variable.
- **`main.tf`** — Pass `should_create_data_lake_storage` to platform module.
- **`outputs.tf`** — New `data_lake_storage_account` root output.
- **`terraform.tfvars.example`** — Added `should_create_data_lake_storage` with documentation.

### Tests (`infrastructure/terraform/modules/platform/tests/`)

- **`dns-zones.tftest.hcl`** — Updated zone counts (6→7 base zones).
- **`security.tftest.hcl`** — Added `data_lake_security` and `data_lake_disabled_by_default` test runs.
- **`conditionals.tftest.hcl`** — Added `data_lake_enabled` and `data_lake_disabled` test runs.

### Documentation

- **`docs/cloud/blob-storage-structure.md`** — Rewritten for two-account architecture: ML workspace storage vs data lake storage, new container/folder structure, updated lifecycle policy references.
- **`.cspell/general-technical.txt`** — Added `stdl` (data lake naming prefix).

## Documentation Impact

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -75,7 +75,8 @@
 				"redhat.vscode-yaml",
 				"ms-toolsai.vscode-ai",
 				"GitHub.copilot@prerelease",
-				"GitHub.copilot-chat@prerelease"
+				"GitHub.copilot-chat@prerelease",
+				"ise-hve-essentials.hve-core-all"
 			]
 		}
 	},
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -5,6 +5,6 @@
         "streetsidesoftware.code-spell-checker-cspell-bundled-dictionaries",
         "ms-python.vscode-pylance",
         "ms-python.python",
-        "ise-hve-essentials.hve-core",
+        "ise-hve-essentials.hve-core-all",
     ]
 }
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ This toolchain builds upon:
 - [NVIDIA Isaac Sim](https://developer.nvidia.com/isaac-sim) — Physics simulation
 - [NVIDIA OSMO](https://developer.nvidia.com/osmo) — Workflow orchestration
 - [LeRobot](https://github.com/huggingface/lerobot) — Imitation learning dataset format
+- Built with [HVE Core](https://github.com/microsoft/hve-core)
 
 ## 🤖 Responsible AI
 


infrastructure/terraform/modules/platform/storage.tf
```diff
@@ -54,11 +54,75 @@ resource "azurerm_storage_container" "ml_workspace" {
 }
 
 // ============================================================
-// Storage Lifecycle Management Policy
+// Data Lake Storage Account (ADLS Gen2)
 // ============================================================
 
-resource "azurerm_storage_management_policy" "main" {
-  storage_account_id = azurerm_storage_account.main.id
+resource "azurerm_storage_account" "data_lake" {
```
katriendg, line 60: **Lifecycle policy regression — `should_create_data_lake_storage = false` silently removes cost controls**

The removal of `azurerm_storage_management_policy.main` introduces a silent regression for all existing deployments that are not simultaneously opting in to the data lake.

**What breaks:**

On the current `main` branch, the three lifecycle rules (`raw/` delete, `converted/` cool tier, `reports/` cool→archive) live on `azurerm_storage_account.main`. This PR moves them exclusively to `azurerm_storage_management_policy.data_lake`, which is gated by `should_create_data_lake_storage`. With `should_create_data_lake_storage = false` (the default), the next `terraform apply` on any existing deployment will:

1. **Destroy** `azurerm_storage_management_policy.main` — raw bags will accumulate with no automated deletion, converted datasets stay on Hot tier indefinitely.
2. **Silently NOP** `should_enable_raw_bags_lifecycle_policy = true` and the other two lifecycle variables — they wire into the data lake resource that doesn't exist, so they have zero effect without an error.

The PR's own type-of-change checklist marks this as a non-breaking new feature, which is inconsistent with this destroy.

**Recommended fix: independent policies per account**

The cleanest model is two independent lifecycle policies, each managing its own concern, with no coupling between them. To avoid the regression for existing deployments during the migration window, add a fallback policy on the ML storage account that is active only when the data lake is disabled:

```hcl
resource "azurerm_storage_management_policy" "main" {
  count              = var.should_create_data_lake_storage ? 0 : 1
  storage_account_id = azurerm_storage_account.main.id
  # same rules as before
  ...
}
```

This is zero-regression: existing deployments keep their lifecycle policy until they explicitly enable the data lake. At that point Terraform destroys the ML policy and creates the data lake policy in the same apply — an intentional, visible transition. The three `should_enable_*_lifecycle_policy` variables remain meaningful in both states.

Alternatively, if the intent is that no one should be writing `raw/`, `converted/`, or `reports/` blobs to the ML storage account anymore (architecturally correct), then the regression is acceptable — but `should_enable_raw_bags_lifecycle_policy` and its siblings should produce a `precondition` error when set to `true` without `should_create_data_lake_storage = true`, so the break is explicit rather than silent.

jjottar: Thanks a lot @katriendg for the review and on-point catch and suggestions!

All three items addressed (and PR description updated accordingly):

1. Lifecycle policy regression — Fixed in commit 837bbd8. azurerm_storage_management_policy.main is now retained with count = var.should_create_data_lake_storage ? 0 : 1, so existing deployments keep their lifecycle rules until the data lake is explicitly enabled. Zero-regression path as you described.

2. [terraform.tfvars.example](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — Added should_create_data_lake_storage in the same commit.

3. [TERRAFORM.md](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) regeneration — Done in a separate commit (44ad760). Note: npm run docs:generate:tf regenerated all 9 directories, not just the 3 we modified. The extra 6 files are pre-existing formatting normalization from the new pipeline (PR #378). I kept them in a standalone commit so you can assess whether to include them or squash/drop that commit if you'd prefer to handle the bulk regeneration separately.

katriendg (commented):
Thank you @jjottar. 
Docs generation all looking good. 

There is just one thing with the existing policy that now applies to the ADLS HNS account, which has containers for the blobs, I believe this may need a final update before we can merge? Left an inline comment on this.

infrastructure/terraform/modules/platform/storage.tf
```diff
@@ -54,12 +54,141 @@ resource "azurerm_storage_container" "ml_workspace" {
...
+    prevent_destroy = true
+  }
+}
+
+// ============================================================
+// Data Lake Lifecycle Management Policy
+// ============================================================
+
+resource "azurerm_storage_management_policy" "data_lake" {
+  count = var.should_create_data_lake_storage ? 1 : 0
```
katriendg, line 200: **Data lake lifecycle policy prefix mismatch — rules will silently never match**

The `azurerm_storage_management_policy.data_lake` rules use bare folder prefixes (`raw/`, `converted/`, `reports/`), which are copied directly from the ML storage policy. These will not match any blobs in the data lake account.

Per the [Azure lifecycle management policy documentation](https://learn.microsoft.com/en-us/azure/storage/blobs/lifecycle-management-policy-structure#prefix-match-filter):

> A prefix string **must start with a container name**. For example, if you want to match all blobs under `https://myaccount.blob.core.windows.net/sample-container/blob1/...`, specify the `prefixMatch` as `sample-container/blob1`.

The ML storage account works with bare prefixes like `raw/` because those folders live directly under the `ml-workspace` container and the rule matches all blobs in the account whose names start with `raw/`. On the data lake, the same blobs live at `datasets/raw/`, `datasets/converted/`, and `datasets/reports/` — the container name is part of the path.

Without the container prefix, the lifecycle rules emit no error but match nothing. Raw bags will accumulate indefinitely even with `should_enable_raw_bags_lifecycle_policy = true`.

**Fix** — update `prefix_match` in `azurerm_storage_management_policy.data_lake`:

```hcl
// Rule 1: Delete raw ROS bags
filters {
  prefix_match = ["datasets/raw/"]   // was: ["raw/"]
  blob_types   = ["blockBlob"]
}

// Rule 2: Tier converted datasets to cool
filters {
  prefix_match = ["datasets/converted/"]   // was: ["converted/"]
  blob_types   = ["blockBlob"]
}

// Rule 3: Tier reports to cool then archive
filters {
  prefix_match = ["datasets/reports/"]   // was: ["reports/"]
  blob_types   = ["blockBlob"]
}
```

The `main` fallback policy (targeting `azurerm_storage_account.main`) is unaffected — that account uses a single flat container where `raw/`, `converted/`, `reports/` are valid account-level prefixes.

jjottar: Thanks for catching that @katriendg! 

Addressed in three follow-up commits:

* `b31304b` regenerates the Terraform docs via `npm run docs:generate:tf`, as requested.
* `9855072` fixes the lifecycle prefix issue and regenerates docs again. In addition to the data lake prefix correction you called out, I also fixed the same prefix-matching problem in the legacy non-data-lake path on the main storage account. That part was pre-existing and not directly noted in review, but it is the same class of issue: for flag-off deployments, the fallback lifecycle policy now targets the existing `ml-workspace/...` paths so behavior remains consistent with the pre-data-lake setup.
* `ebf534a` adds a separate `evaluation` container and updates the reports lifecycle rule to `evaluation/reports/`. That was not directly requested in review, but I made the change for consistency so datasets remain under `datasets`, model artifacts remain under `models`, and evaluation outputs live under `evaluation`.

I reran Terraform tests and validation after the changes.

katriendg: That's perfect thank you!

Issue #41, opened 2026-03-11
infra(cloud): BYOK encryption configuration


WilliamBerryiii:
## Summary

Configure customer-managed encryption keys (BYOK) for Azure Blob Storage via Key Vault integration. The configuration is opt-in, parameterized in the IaC template, with automated 90-day key rotation and fail-closed behavior when the CMK is unavailable.

## Deliverables

- Bicep/Terraform module for Key Vault + CMK configuration (`infra/modules/byok-encryption/`)
- Key rotation policy (90-day auto-rotation)
- BYOK setup documentation (`docs/security/byok-configuration.md`)
- Validation test: encrypt/decrypt cycle with customer key

## Acceptance Criteria

- [ ] Storage account supports BYOK via Azure Key Vault
- [ ] Bicep/Terraform includes Key Vault + CMK configuration (parameterized, opt-in)
- [ ] Key rotation policy: auto-rotate every 90 days
- [ ] Fail-closed behavior: blob operations blocked if CMK is unavailable or revoked
- [ ] BYOK configuration is opt-in via IaC parameter `enable_byok = true|false` (default: false)
- [ ] Validated: data accessible only with correct key

---
_Migrated from [Azure-Samples/azure-nvidia-robotics-reference-architecture#233](https://github.com/Azure-Samples/azure-nvidia-robotics-reference-architecture/issues/233)_
_Original assignees not transferred: MarkForesman_

PR #542, opened 2026-04-22
feat(infrastructure): add conversion pipeline Terraform module

### LINKED ISSUE #39: infra(cloud): conversion pipeline Bicep/Terraform template
## Summary

Create an IaC template (Bicep or Terraform) that provisions all Azure resources for the conversion pipeline: storage accounts, Event Grid topic/subscription, Fabric workspace, networking, and RBAC assignments. Parameterized for environment promotion (dev/staging/prod) and must pass checkov security scans.

## Deliverables

- Bicep or Terraform module (`infra/modules/conversion-pipeline/`)
- Parameter files for dev, staging, and prod environments
- RBAC role assignments and checkov scan configuration

## Acceptance Criteria

- [ ] Deploys: storage account (raw + converted containers), Event Grid topic/subscription, Fabric workspace
- [ ] Parameterized for environment (dev/staging/prod)
- [ ] RBAC assignments for service identities
- [ ] Passes checkov security scan
- [ ] Private endpoints configured for storage account
- [ ] Lifecycle policies for blob containers (30-day raw deletion, 90-day archive)
- [ ] Template validates without errors

---
_Migrated from [Azure-Samples/azure-nvidia-robotics-reference-architecture#231](https://github.com/Azure-Samples/azure-nvidia-robotics-reference-architecture/issues/231)_

nguyena2:
## Summary

Adds a new `conversion-pipeline` Terraform module that provisions the Azure infrastructure for the dataset conversion pipeline, along with supporting examples, root-level wiring, and CI updates.

Fixes #39

## Changes

- **New module** `infrastructure/terraform/modules/conversion-pipeline/`:
  - Storage account with blob/dfs private endpoints, lifecycle policies, and diagnostic settings
  - Event Grid topic with role assignments
  - Fabric workspace integration and role assignments
  - Full module documentation (`README.md`, `TERRAFORM.md`) and `terraform test` coverage
- **Root deployment** (`infrastructure/terraform/`): wire the new module into `main.tf`, expose outputs, add variables, and pin provider versions
- **Examples**: add `terraform.tfvars.dev`, `terraform.tfvars.staging`, and `terraform.tfvars.prod` samples
- **CI**: extend `pr-validation.yml` and add a new `terraform-security.yml` workflow
- **Docs**: regenerate `TERRAFORM.md` files across modules

## Validation

- `terraform test` cases under `modules/conversion-pipeline/tests/`
- Updated PR validation workflow runs `lint:tf:validate` and the new security scan

## CI / Security

This PR introduces a new code-scanning surface for Terraform via Checkov. Extended from #39 to cover security validation of the new module alongside the broader Terraform tree.

- **Scanner**: [Checkov](https://www.checkov.io/) (`bridgecrewio/checkov-action`), framework `terraform`, `download_external_modules: false`.
- **Scope**: scans `infrastructure/terraform` recursively (all root deployments and modules under that path). Configurable via the reusable workflow's `working-directory` input.
- **Trigger**: runs as the `terraform-security` job in `pr-validation.yml` on every PR, and is reusable via `workflow_call` (`.github/workflows/terraform-security.yml`).
- **Soft-fail default**: `soft-fail: true`. Findings do not block merges today; the job is informational while the existing finding matrix is triaged and remediated.
- **Where results land**:
  - SARIF uploaded to the GitHub **Security** tab under the `checkov` category (visible to security and repo admins via code scanning alerts).
  - SARIF also uploaded as a workflow artifact (`checkov-sarif`, 30-day retention) for offline review.
- **Rollout plan to flip soft-fail to hard-fail**:
  1. Triage the initial Checkov findings surfaced in the Security tab and either remediate or file suppressions with justification.
  2. Once the backlog is at zero (or only contains accepted, suppressed findings), flip `soft-fail: false` in the `terraform-security` job in `pr-validation.yml` and add `terraform-security` to the required-status-check aggregator.
  3. Track this rollout via a follow-up issue referencing #39.


infrastructure/examples/terraform.tfvars.staging
```diff
@@ -0,0 +1,82 @@
...
+
+// OSMO Backend Services with HA
+should_deploy_postgresql = true
+should_deploy_redis      = true
+
+// PostgreSQL HA
+postgresql_sku_name = "GP_Standard_D2s_v3"
+postgresql_high_availability = {
+  enabled                   = true
+  standby_availability_zone = "2"
```
rezatnoMsirhC, line 47: The `postgresql_high_availability` block uses `enabled`, but the root variable type in [`variables.tf` line 198](https://github.com/microsoft/physical-ai-toolchain/blob/feat/issue-39-conversion-pipeline-iac/infrastructure/terraform/variables.tf#L198) defines `should_enable`:

```hcl
variable "postgresql_high_availability" {
  type = object({
    should_enable             = bool
    standby_availability_zone = optional(string)
  })
```

Terraform's [object type constraint](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#object) rejects any attribute not declared in the type definition, so `terraform plan -var-file=.../terraform.tfvars.staging` will fail with `An argument named "enabled" is not expected here`. This is not caught by CI because `terraform validate` runs without loading the example tfvars files.

**Suggested fix:**

```hcl
postgresql_high_availability = {
  should_enable             = true
  standby_availability_zone = "2"
}
```

infrastructure/terraform/modules/conversion-pipeline/role-assignments.tf
```diff
@@ -0,0 +1,32 @@
...
+ *   and Storage Blob Data Contributor on converted (only when supplied).
+ */
+
+resource "azurerm_role_assignment" "eventgrid_dlq_writer" {
+  count = var.should_enable_event_grid_dead_letter ? 1 : 0
+
+  scope                = azurerm_storage_account.this.id
+  role_definition_name = "Storage Blob Data Contributor"
+  principal_id         = azurerm_eventgrid_system_topic.blob.identity[0].principal_id
+}
```
rezatnoMsirhC, line 18: The Event Grid system topic's managed identity is granted `Storage Blob Data Contributor` at the storage account scope, but it only needs to write dead-letter blobs to the `event-grid-dlq` container. Storage-account scope also gives it full read/write over `raw` and `converted`, which violates least privilege.

**Suggested fix:** Scope to the DLQ container only:

```hcl
resource "azurerm_role_assignment" "eventgrid_dlq_writer" {
  count                = var.should_enable_event_grid_dead_letter ? 1 : 0
  scope                = azurerm_storage_container.event_grid_dlq[0].id
  role_definition_name = "Storage Blob Data Contributor"
  principal_id         = azurerm_eventgrid_system_topic.blob.identity[0].principal_id
}
```

infrastructure/terraform/modules/conversion-pipeline/event-grid.tf
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  dead_letter_identity {
+    type = "SystemAssigned"
+  }
+
+  delivery_identity {
+    type = "SystemAssigned"
+  }
```
rezatnoMsirhC, line 70: `dead_letter_identity` is a static block while `storage_blob_dead_letter_destination` is dynamic. When `should_enable_event_grid_dead_letter = false`, the subscription has a `dead_letter_identity` with no destination, which is a provider-level inconsistency that may produce a validation error or silently create an invalid configuration.

**Suggested fix:** Wrap `dead_letter_identity` in a matching `dynamic` block:

```hcl
dynamic "dead_letter_identity" {
  for_each = var.should_enable_event_grid_dead_letter ? [1] : []
  content {
    type = "SystemAssigned"
  }
}
```

.github/workflows/terraform-security.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Create logs directory
+        run: mkdir -p logs
+
+      # TODO(security): Pin bridgecrewio/checkov-action to a specific commit SHA
+      # via the dependency-pinning workflow before disabling soft-fail.
+      - name: Run Checkov
+        id: checkov
```
rezatnoMsirhC, line 34: The TODO says this action still needs to be SHA-pinned, but it is already pinned to `38a95e98d734de90a74687a8d986d6b06107f342` on the very next line. Do we still need this TODO?

rezatnoMsirhC (dismissed):
Good module structure and conventions throughout. Four comments below: one high-severity bug in the staging tfvars that will fail at plan time, one least-privilege RBAC issue, one Event Grid reliability concern, and one stale TODO to remove.

infrastructure/terraform/versions.tf
```diff
@@ -31,3 +35,9 @@ provider "azurerm" {
 }
 
 provider "azapi" {}
+
+// Microsoft Fabric provider authenticates via FABRIC_TENANT_ID, FABRIC_CLIENT_ID,
+// FABRIC_CLIENT_SECRET environment variables. The service principal must be in
+// a security group allow-listed under the Fabric tenant admin setting
+// "Service principals can use Fabric APIs".
+provider "fabric" {}
```
WilliamBerryiii: **RI-09 🟨 Low — Empty `provider "fabric" {}` lacks env-var precondition**

The Fabric provider is declared with no configuration, relying entirely on `FABRIC_TENANT_ID`, `FABRIC_CLIENT_ID`, and `FABRIC_CLIENT_SECRET` being set in the operator's environment (correctly noted in the comment above). When any of these is missing, the provider fails late with an opaque auth error during the first Fabric resource plan/apply.

Suggestion: add a `precondition` (e.g., on a `null_resource` guard or directly on the first Fabric resource) that asserts each variable is non-empty, or document the requirement in the root README's prerequisites section. Either approach turns a mid-apply failure into an immediate, actionable error.

infrastructure/terraform/modules/conversion-pipeline/README.md
```diff
@@ -0,0 +1,126 @@
...
+| Dead-letter queue| `should_enable_event_grid_dead_letter` | Backed by an in-account `event-grid-dlq` container          |
+| Fabric capacity  | `should_create_fabric_capacity`        | Disable to reuse an existing capacity                       |
+| Fabric workspace | `should_create_fabric_workspace`       | Requires `fabric_capacity_uuid`. See Two-pass deployment    |
+| Diagnostics      | `should_enable_diagnostic_settings`    | Routes blob and account metrics to Log Analytics            |
+
+## 🔄 Two-pass deployment
+
+The `fabric_workspace.capacity_id` argument expects the Fabric capacity GUID, not the ARM resource ID. The `azurerm_fabric_capacity` resource does not surface the GUID, so workspace creation is split across two applies:
+
+1. First apply: leave `should_create_fabric_workspace = false` (or `fabric_capacity_uuid = null`). The capacity is created.
```
WilliamBerryiii: **RI-06 🟨 Low — UX-fragile manual GUID lookup in deployment instructions**

Asking operators to copy a GUID out of `terraform plan` output and paste it back into `terraform.tfvars` for a second `apply` is error-prone (typos silently produce invalid Fabric workspace IDs that fail late) and breaks any unattended/CI deployment path.

Preferred alternatives:

- Use `terraform apply -target=module.conversion_pipeline.fabric_workspace.this` for pass 1, then a full `apply`, with the second pass reading the workspace ID via `data "fabric_workspace"` or a module output rather than a tfvars edit.
- Or, document the two-pass flow as a wrapper script under `infrastructure/setup/` that calls `terraform output -raw conversion_pipeline_workspace_id` and writes the value to a generated `*.auto.tfvars` file.
- At minimum, add a `precondition` on the second-pass resource that fails fast with a clear message when the workspace ID is empty.

infrastructure/terraform/modules/conversion-pipeline/fabric.tf
```diff
@@ -0,0 +1,40 @@
...
+ * variables FABRIC_TENANT_ID, FABRIC_CLIENT_ID, FABRIC_CLIENT_SECRET. The
+ * service principal must be in a security group that is allow-listed in the
+ * Fabric tenant admin setting "Service principals can use Fabric APIs".
+ */
+
+resource "azurerm_fabric_capacity" "this" {
+  count = var.should_create_fabric_capacity ? 1 : 0
+
+  name                = "fc${var.resource_prefix}${var.environment}${var.instance}"
+  resource_group_name = var.resource_group.name
```
WilliamBerryiii, line 38: **RI-02 🟨 Low — Naming convention violation**

The Fabric capacity name `"fc${var.resource_prefix}${var.environment}${var.instance}"` concatenates segments with no separator, producing values like `fcnvidiadev001`. This deviates from the resource-naming standard documented in `.github/copilot-instructions.md` (Terraform Conventions section), which prescribes `{abbreviation}-{resource_prefix}-{environment}-{instance}` (hyphenated) for most resources, with no-hyphen reserved for Key Vault (`kv`), Storage (`st`), and ACR (`acr`). Fabric Capacity (`fc`) is not on the no-hyphen allow-list.

Suggested fix:

```hcl
name = "fc-${local.resource_name_suffix}"
```

This aligns with the convention used by sibling resources in this very module (e.g., `evgt-${local.resource_name_suffix}` in `event-grid.tf`). If Fabric Capacity has an undocumented naming restriction that forced the no-hyphen form, please add an inline comment justifying the exception and consider adding `fc` to the no-hyphen allow-list in the conventions doc.

infrastructure/terraform/modules/conversion-pipeline/README.md
```diff
@@ -0,0 +1,126 @@
+---
+title: Conversion Pipeline Module
+description: Terraform module that provisions ADLS Gen2 storage, Event Grid, and Microsoft Fabric for the raw -> converted ingest pipeline
+author: Edge AI Team
```
WilliamBerryiii: **RI-07 🟦 Trivial — Author frontmatter inconsistency**

`author: Edge AI Team` does not match the convention used by sibling Terraform module READMEs in this repository (which typically use `author: Physical AI Team` or omit the field). Suggest aligning to the established value to keep generated docs and search facets consistent.

infrastructure/terraform/modules/conversion-pipeline/main.tf
```diff
@@ -0,0 +1,101 @@
...
+  sas_policy {
+    expiration_period = "07.00:00:00"
+    expiration_action = "Log"
+  }
+
+  // checkov:skip=CKV_AZURE_33: Queue service unavailable on hierarchical namespace accounts.
+  // checkov:skip=CKV_AZURE_206: Replication tier parameterized per environment (LRS dev, ZRS staging, GRS prod).
+  // checkov:skip=CKV2_AZURE_1: Customer-managed key encryption tracked in follow-up WI-01.
+  // checkov:skip=CKV2_AZURE_18: Customer-managed key with Key Vault tracked in follow-up WI-01.
+  // checkov:skip=CKV2_AZURE_50: Immutability policy tracked in follow-up WI-01.
```
WilliamBerryiii: **RI-03 🟧 Medium — `WI-01` placeholder unresolved**

These three `checkov:skip` directives (and at least one more above this line) defer customer-managed key encryption (CKV2_AZURE_1, CKV2_AZURE_18) and immutability policy (CKV2_AZURE_50) to a "follow-up WI-01" — but `WI-01` is a planning placeholder, not a tracked work item identifier. Suppressing four security controls against an unresolved token leaves no auditable trail and risks the work being silently dropped.

Required before merge:

1. File the actual follow-up issue(s) in this repo (or the linked work-tracking system) for CMK enablement and immutability policy.
2. Replace `WI-01` in every `checkov:skip` justification with the real issue URL or `#NNN` reference.
3. Optionally consolidate the rationale into the module README under a "Known security deferrals" section so reviewers don't have to grep skip directives.

infrastructure/terraform/modules/conversion-pipeline/event-grid.tf
```diff
@@ -0,0 +1,67 @@
...
+ * # Event Grid System Topic and Subscription
+ *
+ * System topic on the conversion storage account that fires BlobCreated events
+ * to the conversion subscriber (Function App or Fabric pipeline). Multi-suffix
+ * filtering uses advanced_filter.string_ends_with because subject_filter only
+ * supports a single suffix value.
+ */
+
+resource "azurerm_eventgrid_system_topic" "blob" {
+  name                = "evgt-${local.resource_name_suffix}"
```
WilliamBerryiii, line 12: **RI-10 ℹ️ Verify-only — Event Grid System Topic naming**

`evgt-${local.resource_name_suffix}` follows the hyphenated convention correctly. Flagging only to confirm `evgt` is the intended abbreviation for Event Grid System Topic in this codebase — the Azure CAF abbreviation is `egst`/`evgt` depending on source. If `evgt` is the established choice, no action needed; if not, normalize to `egst` and update any sibling references. Resolve as you see fit.

WilliamBerryiii (changes requested):
## Review Summary — Conversion Pipeline IaC

Thanks for this; the module structure (Edge → Raw blob → Event Grid → Function App → Converted blob → Fabric Lakehouse) is coherent and easy to follow. Requesting changes for the items below — most are small, but **RI-03** and **RI-04** are blockers in my view.

### Acknowledgements

- Existing reviewer @rezatnoMsirhC has already covered E1–E4 (Function App SKU/runtime, identity, network ACLs, retention defaults). I've intentionally left those alone to avoid duplicate noise; please address them in the same revision pass.
- I withdrew several internal candidates (provider version pinning, blob CORS, etc.) after confirming they were either already raised or non-issues against repo conventions.

### Blocking items (inline)

- **RI-03 🟧 Medium** — `main.tf:64` — Four `checkov:skip` directives reference unresolved `WI-01` placeholder. File the real follow-up issue(s) and replace `WI-01` with the issue reference before merge.
- **RI-04 🟧 Medium** — `infrastructure/terraform/variables.tf` (root, ~lines 547–573) and `infrastructure/terraform/main.tf` (~lines 217–260) — The `conversion_pipeline_config` object exposed at the root does **not** surface every meaningful module variable. Operators today cannot, for example, override Function App SKU, blob retention windows, or Event Grid retry policy without editing the module directly. Either: (a) widen the root config object to mirror the module's full variable surface, or (b) document explicitly which knobs are intentionally fixed at the module layer and why. As-is the partial passthrough is a hidden footgun.

### Non-blocking items (inline)

- **RI-02 🟨 Low** — `fabric.tf:23` — Fabric capacity name lacks separators; align with hyphenated convention (`fc-${local.resource_name_suffix}`).
- **RI-06 🟨 Low** — `README.md:121` — Two-pass deployment with manual GUID copy is UX-fragile; suggest `-target` first pass or wrapper script.
- **RI-09 🟨 Low** — `versions.tf:43` — Add precondition or doc note for `FABRIC_TENANT_ID` / `FABRIC_CLIENT_ID` / `FABRIC_CLIENT_SECRET` env vars so missing values fail fast.
- **RI-07 🟦 Trivial** — `README.md:4` — `author: Edge AI Team` inconsistent with sibling module READMEs.
- **RI-10 ℹ️ Verify-only** — `event-grid.tf:11` — Confirm `evgt` (vs. `egst`) is the intended Event Grid System Topic abbreviation.

### Body-only finding

- **RI-05 🟨 Low** — `infrastructure/terraform/variables.tf` (root, ~lines 547–573) duplicates module-level defaults (Function App SKU `F2`, storage `ZRS`, raw retention `30 days`, converted retention `90 days`). Each default now lives in two places and will drift. Recommendation: drop the defaults from the root variable block and let the module own them, or remove them from the module and centralize at the root — either is fine, but not both.

### Validation evidence

- All 6 inline anchors verified against the PR diff hunks for commit `01de39ebf3db4eff0eba0cf7966b44e4d71c1f22` before posting.
- Findings cross-checked against `.github/copilot-instructions.md` (Terraform Conventions section) and existing PR conversation to avoid duplication of @rezatnoMsirhC's review.

Happy to re-review once **RI-03** and **RI-04** are addressed.

infrastructure/examples/terraform.tfvars.staging
nguyena2: Renamed enabled → should_enable to match the canonical schema

infrastructure/terraform/modules/conversion-pipeline/role-assignments.tf
nguyena2: Narrowed the eventgrid_dlq_writer role assignment scope from azurerm_storage_account.this.id to azurerm_storage_container.event_grid_dlq[0].id. The Event Grid system topic MI now only has Storage Blob Data Contributor on the DLQ container, not on raw/converted. This matches the pattern already used by the two Fabric SP role assignments in the same file.

infrastructure/terraform/modules/conversion-pipeline/event-grid.tf
nguyena2: Wrapped dead_letter_identity in a dynamic block keyed on var.should_enable_event_grid_dead_letter, mirroring the existing dynamic "storage_blob_dead_letter_destination" gating directly above it. With DLQ disabled, the subscription no longer renders an unused dead_letter_identity block.

.github/workflows/terraform-security.yml
nguyena2: Removed the TODO comment. The action immediately below it (bridgecrewio/checkov-action@38a95e98...) is already SHA-pinned, so the comment was outdated.

infrastructure/terraform/versions.tf
nguyena2: Addressed without data "external" enforcement. Two changes:

1. Added a lifecycle.precondition on fabric_workspace.this requiring non-empty fabric_capacity_uuid so the second-pass requirement fails fast with a clear error pointing at the README.
2. Documented FABRIC_TENANT_ID, FABRIC_CLIENT_ID, FABRIC_CLIENT_SECRET as prerequisites in [README.md] Quick Start, with a link to the conversion-pipeline two-pass deployment section.

infrastructure/terraform/modules/conversion-pipeline/README.md
nguyena2: Expanded the "Two-pass deployment" section in modules/conversion-pipeline/README.md to:

1. Reference the new fabric_workspace.this lifecycle.precondition so reviewers know misconfiguration fails fast.
2. Note the follow-up wrapper infrastructure/setup/05-fabric-capacity-bootstrap.sh that will automate the GUID lookup between applies (tracked separately).

infrastructure/terraform/modules/conversion-pipeline/fabric.tf
nguyena2: Module test default_naming failed with fc-${local.resource_name_suffix}

infrastructure/terraform/modules/conversion-pipeline/README.md
nguyena2: Changed author: Edge AI Team → author: Microsoft Robotics-AI Team to match the canonical string used by the five sibling module READMEs

infrastructure/terraform/modules/conversion-pipeline/main.tf
nguyena2: Replaced WI-01 placeholders with concrete tracking issues:

CKV2_AZURE_1 → #41 (BYOK encryption configuration — existing open issue, covers CMK on storage)
CKV2_AZURE_18 → #41 (same — covers Key Vault CMK)
CKV2_AZURE_50 → #578 (immutability policy on critical blob containers — newly filed)
Also surfaced the suppressions in the module README under a new "🔒 Known security deferrals" section so the audit trail is discoverable, not buried in source comments.

infrastructure/terraform/modules/conversion-pipeline/event-grid.tf
nguyena2: Verified — evgt is the established abbreviation for azurerm_eventgrid_system_topic and is already used elsewhere in this module's resource naming (evgs- for the subscription on the same topic).

katriendg (commented):
Thanks @nguyena2 for this great addition and integrating Fabric into the picture.
Reviewing the work, I have two generic notes which were based on reflections as I was looking through your updates. I'll let you decide being able to integrate these. For the first one I believe it's valuable, the second one I'll leave to you to assess if relevant.

## Comment 1 — Can we drop the two-pass deploy and `FABRIC_*` env vars

**Files referenced:**

- `infrastructure/terraform/modules/conversion-pipeline/fabric.tf`
- `infrastructure/terraform/modules/conversion-pipeline/variables.tf`
- `infrastructure/terraform/modules/conversion-pipeline/README.md`
- `infrastructure/terraform/versions.tf`
- `infrastructure/README.md`

### Comment body

The two-pass workflow and the `FABRIC_TENANT_ID` / `FABRIC_CLIENT_ID` / `FABRIC_CLIENT_SECRET` prerequisite are avoidable. `microsoft/edge-ai` already solves the same `azurerm_fabric_capacity` "GUID not exposed" problem in a single apply and uses Azure CLI auth — no SP secret required. We should align with that pattern, if possible.

Wdty?

**Reference:** [edge-ai/src/000-cloud/031-fabric/terraform/main.tf](https://github.com/microsoft/edge-ai/blob/main/src/000-cloud/031-fabric/terraform/main.tf) and [versions.tf](https://github.com/microsoft/edge-ai/blob/main/src/000-cloud/031-fabric/terraform/versions.tf).

Edge-ai resolves the capacity GUID via a deferred `data "fabric_capacity"` lookup keyed by `display_name`. The `terraform_data` indirection prevents plan-time queries (handles the same testing/build-system constraint that prompted the two-pass design here):

```hcl
resource "terraform_data" "defer_fabric_capacity_created" {
  count      = var.should_create_fabric_capacity ? 1 : 0
  input      = { display_name = local.fabric_capacity_name }
  depends_on = [module.fabric_capacity]
}

data "fabric_capacity" "created" {
  count        = length(terraform_data.defer_fabric_capacity_created)
  display_name = terraform_data.defer_fabric_capacity_created[0].output.display_name
}

locals {
  capacity_id = try(data.fabric_capacity.created[0].id, data.fabric_capacity.existing[0].id, null)
}

resource "fabric_workspace" "this" {
  count        = var.should_create_fabric_workspace ? 1 : 0
  capacity_id  = local.capacity_id
  display_name = "fws-${local.resource_name_suffix}"
  description  = "Conversion pipeline workspace (${var.environment})"
}
```

Edge-ai's `versions.tf` also has **no `provider "fabric" {}` block** and the blueprint README documents no `FABRIC_*` env vars. The `microsoft/fabric` provider supports Azure CLI auth (`use_cli` defaults to `true`) and picks up the same `az login` already used by `azurerm`. The SP-secret env-var path is one auth mode of several, not a requirement.

**Proposed changes:**

1. Replace `fabric_capacity_uuid` (variable + workspace gating + `lifecycle.precondition`) with the deferred `data "fabric_capacity"` pattern above. Drop the entire "Two-pass deployment" README section.
2. Remove the `provider "fabric" {}` block and its comment from `infrastructure/terraform/versions.tf`. The provider auto-discovers via Azure CLI when no provider block is declared.
3. Remove the `FABRIC_TENANT_ID` / `FABRIC_CLIENT_ID` / `FABRIC_CLIENT_SECRET` prerequisite from both `infrastructure/README.md` and `modules/conversion-pipeline/README.md`. Replace with a note that `az login` is sufficient and that the operator's identity (or an opt-in SP, if the user explicitly chooses SP auth) must be in the "Service principals can use Fabric APIs" tenant allow-list.
4. Pin the `fabric` provider to `1.3.0` (matching edge-ai) instead of `~> 1.0`, which would let any new `1.x` minor drift in.

This collapses the workflow to a single `terraform apply`, removes the secret-handling burden in CI, and aligns with the established Microsoft pattern.

## Comment 2 — Storage overlap with the platform data-lake account

**Files referenced:**

- `infrastructure/terraform/modules/platform/storage.tf`
- `infrastructure/terraform/modules/conversion-pipeline/main.tf`
- `infrastructure/terraform/modules/conversion-pipeline/lifecycle.tf`
- `infrastructure/terraform/main.tf`

### Comment body

The platform module already creates an HNS Gen2 data-lake account (`stdl{prefix}{env}{instance}`) with `datasets`, `models`, `evaluation` containers and lifecycle rules on `datasets/raw/` (delete), `datasets/converted/` (cool), and `evaluation/reports/` (cool → archive). See [`modules/platform/storage.tf`](../../../infrastructure/terraform/modules/platform/storage.tf) lines 61–120 and 200+.

The conversion-pipeline module creates a second HNS Gen2 account (`stcp{prefix}{env}{instance}`) with `raw`, `converted`, `event-grid-dlq` containers and the same lifecycle semantics on `raw/` and `converted/`.

With the defaults `should_create_data_lake_storage = true` and `should_deploy_conversion_pipeline = true`, the deployment provisions:

- Two HNS Gen2 storage accounts in the same resource group
- Two `raw` / `converted` lanes with duplicate lifecycle policies
- Two sets of blob + dfs private endpoints in the same PE subnet

The README's blast-radius rationale — *"The AzureML extension storage account is managed separately by `modules/platform`"* — refers to the **`st...`** ML-workspace account, not **`stdl...`** (the data-lake account). The data-lake account already serves the raw/converted use case and is silently duplicated.

There is no cross-gating in `infrastructure/terraform/main.tf` (`should_create_data_lake_storage` and `should_deploy_conversion_pipeline` are independent flags), and the conversion-pipeline README/variables make no mention of `stdl`.

**Requested resolution (pick one and document it):**

- **Option A — Reuse the platform data-lake account (preferred).** Add `data_lake_storage_account` to `modules/conversion-pipeline/variables.deps.tf` as a typed object dependency, and create the `raw` / `converted` / `event-grid-dlq` containers (or use the existing `datasets/` prefix) inside the platform-owned account. Drop `stcp...` and the duplicate private endpoints and lifecycle policy. This matches the `variables.deps.tf` pattern already used by `modules/sil/` and `modules/dataviewer/`.
- **Option B — Mutually exclude.** Add a root-level `precondition` that fails when both `should_create_data_lake_storage` and `should_deploy_conversion_pipeline` are `true`, and update the dev/staging/prod tfvars examples to flip `should_create_data_lake_storage = false` when the conversion pipeline is enabled. Document the rationale in both READMEs.

Option A is preferred — it eliminates the overlap rather than papering over it, and it preserves the intended platform/conversion-pipeline split (compute and event plumbing stay in conversion-pipeline; durable storage stays platform-owned).

infrastructure/terraform/modules/conversion-pipeline/fabric.tf
```diff
@@ -0,0 +1,50 @@
+/**
+ * # Microsoft Fabric Capacity and Workspace
+ *
+ * Provisions an azurerm_fabric_capacity (gated by should_create_fabric_capacity)
+ * and a fabric_workspace via the microsoft/fabric provider.
+ *
+ * The fabric_workspace resource's capacity_id is the Fabric capacity GUID (not
```
katriendg, line 7: Please see my note in the PR named "Comment 1 — Can we drop the two-pass deploy and FABRIC_* env vars"

.github/workflows/terraform-security.yml
```diff
@@ -0,0 +1,57 @@
+name: Terraform Security Scan
```
katriendg, line 1: nit: could you expand the PR description (or add a short "CI / Security" section) covering: scope of files scanned, soft-fail default, where results land, and any rollout plan to flip soft-fail to hard-fail later? Reviewers and downstream owners (security, repo admins) should not have to read the workflow YAML to learn that a new code-scanning surface is being introduced. Or at least reference it in the original associated issue, so we know this was extended from the original issue. Thanks.

infrastructure/terraform/modules/conversion-pipeline/fabric.tf
nguyena2: Changed to model the work from edge-ai

.github/workflows/terraform-security.yml
nguyena2: PR description has been updated.

infrastructure/terraform/main.tf
```diff
@@ -209,3 +209,55 @@ module "sil" {
...
+
+// ============================================================
+// Conversion Pipeline Module - Raw -> Converted Ingest
+// ============================================================
+
+// Precondition guard: the conversion pipeline reuses the platform-owned
+// data-lake account, so the platform must provision it. Module call blocks
+// do not support `lifecycle.precondition` directly, so the check lives on a
+// terraform_data resource that the module depends on.
+resource "terraform_data" "conversion_pipeline_precondition" {
```
katriendg, line 221: nice one

katriendg (approved):
Thanks for update and switch with the Fabric provide & data lake, looks clean now!

WilliamBerryiii (approved):

Issue #588, opened 2026-04-28
fix(workflows): AW Dependabot PR Review posts no comments or review under workflow_run trigger


katriendg:
## Summary

The *AW Dependabot PR Review* agentic workflow runs successfully on Dependabot PR Validation completions, but the agent's review output is silently dropped: no inline review comments are posted and no review is submitted on the triggering PR (e.g. https://github.com/microsoft/physical-ai-toolchain/pull/587).

## Evidence

Run https://github.com/microsoft/physical-ai-toolchain/actions/runs/25049076721 (triggered by PR Validation completion on PR #587) shows the agent emitted 6 messages (5 `create_pull_request_review_comment` + 1 `submit_pull_request_review` with a 6,514-byte body), all dropped by the `safe_outputs` job:

```
Processing message 1/6: create_pull_request_review_comment
Target is "triggering" but not running in pull request context, skipping review comment creation
⏭ Message 1 (create_pull_request_review_comment) skipped — Not in pull request context
... (same for messages 2-5)
Processing message 6/6: submit_pull_request_review
=== Finalizing PR Review ===
##[warning]No review context set - cannot submit review
##[warning]✗ Failed to submit PR review: No review context available

=== Processing Summary ===
Total messages: 6
Successful: 1
Failed: 0
Skipped (no context or limit reached): 5
```

## Root Cause

The workflow currently triggers on `workflow_run` (after `PR Validation` completes). Under `workflow_run`, `github.event.pull_request` is `null`, so gh-aw's safe-output handlers configured with `target: triggering` cannot resolve a PR context for `create_pull_request_review_comment` or `submit_pull_request_review`. The custom resolver step exports `PR_NUMBER` / `PR_HEAD_SHA` for the agent prompt, but the safe-output handlers do not consume those env vars — they require an actual `pull_request` event payload.

This regression was introduced when the trigger was switched from `pull_request` to `workflow_run` in #580 to gate on PR Validation conclusion. The recent #584 (branches filter) and #586 (`bots:` allow-list for `dependabot[bot]`) fixes got the workflow firing and activating, but did not address the missing PR payload in the safe-output processing job.

## Proposed Fix

[...]

Issue #585, opened 2026-04-28
fix(workflows): allow dependabot[bot] to activate AW Dependabot PR Review via on.bots


katriendg:
## Summary

After the PR for #583 corrected the `branches:` filter, the AW Dependabot PR Review workflow finally fired on a Dependabot PR's `workflow_run` completion ([run 25045086278](https://github.com/microsoft/physical-ai-toolchain/actions/runs/25045086278)) — but `pre_activation` immediately skipped activation:

> ⏭️ Workflow Activation Skipped
> Access denied: User 'dependabot[bot]' is not authorized. Required permissions: admin, maintainer, write.
>
> User permission 'none' does not meet requirements: admin, maintainer, write

## Root Cause

gh-aw's pre-activation guard validates the **triggering actor's** repository role against `on.roles:` (default `[admin, maintainer, write]`). For `workflow_run` triggers fired by Dependabot, the effective actor is `dependabot[bot]`, which has role `none` — so activation is blocked even though the trigger filters and PR resolver are correct.

Bots must be allow-listed explicitly via `on.bots:` to bypass the role check, per the [gh-aw schema](https://github.com/github/gh-aw/blob/v0.68.3/.github/aw/github-agentic-workflows.md):

> `bots:` - Bot identifiers allowed to trigger workflow regardless of role permissions

## Fix

Add `on.bots:` (and keep `on.roles:` explicit at the default) in `.github/workflows/aw-dependabot-pr-review.md`:

```yaml
on:
  workflow_run:
    workflows: ["PR Validation"]
    types: [completed]
    branches:
      - "dependabot/**"
  bots: ["dependabot[bot]"]
  roles: [admin, maintainer, write]
```

The downstream `resolve-pr` step still gates on `pr.user.login === 'dependabot[bot]'`, so allow-listing the bot at the trigger boundary does not widen the surface beyond Dependabot.

## Acceptance Criteria

- [ ] `.github/workflows/aw-dependabot-pr-review.md` declares `on.bots: ["dependabot[bot]"]`.
- [ ] `gh aw compile aw-dependabot-pr-review --strict` reports 0 errors and 0 warnings.
- [ ] `npm run lint:yaml` passes.
- [ ] After merge, the next Dependabot `PR Validation` completion produces an `aw-dependabot-pr-review` run whose `pre_activation` job passes (no `User permission 'none'` annotation), and the agent posts an advisory review on the PR.

## References

[...]

Issue #583, opened 2026-04-28
fix(workflows): correct branches filter on AW Dependabot PR Review workflow_run trigger


katriendg:
## Summary

Following the trigger migration in PR #580, the AW Dependabot PR Review workflow has not fired for any Dependabot PR. Root cause is the `branches:` filter on the `workflow_run` trigger:

```yaml
on:
  workflow_run:
    workflows: ["PR Validation"]
    types: [completed]
    branches:
      - main
```

For `workflow_run` events, GitHub matches `branches` against the **`head_branch` of the triggering workflow run**, not the base branch. `PR Validation` always runs on the PR's head branch (e.g. `dependabot/docker/data-management/viewer/frontend/node-e49fd70`), so `branches: [main]` never matches.

## Evidence

- PR #560 (Dependabot Docker bump) — `PR Validation` completed successfully on 2026-04-28 07:50–07:56 UTC ([run 25040729782](https://github.com/microsoft/physical-ai-toolchain/actions/runs/25040729782)). PR #580 merged at 09:49 UTC the same day. No `workflow_run`-event invocation of `aw-dependabot-pr-review` exists for that PR or any later Dependabot PR.
- `gh run list --workflow=aw-dependabot-pr-review.lock.yml` shows zero runs with `event: workflow_run` since #580 merged.

## Fix

Replace the `main`-only filter with a `dependabot/**` glob so the trigger scopes correctly to Dependabot head branches across all ecosystems while still benefiting from `gh-aw`'s required branch-restriction guard:

```yaml
on:
  workflow_run:
    workflows: ["PR Validation"]
    types: [completed]
    branches:
      - "dependabot/**"
```

The existing `resolve-pr` step already gates on `pr.user.login === 'dependabot[bot]'`, so the broader trigger surface stays gated downstream.

## Acceptance Criteria

- [ ] `.github/workflows/aw-dependabot-pr-review.md` uses `branches: ["dependabot/**"]` on the `workflow_run` trigger.
- [ ] `gh aw compile aw-dependabot-pr-review` reports 0 errors and 0 warnings.
- [ ] `npm run lint:yaml` passes.
- [ ] After merge, the next Dependabot PR's `PR Validation` completion triggers an `aw-dependabot-pr-review` `workflow_run` event (verifiable via `gh run list`).
- [ ] An open Dependabot PR (e.g. #560) receives an advisory review after a `PR Validation` re-run (push empty commit or rebase to re-trigger).

## Out of scope

[...]

Issue #579, opened 2026-04-28
feat(workflows): harden the agentic Dependabot AW PR reviewer in experimental phase


katriendg:
## Summary

Track the experimental phase of the agentic Dependabot PR reviewer
(`.github/workflows/aw-dependabot-pr-review.md`) — an advisory-only reviewer
that enriches Dependabot PRs with GHSA / OSV / NVD intel and posts a single
`APPROVE` or `COMMENT` review. This issue captures the architectural
constraints that shaped the current design, why we are running it as
experimental, and the follow-ups required before we promote it to a
default-on workflow.

## Why "Experimental"

The reviewer is opt-in and advisory-only. We are intentionally running it as
experimental because three properties are still being validated in production
traffic:

1. **Sandbox-imposed validation model.** The agent's `bash:` tool runs inside
   the AWF firewall sandbox container (`ghcr.io/github/gh-aw-firewall/agent:0.25.20`).
   Only `${RUNNER_TEMP}/gh-aw` and the workspace are mounted; host toolchains
   installed by `runtimes:` (uv, python, go, node, terraform) are **not**
   visible inside the sandbox. The reviewer therefore cannot run `pytest`,
   `npm ci`, `terraform plan`, or `go build` itself. It must rely on
   deterministic CI signal plus static reasoning over manifests.
2. **Trigger semantics.** The workflow now fires on `workflow_run` after the
   `PR Validation` orchestrator completes on `main`. This guarantees the agent
   sees a final CI conclusion before reviewing, but it changes failure modes
   for fork PRs and grouped Dependabot updates that we have not yet observed
   at scale.
3. **Verdict discipline.** The agent emits `APPROVE` or `COMMENT` only —
   `REQUEST_CHANGES` is forbidden. The risk rubric, surface→check-run map,
   and verdict adjustment logic need real Dependabot traffic to confirm they
   produce useful banners without false-positive noise.

While experimental, maintainers retain full ownership of merge decisions; the
reviewer surfaces signal but never gates.

### Architecture

* **Source workflow:** `.github/workflows/aw-dependabot-pr-review.md`
* **Persona:** `.github/agents/dependabot-pr-reviewer.agent.md`
* **Compiled lock:** `.github/workflows/aw-dependabot-pr-review.lock.yml`
  (regenerate via `gh aw compile aw-dependabot-pr-review`)

### Trigger

`workflow_run` after `PR Validation` completes on `main`. A resolver step
[...]

PR #612, opened 2026-05-05
refactor(agents): migrate dependabot AW review to workflow_run trigger

katriendg:
## Description

The `aw-dependabot-pr-review` agentic workflow used to fire on `pull_request_target`, which meant the resolver step captured a snapshot of `PR Validation` while it was still `pending` or `in_progress:*`,  and the advisory review was posted before the orchestrator ever finished. PR #608 was the canonical example: the review correctly applied the Isaac Sim *numpy 2.x* ABI guard, but its CI banner quoted a stale `in_progress:in_progress` conclusion.

This PR migrates the workflow to `workflow_run` keyed on `PR Validation` *completed*, reads the orchestrator's terminal conclusion straight from `context.payload.workflow_run.conclusion`, and pre-resolves failing per-surface check-runs once in the resolver step. The persona rubric is rewritten to consume those env vars and to map every terminal conclusion explicitly - `pending` and `in_progress:*` branches are gone because they are now unreachable.

Related to #579.

## Type of Change

- [ ] 📚 Documentation update

## Component(s) Affected

- [ ] `docs/` - Documentation

### Workflow trigger and resolver

> Switching to `workflow_run` runs the agent step against the trusted, default-branch copy of the workflow, so the gh-aw compiler can auto-inject fork-PR exclusion and the `repository.id` guard.

- Replaced `pull_request_target` with `workflow_run` on **`workflows: ["PR Validation"]`**, `types: [completed]`, `branches: ["dependabot/**"]`. The `branches:` filter on `workflow_run` matches the *triggering run's `head_branch`* (not the base), so `dependabot/**` is the only value that fires for Dependabot PRs — using `main` here was the [#583](https://github.com/microsoft/physical-ai-toolchain/issues/583) regression fixed in [#584](https://github.com/microsoft/physical-ai-toolchain/pull/584). The workflow-level `if:` filters on `workflow_run.event == 'pull_request'`, `workflow_run.actor.login == 'dependabot[bot]'`, and a whitelist of seven terminal conclusions.
- Kept `on.bots: ["dependabot[bot]"]` and `on.roles: [admin, maintainer, write]` at the top level — gh-aw's `pre_activation` guard checks the triggering actor against `on.bots` / `on.roles` independently of the workflow `if:`, so dropping these would resurrect the [#585](https://github.com/microsoft/physical-ai-toolchain/issues/585) / [#586](https://github.com/microsoft/physical-ai-toolchain/pull/586) `User permission 'none'` activation block.
- Added `checks: read` to `permissions:` for server-side check-run enumeration; existing `contents`, `pull-requests`, and `actions` scopes are unchanged.
- Rewrote the *resolve-pr* step. It reads `context.payload.workflow_run`, prefers `workflow_run.pull_requests[0]`, and falls back to `search.issuesAndPullRequests` keyed on `head_sha` for the fork case. Both paths re-hydrate via `pulls.get` so `body` and `draft` are reliable.
- Dropped the previous `listWorkflowRunsForRepo` lookup. **`PR_VALIDATION_CONCLUSION`** now reads directly from `run.conclusion`, which under `types: [completed]` is always one of `success`, `failure`, `cancelled`, `timed_out`, `neutral`, `skipped`, or `action_required`.
- Added two new env vars exported by the resolver:
  - **`PR_VALIDATION_FAILING_CHECKS`** — JSON array of `{name, html_url, conclusion}` from `checks.listForRef(ref=pr.head.sha)` filtered to completed non-success/non-neutral/non-skipped runs.
  - **`PR_BODY`** — PR body hydrated server-side so the agent does not depend on the integrity-filtered MCP read of the PR.
- New skip reasons in `PR_DEPENDABOT_SKIP_REASON`: `not-a-pr-run` and `pr-resolution-failed`, alongside the existing `not-dependabot` / `draft`.
- Retargeted `safe-outputs`:
  - `submit-pull-request-review.target` → `${{ env.PR_NUMBER }}`
  - `add-comment.target` → `${{ env.PR_NUMBER }}` (was `triggering`, which is undefined under `workflow_run`)
  - `create-pull-request-review-comment.target` → `"*"`

### Persona verdict rubric

> The agent now reasons over a *final* CI signal, so the rubric collapses to a clean terminal-conclusion map.

- Rewrote the *Validation Signal* section in `.github/agents/dependabot-pr-reviewer.agent.md`. The persona is told the workflow runs after `PR Validation` reaches a terminal conclusion, and is explicitly forbidden from calling `checks.listForRef` or `commits/{sha}/check-runs` — it reads `PR_VALIDATION_FAILING_CHECKS` from the environment instead.
- Reframed the *Surface to Check Run Map* as an informational lookup for mapping a failing check name back to its dependency surface. The persona no longer walks it via the API.
- Rewrote the *Verdict Adjustment* block as an explicit terminal-conclusion map:
  - `success` + no static concern + no sticky high-risk trigger → `APPROVE`-eligible, citing the orchestrator conclusion plus an empty `PR_VALIDATION_FAILING_CHECKS`.
  - `failure | cancelled | timed_out | action_required` → `COMMENT`; body MUST quote every entry from `PR_VALIDATION_FAILING_CHECKS` (`name` plus `html_url`).
  - `neutral | skipped | unknown` or `PR_DEPENDABOT_SKIP_REASON == 'pr-resolution-failed'` → `COMMENT` with a `> [!CAUTION]` banner: *Deterministic CI signal unavailable (`{conclusion}`); review is advisory only.*
- Preserved the sticky Isaac Sim ABI guard verbatim — a `numpy` 2.x bump still keeps the verdict at `COMMENT` and forces the `⚠️ Maintainer review recommended` banner regardless of CI conclusion.

### Workflow documentation and lock files

- Rewrote the *Trigger Posture* and step-by-step prose in `aw-dependabot-pr-review.md` to describe the `workflow_run` execution model, the gh-aw compiler's auto-injected fork-PR exclusion and `repository.id` guard, and the new env-var contract.
- Bumped `github/gh-aw-actions/setup` `v0.68.3` → `v0.71.1` in `.github/aw/actions-lock.json` (SHA `ba90f21…` → `239aec4…`), picked up by recompilation.
- Regenerated `.github/workflows/aw-dependabot-pr-review.lock.yml` via the gh-aw compiler — diff reflects the trigger swap, the new env vars, and the setup-action SHA bump. No hand edits.

## Testing Performed

> None of the templated test surfaces apply — this PR only touches `.github/agents/` and `.github/workflows/`. Validation evidence: `npm run lint:md` and `npm run lint:yaml` pass on the changed files; the `aw-dependabot-pr-review.lock.yml` artifact is regenerated rather than hand-edited and matches the gh-aw compiler output for the new source. The behavioural change is observable on the next Dependabot PR — the advisory review will fire after `PR Validation` completes and quote the orchestrator's terminal conclusion plus any failing per-surface checks.

## Documentation Impact

## Bug Fix Checklist

*Not a bug fix — this is a refactor of an agentic-workflow trigger surface.*

## Checklist

## Related Issues

Related to #579

## Notes

> The `min-integrity: approved` setting on `tools.github` is intentionally preserved. The agent's MCP PR-body read is therefore filtered, which is why the resolver hydrates `PR_BODY` from the REST API server-side — the persona consumes the env var rather than relying on the filtered MCP payload.

- Lowering `min-integrity` to `unapproved` was rejected on prompt-injection grounds; the resolver-side hydration is the chosen mitigation.
- `workflow_run` runs in default-branch context, which means changes to the AW workflow itself cannot be exercised by a Dependabot PR — this is the secure-by-design tradeoff documented in the GitHub Security Lab "preventing pwn requests" guide and aligns with the gh-aw `workflow_run` recommendation.

## Follow-up Tasks

- Validate behaviour on a grouped Dependabot update that produces multiple `PR Validation` runs against the same head SHA — confirm that only the latest completed run drives the advisory review.
- After the first live Dependabot PR runs through the new trigger, compare the posted review's CI banner against the orchestrator's final conclusion and the failing-check list to confirm the staleness regression observed in PR #608 is gone.
- Confirm that `safe-outputs.submit-pull-request-review` and `add-comment` post successfully under `workflow_run` — the `target: ${{ env.PR_NUMBER }}` overrides are the [#588](https://github.com/microsoft/physical-ai-toolchain/issues/588) / [#589](https://github.com/microsoft/physical-ai-toolchain/pull/589) mitigation; a *Not in pull request context* skip in `safe_outputs` would mean the env var did not resolve.


.github/workflows/aw-dependabot-pr-review.md
```diff
@@ -64,16 +63,47 @@ steps:
...
+          core.exportVariable('PR_DEPENDABOT_SKIP_REASON', 'not-a-pr-run');
           return;
         }
+
+        let pr = (run.pull_requests || [])[0];
+        if (!pr) {
+          // Fork PRs do not appear in workflow_run.pull_requests; fall back to search.
+          const q = `repo:${context.repo.owner}/${context.repo.repo} is:pr sha:${run.head_sha}`;
+          const { data } = await github.rest.search.issuesAndPullRequests({ q });
+          const hit = (data.items || [])[0];
```
WilliamBerryiii: Thanks for adding the fork-PR fallback here — it's a thoughtful safety net. One thing we noticed: `items[0]` will pick whichever PR the search returns first, which can briefly be ambiguous when Dependabot opens several PRs in the same batch. Filtering to the unique match (e.g., `head.sha === run.head_sha && state === 'open' && user.login === 'dependabot[bot]'`) and failing loudly if more than one survives would make the resolver self-defending against that edge case.

.github/workflows/aw-dependabot-pr-review.md
```diff
@@ -82,38 +112,42 @@ steps:
...
-          const { data } = await github.rest.actions.listWorkflowRunsForRepo({
+          const { data: checks } = await github.rest.checks.listForRef({
             owner: context.repo.owner,
             repo: context.repo.repo,
-            head_sha: pr.head.sha,
-            event: 'pull_request',
-            per_page: 30,
+            ref: pr.head.sha,
+            per_page: 100,
           });
```
WilliamBerryiii: Small future-proofing thought: `checks.listForRef` here pages at 100 results, and as the matrix grows we could silently miss a check on a busy commit. Wrapping the call in `github.paginate(github.rest.checks.listForRef, { ... })` keeps the gate accurate without changing the surface contract.

.github/workflows/aw-dependabot-pr-review.md
```diff
@@ -82,38 +112,42 @@ steps:
...
         core.exportVariable('PR_AUTHOR', pr.user.login);
         core.exportVariable('PR_HEAD_SHA', pr.head.sha);
 
-        // Look up the most recent PR Validation workflow run for this head SHA.
-        // It may still be in progress when pull_request_target fires.
-        let conclusion = 'pending';
-        let runUrl = '';
+        // PR Validation conclusion comes directly from the triggering workflow_run payload;
+        // it is always final under `types: [completed]`.
+        core.exportVariable('PR_VALIDATION_CONCLUSION', run.conclusion || 'unknown');
```
WilliamBerryiii: Tiny cleanup nit: with the `workflow_run` trigger pinned to `types: [completed]`, `run.conclusion` is always set, so the `|| 'unknown'` fallback can never fire. Dropping it would make the contract a bit easier to reason about — happy to leave for a follow-up if it's not worth the churn here.

scripts/linting/Invoke-YamlLint.ps1
```diff
@@ -80,6 +80,9 @@ function Invoke-YamlLintCore {
         }
     }
 
+    # Exclude gh-aw compiler-generated lock files (machine-generated, not editable)
+    $filesToLint = @($filesToLint | Where-Object { $_ -notmatch '\.lock\.(yml|yaml)$' })
```
WilliamBerryiii, line 84: Nice catch including the post-filter — the regex form also handles `.yaml` cleanly. Since the inline exclusions in the `-Path` and `-ChangedFiles` branches now cover the same ground, we could pick one source of truth (the post-filter is the more future-proof of the two) and drop the inline ones, with a short comment like `# Lock files are gh-aw machine-generated; never lint them.` Pure refactor, no behavioral change — totally fine to defer.

WilliamBerryiii (commented):
Thanks @katriendg — really nice migration. Pulling the trigger, resolver, and dispatch logic into a clean `workflow_run` shape makes the Dependabot review path feel a lot more durable, and the explicit no-`checks.listForRef`-in-resolver guard is a great touch.

A couple of thoughts that aren't blocking:

- **Concurrency group scope** (`.lock.yml`) — `gh-aw-copilot-${{ github.workflow }}` serializes every Dependabot review, so a batch of ~10 PRs would stack end-to-end (~10× wall time). Scoping per-PR (e.g., `gh-aw-copilot-${{ github.workflow }}-${{ github.event.workflow_run.pull_requests[0].number || github.event.workflow_run.head_sha }}`) would let independent reviews run in parallel while still serializing within a single PR.
- **Dead "Checkout PR branch" step** (`.lock.yml`) — under `workflow_run`, neither `github.event.pull_request` nor `github.event.issue.pull_request` is set, so that step is permanently skipped. That's intentional (the resolver pulls everything from REST), but a one-line note in the source `.md` "Trigger Posture" section — something like "The agent runs without a working tree — all PR context comes from REST APIs in the resolver. Do not add a checkout step." — would save the next reader from trying to "fix" the missing checkout.

A few other items we looked at and don't think need action: the XPIA prompt-injection guidance reads correctly under the new trigger; the surface-to-check-run name mapping in the resolver matches the matrix names we could verify; and the `> [!CAUTION]` callout is the project's preferred alert syntax.

Thanks again for moving this forward!

bindsi (approved):
Nice Katrien!!!

scripts/linting/Invoke-YamlLint.ps1
katriendg: good catch, cleaned up

.github/workflows/aw-dependabot-pr-review.md
katriendg: Drop dead || 'unknown' fallback — under types: [completed], run.conclusion is always set

.github/workflows/aw-dependabot-pr-review.md
katriendg: github.paginate() for full enumeration

katriendg:
> Thanks @katriendg — really nice migration. Pulling the trigger, resolver, and dispatch logic into a clean `workflow_run` shape makes the Dependabot review path feel a lot more durable, and the explicit no-`checks.listForRef`-in-resolver guard is a great touch.
> 
> A couple of thoughts that aren't blocking:
> 
> * **Concurrency group scope** (`.lock.yml`) — `gh-aw-copilot-${{ github.workflow }}` serializes every Dependabot review, so a batch of ~10 PRs would stack end-to-end (~10× wall time). Scoping per-PR (e.g., `gh-aw-copilot-${{ github.workflow }}-${{ github.event.workflow_run.pull_requests[0].number || github.event.workflow_run.head_sha }}`) would let independent reviews run in parallel while still serializing within a single PR.
> * **Dead "Checkout PR branch" step** (`.lock.yml`) — under `workflow_run`, neither `github.event.pull_request` nor `github.event.issue.pull_request` is set, so that step is permanently skipped. That's intentional (the resolver pulls everything from REST), but a one-line note in the source `.md` "Trigger Posture" section — something like "The agent runs without a working tree — all PR context comes from REST APIs in the resolver. Do not add a checkout step." — would save the next reader from trying to "fix" the missing checkout.
> 
> A few other items we looked at and don't think need action: the XPIA prompt-injection guidance reads correctly under the new trigger; the surface-to-check-run name mapping in the resolver matches the matrix names we could verify; and the `> [!CAUTION]` callout is the project's preferred alert syntax.
> 
> Thanks again for moving this forward!

Thanks for your review @WilliamBerryiii - I believe I have addressed all comments, and by digging more into the docs, the `concurrency group` scope should also be addressed. 
`concurrency` is a first-class frontmatter key. The docs show both workflow-level concurrency.group and concurrency.job-discriminator for fan-out patterns. For the workflow_run trigger, the default compiler behavior maps to Schedule/Other → gh-aw-${{ github.workflow }} (no PR scoping), which is exactly the serialization you flagged.

For the second comment, added a note to the MD file.

Ready for a second pass review. Thanks.

rezatnoMsirhC (approved):

PR #677, opened 2026-05-11
feat(training): add optional Azure ML mirror for OSMO training runs

### LINKED ISSUE #668: feat(deploy): Optional Azure ML mirror for OSMO training runs
### Component

OSMO Control Plane

### Problem Statement

OSMO-submitted training runs (IL and RL) write checkpoints and tensorboard logs to the cluster's PVC, but there is no first-class path to land them in an Azure ML workspace as governed model versions. Operators who want a versioned, shareable home for trained policies — or want to feed AzureML-based deployment gating — have to either rely on the in-process registration paths inside training (which already double-register in IL and are coupled to training success) or hand-roll an upload after the fact. There is also no way to retro-mirror an older run whose registration failed or never happened.

### Proposed Solution

Add an opt-in Azure ML mirror that lives **off the training critical path**:

* **Deploy-time toggle** in `infrastructure/setup/04-deploy-osmo-backend.sh` (`--skip-azureml-pod-template`) that injects `AZURE_SUBSCRIPTION_ID`, `AZURE_RESOURCE_GROUP`, `AZUREML_WORKSPACE_NAME` into the existing `default_user` OSMO pod template. Auto-skips when no AzureML workspace appears in Terraform outputs.
* **Standalone mirror script** at `training/utils/aml_mirror.py` — framework-agnostic, authenticates via `DefaultAzureCredential` (Workload Identity), uploads tensorboard + filtered final checkpoint, registers a new model version via `mlflow.register_model()`.
* **Replay workflow** at `workflows/osmo/replay-azureml.yaml` plus submission helper `training/utils/replay-azureml.sh` so any completed run can be replayed by `RUN_ID` without modifying training workflows.
* **Documentation** as a single "Azure ML Mirror (Optional)" section in `infrastructure/setup/README.md` covering When/Prerequisites/Enabling/Using/Disabling/Troubleshooting.

No existing IL or RL training workflow YAML is modified. Mirror failure cannot break a training run.

### Alternatives Considered

* **Inline mirror in every training workflow YAML** — rejected: would create a 4th AzureML registration path on top of the three existing in-process paths in IL, producing ~12 model versions per 5-checkpoint run, and would force every workflow author to opt out.
* **New `azureml_config` pod template added to `override_pod_template`** — deferred: OSMO container env-merge semantics across multiple `override_pod_template` entries are not verified in this codebase. Extending `default_user` directly sidesteps the unknown.
* **Bake `aml_mirror.py` into a utility container image** — deferred: cleaner long-term but requires an image build pipeline. Use OSMO `files:` distribution first.
* **New Terraform `should_enable_azureml_mirror` variable** — rejected: no `should_*` precedent in deploy scripts; `tf_get` auto-detection on the existing AzureML output is sufficient and matches repo convention.
* **Add `osmo_aml_storage` role assignment in Terraform** — rejected: AzureML workspace's backing storage IS the main storage account; the existing `Storage Blob Data Contributor` role already covers it.

### Additional Context

**Prerequisites already in place:**

* OSMO managed identity already has `AzureML Data Scientist` and `Storage Blob Data Contributor` on the workspace's backing storage (verified in `infrastructure/terraform/modules/platform/role-assignments.tf`).
* `azureml_workspace.value.{name,id,workspace_id}` already exposed at root Terraform output level.
* Federated credential for `system:serviceaccount:<workflows-ns>:osmo-workflow` already present.

katriendg:
## Description

Adds an opt-in post-training mirror that replays completed OSMO training runs into Azure ML, plus deploy-time automation for the existing live-tracking integration.

&gt; The existing RL and IL training pipelines already log metrics, artifacts, and register checkpoints in Azure ML via `bootstrap_azure_ml()` during training. This PR adds a **post-hoc replay** path for runs where live MLflow logging was unavailable or incomplete, and hardens the deploy-time configuration.

### Deploy-time pod-template extension

The OSMO backend deploy script gained a `--skip-azureml-pod-template` flag (with `SKIP_AZUREML_POD_TEMPLATE` env var override) that gates injection of three AzureML environment variables into the *default_user* pod template. The feature **auto-detects** the workspace from Terraform outputs via `tf_get` and gracefully skips when no AzureML workspace exists or the flag is set.

- Extended *04-deploy-osmo-backend.sh* with the flag, argument parser case, config-preview status line, and a post-render `jq` step that injects `AZURE_SUBSCRIPTION_ID`, `AZURE_RESOURCE_GROUP`, and `AZUREML_WORKSPACE_NAME` into the rendered pod-template output JSON
- `az account show` is deferred until the `else` branch to avoid failures when AzureML is skipped

### Mirror script and replay workflow

Created a **standalone post-training uploader** (*training/utils/aml_mirror.py*) that authenticates via `DefaultAzureCredential`, uploads tensorboard logs and a filtered final checkpoint, and registers a new model version via `mlflow.register_model()`. Optimizer, scheduler, and RNG state files are filtered out to reduce upload size.

- Created *workflows/osmo/replay-azureml.yaml* — an OSMO workflow that distributes `aml_mirror.py` via base64-encoded env vars and installs runtime dependencies at execution time
- Created *training/utils/replay-azureml.sh* — a submission helper that sources `scripts/lib/common.sh` and `scripts/lib/terraform-outputs.sh`, then submits the replay workflow via `osmo workflow submit`; uses portable `base64 | tr -d '\n'` encoding
- Created *workflows/osmo/requirements-aml-mirror.txt* — pinned dependencies for the mirror runtime (resolved independently from `training/rl/pyproject.toml`; ceiling documented)

### MLflow registry URI handling

The IL bootstrap path (`training/il/scripts/lerobot/bootstrap.py`) sets `MLFLOW_REGISTRY_URI` via `os.environ.setdefault()` before importing `azure.ai.ml`, preventing registry validation failures during `MLClient.__init__`. The RL path does not require this workaround. No workflow YAML changes are needed — existing IL and RL workflows are unmodified by this PR.

Additional changes:

- `training/rl/pyproject.toml` — added `azureml-mlflow==1.62.0.post2` + `jsonpickle==4.1.1` (registers `azureml://` URI handler, authenticates via `DefaultAzureCredential`)
- `training/rl/scripts/skrl_training.py` — artifact upload wrapped in try/except with `exc_info=True` (non-fatal, preserves traceback)
- `training/rl/scripts/smoke_test_azure.py` — `log_dict` wrapped for graceful degradation with `exc_info=True`
- `training/utils/context.py` — documented that azureml-mlflow plugin handles auth

### Tests

- Created *training/tests/test_aml_mirror.py* — unit tests for the mirror utility (100% coverage, pandas-free mocks)
- Extended *training/tests/test_skrl_training.py* — coverage for artifact upload error handling
- Extended *training/tests/test_smoke_test_azure.py* — coverage for `log_dict` fallback

### Documentation

- Added a **☁️ Azure ML Mirror (Optional)** section to *infrastructure/setup/README.md* covering prerequisites, enabling, using, component table, disabling after deploy, and troubleshooting
- Added a `&gt; [!TIP]` pointer in *training/README.md* directing readers to the mirror documentation

### Security

- Bumped `pyarrow` from 19.0.1 to 23.0.1 in mirror requirements (CVE-2026-25087)
- Added two mlflow CVE allowlist entries to `dependency-review.yml` (GHSA-46r5-x6jq-v8g6, GHSA-fh64-r2vc-xvhr) — server-side only vulnerabilities, repo uses client only; rationale documented inline
- Removed five stale GHSA allowlist entries for lodash (resolved to 4.18.1) and picomatch (resolved to 2.3.2/4.0.4)

### Ancillary

- Deduplicated and updated cspell dictionaries with new terms
- Updated `training/rl/pyproject.toml` with dependency alignment
- Minor table formatting in `docs/README.md`, `data-management/viewer/README.md`, and `infrastructure/terraform/modules/conversion-pipeline/README.md`

Validated end-to-end: OSMO workflow completed with PPO training (50 iterations), MLflow metrics logged, artifacts uploaded (HTTP 201), experiment visible in Azure ML Studio.

Closes #668

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation

## Testing Performed

- [x] Smoke tests passed (`smoke_test_azure.py`)
- [x] Unit tests pass (`test_aml_mirror.py`, `test_skrl_training.py`, `test_smoke_test_azure.py`)

## Documentation Impact

## Checklist


training/rl/requirements.txt
```diff
@@ -622,3 +622,5 @@ yarl==1.23.0
     # via aiohttp
 zipp==3.23.1
     # via importlib-metadata
+azureml-mlflow==1.62.0.post2
```
nguyena2, line 630: Looks like these two libraries are added manually. Add them to the pyproject.toml to ensure they are captured in future generations

katriendg: Good catch! This was edited by the agent, fixed and had to downgrade another library for compat with azure-mlflow.

akhanattentive (approved):

infrastructure/setup/04-deploy-osmo-backend.sh
```diff
@@ -406,6 +410,30 @@ export ACR_LOGIN_SERVER="$acr_login_server"
 # Render shared pod template (substitutes WORKFLOW_SERVICE_ACCOUNT)
 envsubst < "$pod_template_file" > "$CONFIG_DIR/out/pod-template-config.json"
 
+aml_workspace_name=$(tf_get "$tf_output" "azureml_workspace.value.name")
+aml_resource_group=$(tf_get "$tf_output" "resource_group.value.name")
+aml_subscription_id=$(az account show --query id -o tsv)
+
```
rezatnoMsirhC: Small bug: `az account show` runs unconditionally before the `--skip-azureml-pod-template` guard. With `set -euo pipefail`, an expired `az login` session will cause the script to exit before reaching the skip branch, so a caller opting out of AzureML setup still gets an unexpected failure.

Moving all three assignments inside the `else` branch (where they are actually needed) avoids this:

```bash
if [[ "$skip_azureml_pod_template" == "true" ]]; then
  info "AzureML pod-template extension explicitly skipped (--skip-azureml-pod-template)"
else
  aml_workspace_name=$(tf_get "$tf_output" "azureml_workspace.value.name")
  if [[ -z "$aml_workspace_name" ]]; then
    info "No AzureML workspace in Terraform outputs; skipping pod-template extension"
  else
    aml_resource_group=$(tf_get "$tf_output" "resource_group.value.name")
    aml_subscription_id=$(az account show --query id -o tsv)
    section "Configure Azure ML Pod Template (Optional)"
    ...
  fi
fi
```

training/rl/scripts/skrl_training.py
```diff
@@ -682,7 +682,11 @@ def _finalize_mlflow_run(state: MLflowRunState) -> None:
     mlflow = state.mlflow
     mlflow.set_tag("training_outcome", state.outcome)
 
-    latest_checkpoint_uri = _log_artifacts(mlflow, state.log_dir, state.resume_path)
+    try:
+        latest_checkpoint_uri = _log_artifacts(mlflow, state.log_dir, state.resume_path)
+    except Exception:
+        _LOGGER.warning("MLflow artifact upload failed; checkpoint registration will be skipped")
+        latest_checkpoint_uri = None
```
rezatnoMsirhC: The exception is caught without `exc_info=True`, so the traceback is discarded. When artifact upload fails in production there is no diagnostic signal about the root cause. Adding `exc_info=True` preserves the graceful-degradation intent while keeping the traceback in the logs:

```python
except Exception:
    _LOGGER.warning("MLflow artifact upload failed; checkpoint registration will be skipped", exc_info=True)
    latest_checkpoint_uri = None
```

training/rl/scripts/smoke_test_azure.py
```diff
@@ -232,7 +232,10 @@ def _start_run(
             context.storage.container_name if context.storage else "not-configured",
         )
         mlflow_module.log_metric(args.metric_name, 1.0)
-        mlflow_module.log_dict(summary, "smoke-test-summary.json")
+        try:
+            mlflow_module.log_dict(summary, "smoke-test-summary.json")
+        except Exception:
+            _LOGGER.warning("Artifact logging unavailable (MLflow 3.x without azureml-mlflow plugin)")
```
rezatnoMsirhC: Same pattern as in `skrl_training.py`: exception swallowed without `exc_info=True`, which discards the traceback. Suggest:

```python
except Exception:
    _LOGGER.warning("Artifact logging unavailable (MLflow 3.x without azureml-mlflow plugin)", exc_info=True)
```

training/utils/aml_mirror.py
```diff
@@ -0,0 +1,160 @@
...
+import os
+import pathlib
+import shutil
+import sys
+import tempfile
+
+os.environ.setdefault("AZUREML_ARTIFACTS_DEFAULT_TIMEOUT", "7200")
+
+import mlflow
+from azure.ai.ml import MLClient
```
rezatnoMsirhC, line 26: `os.environ.setdefault(...)` at module level runs at import time, mutating the process environment as a side effect of importing the module. Tests that verify missing-env-var handling or expect a clean environment will have this value injected silently.

Moving the call inside `main()`, immediately before the Azure SDK setup that consumes it, keeps the side effect bounded and intentional:

```python
def main() -> int:
    missing = [k for k in REQUIRED_ENV if not os.environ.get(k)]
    if missing:
        print(f"aml_mirror: missing env vars: {missing}", file=sys.stderr)
        return 1

    os.environ.setdefault("AZUREML_ARTIFACTS_DEFAULT_TIMEOUT", "7200")
    ...
```

workflows/osmo/requirements-aml-mirror.txt
```diff
@@ -0,0 +1,504 @@
...
+alembic==1.18.4
+    # via mlflow
+annotated-doc==0.0.4
+    # via fastapi
+annotated-types==0.7.0
+    # via pydantic
+anyio==4.13.0
+    # via starlette
+asgiref==3.11.1
+    # via opentelemetry-instrumentation-asgi
```
rezatnoMsirhC, line 26: `azure-ai-ml==1.32.0` and `azureml-mlflow==1.60.0` here diverge from the versions in `training/rl/pyproject.toml` (`1.33.0` and `1.62.0.post2`). Since the mirror runs in an isolated runtime this is expected, but there is nothing documenting why the versions differ or what the ceiling constraint is. A brief note alongside the existing `uv pip compile` header would make the divergence intentional and visible as these versions continue to drift independently.

rezatnoMsirhC (approved):

bindsi (approved):
Automated review: Feature adds optional Azure ML mirror for OSMO training runs. Already has 2 approvals from team members. No blocking issues found on scan. LGTM.

training/utils/aml_mirror.py
```diff
@@ -0,0 +1,160 @@
...
+            metric_count = _log_tensorboard_metrics(tb_dir)
+            print(f"logged tensorboard from {tb_dir} ({metric_count} metric points)")
+
+        ckpts = sorted(
+            glob.glob(str(output_dir / "checkpoint-*")),
+            key=lambda p: int(p.rsplit("-", 1)[-1]) if p.rsplit("-", 1)[-1].isdigit() else -1,
+        )
+        if not ckpts:
+            print("aml_mirror: no checkpoint-* dirs found", file=sys.stderr)
+            return 1
```
bindsi, line 118: Non-blocking: returning `1` from inside `with mlflow.start_run(...)` exits the context normally, so MLflow can record the replay run as finished even though the script failed because no checkpoints were found. Please mark the run failed explicitly or raise an exception before exiting so the mirror/replay status reflects the failure.

training/utils/replay-azureml.sh
```diff
@@ -0,0 +1,57 @@
...
+aml_mirror_requirements="$REPO_ROOT/workflows/osmo/requirements-aml-mirror.txt"
+if [[ ! -f "$aml_mirror_requirements" ]]; then
+  fatal "requirements-aml-mirror.txt not found at $aml_mirror_requirements"
+fi
+
+section "Submit Azure ML Replay"
+print_kv "Run ID"     "$run_id"
+print_kv "Model name" "$model_name"
+
+aml_mirror_b64=$(base64 -w0 < "$aml_mirror_script")
```
bindsi: Non-blocking: `base64 -w0` is GNU-specific and fails on macOS/BSD `base64`, which makes this replay helper unusable for many local developers. A portable form is `base64 < "$file" | tr -d '\n'` or a small wrapper that handles both GNU and BSD flags.

bindsi (commented):
Automated batch review: two non-blocking reliability/portability issues noted.

training/utils/replay-azureml.sh
katriendg: Addressed in 90bda1c — replaced `base64 -w0` with the portable `base64 < "$file" | tr -d '\n'` form.

workflows/osmo/requirements-aml-mirror.txt
katriendg: Addressed in 90bda1c — added a comment block explaining that the mirror runtime installs into a fresh venv on OSMO pods and versions are resolved independently, with the ceiling constraint (`azureml-mlflow`'s `mlflow<3.0.0` pin) documented.

training/utils/aml_mirror.py
katriendg: Addressed in 90bda1c — added `mlflow.end_run(status="FAILED")` before returning 1 so the replay run status reflects the failure.

training/rl/scripts/smoke_test_azure.py
katriendg: Addressed in 90bda1c — added `exc_info=True` to match the `skrl_training.py` fix.

training/utils/aml_mirror.py
katriendg: Addressed in 90bda1c — moved `os.environ.setdefault("AZUREML_ARTIFACTS_DEFAULT_TIMEOUT", "7200")` inside `main()`, right after the env-var validation check.

training/rl/scripts/skrl_training.py
katriendg: Addressed in 90bda1c — added `exc_info=True` to preserve the traceback.

infrastructure/setup/04-deploy-osmo-backend.sh
katriendg: Addressed in 90bda1c — moved `aml_resource_group` and `aml_subscription_id` (including `az account show`) inside the `else` branch so they only run when a workspace is detected and the skip flag is not set.

bindsi (commented):
Automated batch re-review: prior bindsi findings appear fixed; three non-blocking follow-ups remain.

1. `workflows/osmo/replay-azureml.yaml` sets Azure environment variables from empty defaults, which can override pod-template-injected values. Omit empty overrides or require non-empty submit values.
2. The PR body says #668 does not modify existing IL/RL workflow YAML, but this PR does modify existing training workflows. Please update or split the scope.
3. `.github/workflows/dependency-review.yml` adds GHSA allowlist entries without adjacent rationale. Please document why each allowlist entry is safe or remove them.

katriendg:
> Automated batch re-review: prior bindsi findings appear fixed; three non-blocking follow-ups remain.
> 
> 1. `workflows/osmo/replay-azureml.yaml` sets Azure environment variables from empty defaults, which can override pod-template-injected values. Omit empty overrides or require non-empty submit values.
> 2. The PR body says #668 does not modify existing IL/RL workflow YAML, but this PR does modify existing training workflows. Please update or split the scope.
> 3. `.github/workflows/dependency-review.yml` adds GHSA allowlist entries without adjacent rationale. Please document why each allowlist entry is safe or remove them.

Good callouts, since rebasing I have now addressed all above. Thanks

PR #691, opened 2026-05-14
fix(ci): repair ERE regex in pr-validation path filter

### LINKED ISSUE #688: fix(ci): broken ERE regex in pr-validation path filter silently skips Pester tests and partial fuzz tests
## Summary

The path filter in `pr-validation.yml` uses double-backslash (`\\`) inside single-quoted bash strings passed to `grep -qE`. In GNU ERE, `\\` matches a literal backslash character — not an escaped dot. Since Unix-style file paths never contain a backslash before a file extension, the affected filter conditions never evaluate to `true`.

This was introduced in #520 (commit `357e7451`, 2026-04-22, "chore(build): redesign codecov flags and split pytest CI per component").

## Affected Filters

**`pester` — completely broken**

```bash
echo "pester=$(match '(\\.ps1|\\.psm1|\\.psd1)$')"
```

The pattern `\\.ps1` in ERE requires a literal `\` character before the extension. No Unix path satisfies this. The `pester-tests` job has been silently skipped for every PS1-touching PR since #520 was merged.

**`fuzz` — partially broken**

```bash
echo "fuzz=$(match '^(tests/fuzz|tests/fuzz_harness\\.py|tests/generate_fuzz_corpus\\.py|data-management/viewer/backend/src/|training/)')"
```

The directory-level anchors (`^tests/fuzz`, `^training/`, `^data-management/viewer/backend/src/`) still match correctly. The two standalone file paths fail:

| Path | Expected | Actual |
|------|----------|--------|
| `tests/fuzz_harness.py` | `true` | `false` |
| `tests/generate_fuzz_corpus.py` | `true` | `false` |

## Verification

```bash
$ echo "scripts/linting/Foo.ps1" | grep -qE '(\\.ps1|\\.psm1|\\.psd1)$' && echo MATCHED || echo NOT MATCHED
NOT MATCHED

$ echo "scripts/linting/Foo.ps1" | grep -qE '(\.ps1|\.psm1|\.psd1)$' && echo MATCHED || echo NOT MATCHED
MATCHED

$ echo "tests/fuzz_harness.py" | grep -qE '^(tests/fuzz|tests/fuzz_harness\\.py|tests/generate_fuzz_corpus\\.py|data-management/viewer/backend/src/|training/)' && echo MATCHED || echo NOT MATCHED
NOT MATCHED
```

## Proposed Fix

In `.github/workflows/pr-validation.yml`, replace `\\` with `\` in both affected patterns:

```bash
# Before
echo "fuzz=$(match '^(tests/fuzz|tests/fuzz_harness\\.py|tests/generate_fuzz_corpus\\.py|data-management/viewer/backend/src/|training/)')"
echo "pester=$(match '(\\.ps1|\\.psm1|\\.psd1)$')"

# After
echo "fuzz=$(match '^(tests/fuzz|tests/fuzz_harness\.py|tests/generate_fuzz_corpus\.py|data-management/viewer/backend/src/|training/)')"
echo "pester=$(match '(\.ps1|\.psm1|\.psd1)$')"
```

## Labels

`bug`, `ci`

ammmanism:
This PR fixes a bug in the PR Validation workflow's path filter logic that was causing the Fuzz Regression tests to be silently skipped. 

The `grep -qE` match pattern for the `fuzz` tests contained double-escaped backslashes (`\\`) inside a single-quoted string. Because `\ ` is the escape character in Extended Regular Expressions (ERE), the double backslash was evaluated as a literal backslash, causing the dot to lose its literal meaning (e.g., matching `tests/fuzz_harness\Xpy` instead of `tests/fuzz_harness.py`). 

This has been corrected to use a single backslash (`\.`), ensuring changes to fuzzing harnesses and corpuses correctly trigger the required test suite.

Closes #688.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Build/CI configuration change

## Component(s) Affected

- [x] CI/CD Workflows (`.github/workflows/pr-validation.yml`)

## Bug Fix Checklist

- [x] Root cause identified and addressed — double-escaped backslashes in the `grep -qE` matching block removed.
- [x] My commit messages follow Conventional Commits format and include the DCO `Signed-off-by` trailer.


WilliamBerryiii (approved):

WilliamBerryiii:
Thank you for this fix, @ammmanism! Your correction to the `pr-validation.yml` path filter restores the `fuzz` ERE pattern so changes under `tests/generate_fuzz_corpus.py` (and the broader fuzz surface) reliably trigger the regression suite again. The clear root-cause write-up in the PR description made verification quick — we appreciate the care you put into both the fix and the explanation.

Issue #770, opened 2026-05-26
track: 5-PR security and dependency refresh (scorecard + OSV + torch + dependabot)


WilliamBerryiii:
# Track: 5-PR security and dependency refresh

Umbrella tracking issue for a coordinated security and dependency-maintenance effort spanning three driver plans:

- `.copilot-tracking/plans/2026-05-22/scorecard-vulnerabilities-remediation-plan.instructions.md`
- `.copilot-tracking/plans/2026-05-23/osv-scanner-and-lerobot-torch-plan.instructions.md`
- `.copilot-tracking/plans/2026-05-25/dependabot-coverage-alignment-plan.instructions.md`

## Relationship to #712

The original umbrella #712 was scoped to a narrower remediation (gitpython/mako lockfiles, Docusaurus refresh) and has since been closed. Implementation expanded across the three plans above into a broader set of changes (OSV scanner adoption, torch pin coordination, RL resolver alignment, dependabot manifest refresh). This new tracking issue reflects the actual scope being delivered.

## Sub-PR breakdown

PRs are sized for atomic review by audience and to keep each change independently revertable.

| # | Title | Reviewer audience | Rationale |
|---|-------|-------------------|-----------|
| 1 | `build(scripts): adopt OSV-Scanner v2.3.8 for vulnerability scanning` | Security / CI | New tooling adoption; touches install scripts, CI workflows, contributing docs |
| 2 | `security(deps): bump qs and pin uuid in Docusaurus to remediate CVEs` | Frontend / docs | Isolated Docusaurus dependency bumps with regenerated lockfile |
| 3 | `security(deps): pin torch 2.11.0 across IL training and evaluation` | ML / training | Cross-cutting torch coordination requires resolver override + CI image + OSMO base image alignment |
| 4 | `security(deps): align RL resolver pins and SIL eval container deps` | Python deps | RL resolver pins (marshmallow, cryptography) + SIL container `av` downgrade |
| 5 | `build(deps): refresh dependabot manifest and root uv constraints` | Infra / CI | Dependabot reconfiguration (pip→uv, +2 ecosystems, torch ignore re-anchor) and root constraint refresh |

## Scope verification

A pre-issue deep review confirmed:

- All 27 remaining working-tree changes are allocated to one of the 5 PRs.
- Cruft removed before sub-issue creation: `temp-artifacts/` (ad-hoc helper + stale log) and root `requirements.txt` (empty `uv pip compile` output; historically removed by PR #447).
[...]

PR #778, opened 2026-05-26
feat(training): add single-node multi-GPU lerobot training on AzureML

algattik:
# Pull Request

## Description

Adds single-node multi-GPU support for LeRobot imitation-learning training on AzureML, following the upstream Hugging Face [multi-GPU training guide](https://huggingface.co/docs/lerobot/multi_gpu_training).

GPU count is detected at runtime from `torch.cuda.device_count()` — i.e., from the `nvidia.com/gpu: N` request of the AzureML-on-AKS `InstanceType` the pod runs on — so no extra CLI flag is required. When more than one GPU is visible, the entry point auto-wraps the LeRobot trainer with `accelerate launch`. Single-GPU and CPU runs are unchanged.

Closes #

## Type of Change

- [x] 📚 Documentation update
- [x] 🏗️ Infrastructure change (Terraform/IaC) — new `InstanceType` manifests

## Component(s) Affected

- [x] `infrastructure/setup/` - OSMO control plane / Helm — new GPU `InstanceType` manifests
- [x] `docs/` - Documentation

## What changed

- `training/il/scripts/lerobot/train.py`: detect GPU count from `torch.cuda.device_count()`; when `>1`, wrap the trainer with `accelerate launch --num-processes=N --mixed_precision=...`; strip `--use_amp` under `accelerate` to avoid double-AMP; install a kill-process-group `SIGTERM/SIGINT` handler so child processes terminate cleanly.
- `training/il/scripts/submit-azureml-lerobot-training.sh`: add `--mixed-precision` flag (`no|fp16|bf16`) with validation; document multi-GPU usage in the help block.
- `training/il/scripts/lerobot/azureml-train-entry.sh`: install lockfile with `--no-cache-dir --no-deps` so the resolved `requirements.txt` is trusted (matches `training/rl/scripts/train.sh`); the `[tool.uv] override-dependencies` block is honored at compile time only and would otherwise be re-resolved away on install.
- `infrastructure/setup/manifests/azureml-instance-types.yaml`: add 1/2/4-GPU `InstanceType` CRDs sized for typical SKUs.
- `training/il/lerobot/pyproject.toml`, `training/il/lerobot/requirements.txt`: pin `accelerate==1.13.0` and add `[tool.uv] override-dependencies` for `azure-storage-blob`, `diffusers`, `mlflow-skinny`, `numpy`, `packaging`, `torchcodec`; regenerate the lockfile.
- Workflow YAMLs (`training/il/workflows/azureml/lerobot-train.yaml`, `training/il/workflows/osmo/lerobot-train.yaml`, `evaluation/sil/workflows/osmo/lerobot-eval.yaml`): plumb new env vars; align install commands with the new `--no-deps` pattern.
- `docs/training/azureml-training.md`: document multi-GPU submission and how `InstanceType` GPU count drives the trainer.
- `.cspell/general-technical.txt`: add `accelerate`-related terms.
- `training/tests/test_lerobot_train.py`: cover GPU-count detection, accelerate wrapping, `--use_amp` stripping, mixed-precision pass-through, and signal-handler installation.

## Testing Performed

- [ ] Terraform `plan` reviewed (no unexpected changes) — N/A
- [ ] Terraform `apply` tested in dev environment — N/A
- [ ] Training scripts tested locally with Isaac Sim — N/A (IL, not RL)
- [ ] OSMO workflow submitted successfully — not yet
- [ ] Smoke tests passed (`smoke_test_azure.py`) — N/A
- [x] `pytest training/tests/test_lerobot_train.py` — green (29 tests)
- [x] `ruff check training/`
- [x] `shellcheck` on the modified entry/submit scripts
- [x] `npm run lint:md` and `npm run spell-check`

End-to-end AzureML submission against a real multi-GPU `InstanceType` performed. Metrics show high GPU utilization rate for GPU 0 and GPU 1.

## Documentation Impact

## Bug Fix Checklist

*Not applicable — feature PR.*

## Checklist

🚀 - Generated by Copilot


training/il/scripts/lerobot/train.py
rezatnoMsirhC, line 1: `--policy.use_amp` is a draccus boolean field. Draccus booleans only accept the bare flag form (`--policy.use_amp`, sets True) or the equals form (`--policy.use_amp=true/false`). They never emit a space-separated value token. The current `_strip_use_amp` logic unconditionally consumes the next token when it encounters the bare form, so a list like `["--policy.use_amp", "--steps=10000"]` would silently strip `--steps=10000` from the command.

Suggested fix: since the bare form never carries a value, simply advance past the flag itself.

```python
if arg == "--policy.use_amp":
    skipped = True
    i += 1
    continue
```

The test `test_strip_use_amp_bare_with_value` exercises `["--policy.use_amp", "true", ...]`, which is a form draccus cannot emit. Replacing it with a realistic case would close the gap:

```python
def test_strip_use_amp_bare_followed_by_flag(self):
    cleaned = _MOD._strip_use_amp(["--policy.use_amp", "--steps=10000"])
    assert cleaned == ["--steps=10000"]
```

infrastructure/setup/manifests/azureml-instance-types.yaml
rezatnoMsirhC, line 1: The PR description's "What changed" section states: "add 1/2/4/8-GPU InstanceType CRDs sized for typical SKUs." This manifest adds `gpu2`, `gpuspot2`, `gpu4`, and `gpuspot4`, but no `gpu8` or `gpuspot8`. If 8-GPU types were intentionally omitted (e.g., no supported SKU is available in the target environment), the PR description should be updated to say 1/2/4-GPU.

training/il/lerobot/requirements.txt
rezatnoMsirhC, line 1: Regenerating the lockfile with `--python-platform manylinux_2_28_x86_64` produces some notable downgrades: `torch` 2.12.0 to 2.10.0, `cryptography` 48.0.0 to 46.0.7, `wandb` 0.27.0 to 0.24.2, and several CUDA kernel packages. Can you confirm these are expected platform-wheel constraints rather than regressions? The end-to-end AzureML test results mentioned in the PR description are reassuring, but a brief comment near the compile command at the top of this file would help future maintainers understand why the platform flag resolves older wheels, for example:

```
# --python-platform manylinux_2_28_x86_64 resolves wheels compatible with the
# AzureML CUDA container's glibc version. Some packages resolve older versions
# than the default (native Linux) resolution; this is expected and intentional.
```

training/il/workflows/osmo/lerobot-train.yaml
rezatnoMsirhC, line 1: This workflow is missing feature parity with the AzureML path. The resources block hardcodes `gpu: 1` and `MIXED_PRECISION` is not wired as an environment variable. A user running this workflow on an OSMO pool node with multiple GPUs would silently get single-GPU training: `gpu: 1` caps the container's device allocation regardless of the pool SKU, and `train.py` defaults `MIXED_PRECISION` to `"no"` when the variable is absent.

The fix mirrors the pattern already used in `training/il/workflows/azureml/lerobot-train.yaml`:

```yaml
# resources block
resources:
  default:
    gpu: "{{ num_gpus }}"

# environment block
environment:
  MIXED_PRECISION: "{{ mixed_precision }}"

# default-values
default-values:
  num_gpus: 1
  mixed_precision: "no"
```

With `num_gpus: 1` and `mixed_precision: "no"` as defaults, existing single-GPU submissions are unaffected.

infrastructure/setup/manifests/azureml-instance-types.yaml
algattik: Thanks, fixed PR description.

training/il/scripts/lerobot/train.py
algattik: Investigating the behavior, it appears that Draccus is not store_true — argparsing.py (https://github.com/dlwh/draccus/blob/main/draccus/argparsing.py) coerces every action to type=str, and field_wrapper.py uses the default action="store" with no nargs. In argparse that means exactly one value is required, accepting both --policy.use_amp=true and --policy.use_amp true.

This is easily verified:
```bash
$ lerobot-train --policy.use_amp
lerobot-train: error: argument --policy.use_amp: expected one argument
```

The bare form errors in draccus itself before reaching _strip_use_amp, so the suggested fix would actually break the legitimate space-separated case: ["--policy.use_amp", "true", "--steps=10000"] → leaves "true" as an orphan positional that draccus then rejects.

 The test name was misleading — renamed to test_strip_use_amp_space_separated_value with a comment pointing at the draccus semantics (73629c7). Keeping the consume-next-token behaviour.

training/il/workflows/osmo/lerobot-train.yaml
algattik: The comment is out of scope for this PR.  The only change to this file is adding `--no-cache-dir --no-deps` to the `uv pip install` line — a load-bearing correctness fix tied to the lockfile regeneration in this PR, not a multi-GPU change.

The rationale lives in training/il/scripts/lerobot/azureml-train-entry.sh:37-41: requirements.txt is now a fully-resolved uv pip compile lockfile whose pins depend on [tool.uv] override-dependencies in pyproject.toml (e.g. azure-storage-blob==12.29.0 above azureml-mlflow's <=12.27.1 cap), and overrides apply at compile time only. Re-resolving at install time would fail with the conflicts the overrides exist to bypass. The OSMO workflow needed the same --no-deps switch for the same reason.

Suggesting to submit an issue to separately resolve the gap you've identified.

training/il/lerobot/requirements.txt
algattik: The file is supposedly autogenerated, but it seems that dependabot decorrelated them. Here we are just enforcing what's documented.

Suggesting to close the thread. We shouldn't add comments in this file, as it's autogenerated.

See also https://github.com/microsoft/physical-ai-toolchain/issues/770#issuecomment-4552016935

training/il/workflows/osmo/lerobot-train.yaml
rezatnoMsirhC: Agreed, the scope boundary is reasonable. Tracked the multi-GPU and mixed-precision OSMO parity gap in #785.

rezatnoMsirhC (approved):

PR #784, opened 2026-05-27
fix(deployment): make robotics chart setup reproducible

### LINKED ISSUE #786: Fix robotics chart setup portability and GPU Operator rendering
## Summary

Robotics chart deployment should use the configured pinned GPU Operator chart version, render a valid `ClusterPolicy`, and verify downloaded Helm chart archives portably on macOS and Linux.

## Problem

The setup path has several portability and correctness failures:

- BSD `find` does not support `find -printf`, causing macOS chart archive verification to fail after `helm pull`.
- `sha256sum` is not installed by default on macOS; `shasum -a 256` is required there.
- GNU and BSD `sed -i` use different syntax, making the hash update helper non-portable.
- Chart helper functions use stdout as their return channel, so `helm pull` status output or diagnostics can corrupt captured values.
- The pinned GPU Operator chart can render an invalid empty `validator.plugin:` block unless values explicitly avoid that rendering path.

## Acceptance Criteria

- Robotics chart setup uses the configured pinned `GPU_OPERATOR_VERSION`.
- Downloaded chart archives are selected and SHA256-verified on macOS and Linux.
- Chart helper diagnostics do not contaminate stdout return values.
- GPU Operator values render a valid `ClusterPolicy`.
- ShellCheck and Helm rendering validation pass for the changed scripts and values.

algattik:
# Pull Request

## Description

Fix robotics chart deployment so the GPU Operator uses the configured pinned chart version, renders a valid `ClusterPolicy`, and verifies downloaded chart archives portably on macOS.

The macOS bug was caused by Linux/GNU-specific shell assumptions in the chart verification path:

- `find -printf` is not supported by BSD `find`, so the setup script failed after `helm pull` even though the chart archive had been downloaded.
- `sha256sum` is not installed by default on macOS, while `shasum -a 256` is.
- GNU and BSD `sed -i` use different syntax, which made the hash-update helper non-portable.

The shell command-substitution bug was that chart helper functions use stdout as their return channel. Any `helm pull` status output or `[INFO]` diagnostics on stdout could be captured into variables such as `gpu_tgz` or hash values, corrupting later Helm arguments. The fix sends those diagnostics to stderr.

The GPU Operator values also set `validator.plugin: null` as a targeted workaround for the pinned `v25.3.x` chart rendering an invalid empty `validator.plugin:` block. This does not disable plugin validation: in the GPU Operator API the plugin validator spec is a non-pointer field, and the controller only uses `spec.validator.plugin.env` to append extra env vars to the plugin-validation init container.

The fix replaces those assumptions with portable archive selection, checksum calculation, and defaults-file rewriting. It also keeps the GPU Operator default on the pinned `GPU_OPERATOR_VERSION` so the chart version matches the configured SHA256 hash.

Closes #786.

## Type of Change

- [ ] 📚 Documentation update

## Component(s) Affected

- [ ] `docs/` - Documentation

## Testing Performed

- [x] Helm render verified for GPU Operator `ClusterPolicy`
- [x] Shell syntax and ShellCheck passed for changed scripts
- [x] Chart helper stdout contract verified
- [x] GPU Operator source checked for validator plugin behavior
- [x] GPU Operator installed in dev AKS and advertised `nvidia.com/gpu: 1`

`npm run lint:yaml` was attempted locally but blocked because `actionlint` is not installed on PATH.

## Documentation Impact

## Bug Fix Checklist

- [x] Regression test included, OR
- [x] Justification for no regression test: Helm rendering and live dev-cluster installation validate the chart behavior; no automated chart-deploy test exists for this setup script.

## Checklist

- [ ] No new linting warnings introduced


scripts/lib/common.sh
```diff
@@ -81,16 +106,16 @@ pull_and_verify_chart() {
...
+  tgz=$(find_latest_chart_archive "$output_dir")
   [[ -n "$tgz" ]] || fatal "No .tgz found in $output_dir after helm pull"
 
   if [[ -n "$expected_sha" ]]; then
     local actual_sha
-    actual_sha=$(sha256sum "$tgz" | awk '{print $1}')
+    actual_sha=$(calculate_sha256 "$tgz")
     if [[ "$actual_sha" != "$expected_sha" ]]; then
       fatal "SHA256 mismatch for $tgz: expected=$expected_sha actual=$actual_sha. Run scripts/update-chart-hashes.sh to update pinned hashes."
     fi
```
bindsi, line 118: NIT: Nice catch fixing the stdout leak here — before this redirect, `gpu_tgz=$(pull_and_verify_chart ...)` in `01-deploy-robotics-charts.sh` would have captured the `[INFO] Chart hash verified: ...` line into the variable and passed it as the first arg to `helm upgrade --install`.

Two small follow-ups worth considering in the same PR:

1. Add a one-line note to the `pull_and_verify_chart` header reminding maintainers that the function returns the archive path on stdout, so all diagnostics must go to stderr. The return contract is already documented; the stderr requirement is implicit.
2. The companion helper `pull_chart_sha` in `scripts/update-chart-hashes.sh` has the same stdout-is-return-value contract. It emits no `info` today, but `helm pull` itself can print a short status line on newer versions. Consider `helm pull ... >&2` there too, to make it robust against helm output changes.

infrastructure/setup/values/nvidia-gpu-operator.yaml
```diff
@@ -119,6 +119,7 @@ nodeStatusExporter:
 # Validator
 validator:
   enabled: true
+  plugin: null
```
bindsi, line 125: NIT: This looks like a workaround for a chart-rendering bug at the pinned `v25.3.4`, but the diff carries no explanation. Six months from now, a reader bumping the chart won't know whether to keep this line.

Could you add a short comment matching the convention used a few lines above for `mig.strategy: single` — e.g.:

```yaml
# Workaround: gpu-operator chart v25.3.x renders an invalid ClusterPolicy
# when validator.plugin uses chart defaults. Remove once upstream is fixed
# (track via <issue-link>).
plugin: null
```

Also worth confirming whether `plugin: null` actually disables the plugin validation step at runtime, or just shifts the bad render — want to make sure we're not silently regressing GPU-readiness checks.

bindsi (approved):
Nice macOS portability fix. Removing the runtime version resolution also closes a real reproducibility gap — the pinned SHA was previously verifying a chart whose version was resolved upstream at runtime.

Two NITs inline, both non-blocking.

scripts/lib/common.sh
algattik: Fixed

infrastructure/setup/values/nvidia-gpu-operator.yaml
algattik: Fixed

rezatnoMsirhC (approved):

Issue #794, opened 2026-05-28
fix(infrastructure): KAI Scheduler v0.5.5 binder webhook caBundle uses non-CA cert, blocks all OSMO pod admission


algattik:
## Component

OSMO Control Plane (Helm/Kubernetes) — KAI Scheduler dependency installed by `infrastructure/setup/01-deploy-robotics-charts.sh`.

## Summary

A fresh `infrastructure/setup` deployment installs KAI Scheduler `v0.5.5`, which generates its own self-signed serving certificate for `binder.kai-scheduler.svc` and patches that same leaf certificate into the `caBundle` of its `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration`. The Kubernetes API server treats `caBundle` as a CA trust anchor, but the certificate KAI installs lacks `CA:TRUE` / `keyCertSign`. The API server therefore rejects the chain with `parent certificate cannot sign this kind of certificate`, every admission call to `binder.kai-scheduler.svc` returns HTTP 500, and **all pod creates in `osmo-workflows` are blocked**, including OSMO `CreateGroup`. The failure is silent until a workflow is submitted.

## Symptoms

OSMO workflow submission fails immediately with backend exit code `3002`:

```text
CreateGroup job failed: Got exception when running backend execute:
InternalServerError: 500
Internal error occurred: failed calling webhook "binder.run.ai":
  failed to call webhook: Post "https://binder.kai-scheduler.svc:443/mutate--v1-pod?timeout=10s":
  tls: failed to verify certificate:
  x509: certificate signed by unknown authority
  (possibly because of "x509: invalid signature: parent certificate cannot sign this kind of certificate"
   while trying to verify candidate authority certificate "binder.kai-scheduler.svc")
```

Equivalent reproduction outside OSMO with a server-side dry-run:

```bash
kubectl run probe --image=busybox --restart=Never -n osmo-workflows --dry-run=server -- /bin/true
# Error from server (InternalError): Internal error occurred: failed calling webhook "binder.run.ai" ...
```

## Steps to reproduce

1. Run the documented deploy sequence on a fresh AKS cluster:
   ```bash
   ./infrastructure/setup/01-deploy-robotics-charts.sh
   ./infrastructure/setup/02-deploy-azureml-extension.sh
   ./infrastructure/setup/03-deploy-osmo-backend.sh
   kubectl port-forward svc/osmo-service -n osmo-control-plane 8080:80 &
[...]

PR #777, opened 2026-05-26
feat(training): add AzureML data-asset support for LeRobot training

algattik:
# Pull Request

## Description

Add native AzureML Data Asset support to the LeRobot training submission path so version-pinned datasets can be mounted read-only into the training container without a download step, alongside the existing HuggingFace Hub and Azure Blob sources. The same change set hardens the submission and entry scripts (validation, signal handling, portable `mktemp`, dev-host safety) and lands comprehensive coverage for the new code paths.

Closes #787

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change fixing an issue)
- [x] ✨ New feature (non-breaking change adding functionality)
- [ ] 💥 Breaking change (fix or feature causing existing functionality to change)
- [x] 📚 Documentation update
- [ ] 🏗️ Infrastructure change (Terraform/IaC)
- [ ] ♻️ Refactoring (no functional changes)

## Component(s) Affected

- [ ] `infrastructure/terraform/prerequisites/` - Azure subscription setup
- [ ] `infrastructure/terraform/` - Terraform infrastructure
- [ ] `infrastructure/setup/` - OSMO control plane / Helm
- [ ] `workflows/` - Training and evaluation workflows
- [x] `training/` - Training pipelines and scripts
- [x] `docs/` - Documentation

### New feature: AzureML Data Assets as a LeRobot dataset source

- `training/il/scripts/submit-azureml-lerobot-training.sh` accepts `--dataset-asset azureml:NAME:VERSION` (repeatable) and renders the AzureML job YAML with `mode: ro_mount` inputs so the data asset is mounted natively into the container — no download, no extra copy.
- URIs must be version-pinned. Shorthands such as `@latest` are rejected up front to keep runs reproducible.
- Data assets, blob URLs, and HuggingFace Hub repos can be combined; sources are merged at runtime via `lerobot-edit-dataset` with fail-fast dataset compatibility validation.
- AzureML lineage tags are bounded (length-capped) and the full resolved dataset lineage is stored as a model artifact for downstream auditability.
- `--compute` is required at parse time with an actionable error; AzureML model-name pattern is validated client-side before job submission.

### Runtime entrypoint hardening (`training/il/scripts/lerobot/azureml-train-entry.sh`)

- New `BLOB_URLS` branch downloads blob-backed datasets when present.
- `EXTERNALLY-MANAGED` marker removal is gated on `AZUREML_RUN_ID` so it cannot mutate a developer workstation if the script is executed locally.
[...]

PR #795, opened 2026-05-29
fix(training): repair OSMO LeRobot Azure Blob submissions

### LINKED ISSUE #790: fix(training): fix OSMO LeRobot Azure Blob submissions
## Component

Training Workflows (OSMO)

## Bug Description

`submit-osmo-lerobot-training.sh` does not yet match the LeRobot runtime and data-source fixes recently made for the AzureML path. An OSMO LeRobot submission using Azure Blob data fails during dependency resolution because the runtime reports Python 3.11.9 while `lerobot==0.5.1` requires Python >=3.12.

Azure Blob submissions should also not require a Hugging Face dataset ID. The OSMO submission path should accept Azure Blob data directly and pass enough dataset metadata to the training workflow without forcing a Hugging Face identifier.

This should apply the relevant submission/runtime/requirements fixes from the recent AzureML branch work in #777 and #778 to the OSMO LeRobot workflow path.

## Steps to Reproduce

1. Run `submit-osmo-lerobot-training.sh` for a LeRobot training job using Azure Blob data, for example storage account/container/path input rather than a Hugging Face dataset ID.
2. Let the OSMO workflow start installing dependencies in the training container.
3. Observe dependency resolution fail because the runtime is Python 3.11.9 and `lerobot==0.5.1` requires Python >=3.12.

## Expected Behavior

OSMO LeRobot submissions using Azure Blob data complete dependency setup successfully and do not require a Hugging Face dataset ID. The OSMO runtime should use Python >=3.12 or an otherwise compatible container/environment, and the installed requirements should match the Linux/Python 3.12 constraints used by LeRobot.

## Logs / Error Output

```shell
2026/05/28 09:13:58 [lerobot-train] Data Source: Azure Blob (mshexhackdata/hackathon-data/v2_20260507T180450Z)
2026/05/28 09:13:58 [lerobot-train] Installing system dependencies...
2026/05/28 09:14:55 [lerobot-train] Installing UV package manager...
2026/05/28 09:14:56 [lerobot-train] Training scripts unpacked to /workspace/lerobot_payload
2026/05/28 09:14:56 [lerobot-train] Using Python 3.11.9 environment at: /opt/conda
2026/05/28 09:14:57 [lerobot-train]   × No solution found when resolving dependencies:
2026/05/28 09:14:57 [lerobot-train]   ╰─▶ Because the current Python version (3.11.9) does not satisfy
2026/05/28 09:14:57 [lerobot-train]       Python>=3.12 and lerobot==0.5.1 depends on Python>=3.12, we can conclude
2026/05/28 09:14:57 [lerobot-train]       that lerobot==0.5.1 cannot be used.
2026/05/28 09:14:57 [lerobot-train]       And because you require lerobot==0.5.1, we can conclude that your
2026/05/28 09:14:57 [lerobot-train]       requirements are unsatisfiable.
2026/05/28 09:14:57 [lerobot-train] Command failed with error: exit status 1
2026/05/28 09:14:57 [lerobot-train][osmo] Upload Start
2026/05/28 09:14:57 [lerobot-train][osmo] No Files in Output Folder
```

## Checklist

- [x] I have searched existing issues for duplicates
- [x] I have read the documentation

algattik:
# Pull Request

## Description

OSMO LeRobot training from Azure Blob datasets was broken: the OSMO training
container ships Python 3.11.9 while `lerobot==0.5.1` requires Python >=3.12,
so every submission failed at dependency resolution before training could
start (#790). This PR builds a Python 3.12 venv inside the OSMO workflow,
aligns the OSMO submission path with the AzureML LeRobot work that landed
in #777 and #778, removes the legacy `STORAGE_ACCOUNT` / `BLOB_PREFIX`
data-source convention end-to-end, and repairs collateral runtime breakage /
adds troubleshooting guidance for issues discovered while testing.

Closes #790

#### CLI and submission path

- **`submit-osmo-lerobot-training.sh`**: replaces `--from-blob` /
  `--storage-account` / `--storage-container` / `--blob-prefix` with
  `--blob-url URL` (repeatable, full `https://…blob.core.windows.net/…`
  URIs). Unknown flags now fail fast; pass-through to `osmo workflow submit`
  only via `--`.
- **`submit-azureml-lerobot-training.sh`**: wires the shared
  `_validate_blob_urls.py` helper into the AzureML submit path
  (previously unprotected — malformed URLs only surfaced after the
  AzureML job had launched). The rest of the script's AzureML
  data-asset / multi-GPU surface landed in #777 / #778.
- **`run-lerobot-pipeline.sh`** *(breaking)*: removes the no-op
  `--mlflow-enable` flag (the downstream OSMO submit logs MLflow
  unconditionally). Azure subscription / resource group / workspace
  are now always required.

#### OSMO workflow runtime (root cause of #790)

- **`training/il/workflows/osmo/lerobot-train.yaml`**: builds a
  `uv`-managed Python 3.12 venv at `/opt/lerobot-venv` inside the OSMO
  pod and installs `lerobot==0.5.1` into it (the actual Python pin —
  the OSMO platform / pod templates are not Python-aware). Also:
  - Forwards `BLOB_URLS` (JSON array, single-quoted to survive YAML
    escaping) so the training pod downloads datasets via the Azure
    SDK with the pod's federated identity — no SAS tokens required.
  - Hardens `apt-get`: `DEBIAN_FRONTEND=noninteractive`,
    `Acquire::Retries=3`, `--no-install-recommends`, drops
    `libgl1-mesa-glx`, conditional `libglib2.0-0t64` for Ubuntu 24+.
  - Removes the PEP 668 `EXTERNALLY-MANAGED` marker before bootstrapping
    `uv` (single-purpose container, no system-package interference risk).
  - Adds `--policy.push_to_hub=false` and `--wandb.enable=false` to
    `TRAIN_ARGS` as defensive defaults.

#### AzureML training runtime

- **`training/il/scripts/lerobot/azureml-train-entry.sh`**: replaces the
  `rm /usr/lib/python3.*/EXTERNALLY-MANAGED` hack (gated on
  `AZUREML_RUN_ID` to guard developer machines) with
  `pip install --break-system-packages uv`. The previous approach was
  image-fragile — the marker lives under `/usr/local/lib/...` on
  python.org / PyTorch images, so the `/usr/lib/` glob missed it — and
  needed the gate at all only because deleting host system files was a
  foot-gun. The new path is portable across base images and
  self-documenting.

#### OSMO pod template (shared memory)

- **`pool-config.template.json`** / **`platform-template-config.template.json`** /
  **`pod-template-config.template.json`**: add `USER_SHM_SIZE` (default
  `8Gi`) and inject a `Memory`-medium `emptyDir` volume mounted at
  `/dev/shm` in every OSMO user pod. PyTorch DataLoader workers
  exchange tensors through `/dev/shm`; the Kubernetes default of 64 MiB
  causes silent worker crashes on multi-worker LeRobot training.
  Applies to **all OSMO user workloads**, not only LeRobot.

#### OSMO control-plane deploy

- **`03-deploy-osmo-control-plane.sh`**: adds `--force-conflicts` to
  the Helm 4 upgrade. HPA controllers take ownership of
  `.spec.replicas` after first install, and Helm's SSA path would
  otherwise fail subsequent upgrades. Helm wins the field once; HPA
  reclaims it within seconds.

#### Lineage and registration

- **`train.py`**: MLflow `dataset.source` lineage tag is now
  `{osmo,azureml}-lerobot-{hf,blob}` (discriminated via `AZUREML_RUN_ID`
  and `BLOB_URLS`). **Note for MLflow consumers**: this *renames* the
  previous values (`osmo-lerobot-training`, `osmo-azure-data-training`)
  — any downstream filters need updating. Also: the
  `register_final_checkpoint()` exit code is now propagated to the job
  exit code (previously discarded), and the dead `STORAGE_ACCOUNT`
  branch and an identical-condition `if`/`elif` are removed.
- **`checkpoints.py`**: drops the legacy
  `STORAGE_ACCOUNT` / `STORAGE_CONTAINER` / `BLOB_PREFIX` env-var
  convention (URL synthesis, lineage fields, AzureML tags) — the only
  remaining data-source conventions are `BLOB_URLS`, `DATASET_ASSETS`,
  and `DATASET_REPO_ID`. Hardens the AML registration path:
  `_register_model_via_aml` now wraps `client.models.create_or_update`
  in a bounded exponential-backoff retry (3 attempts, ~3 s worst-case)
  to absorb transient 5xx blips; both the periodic and the
  post-training registration calls benefit. Deletes the redundant
  `upload_checkpoints_to_azure_ml` reconciliation pass (its only
  purpose was to cover the earlier mark-before-upload ordering bug in
  `upload_new_checkpoints`, which is now fixed at the source — a
  checkpoint is marked `uploaded` only after the registration attempt
  completes) and reworks `upload_new_checkpoints` with explicit
  drop-on-failure semantics for intermediate checkpoints. The
  canonical "this is the deployable model" registration is
  `register_final_checkpoint`, whose non-zero exit still propagates
  to the job exit code.
- **`register_checkpoint.py`** *(deleted, 59 LOC)*: superseded by the
  consolidated `checkpoints.py` registration path; its standalone
  entry point had no remaining callers.

#### Shared helpers

- **`training/il/scripts/lerobot/_env.py`** *(new)*: replaces five
  inconsistent string comparisons (`!= "[]"`, `!= "{}"`, missing
  `.strip()`, mismatched sentinel sets) with a single JSON-aware
  helper — `parse_url_list_env` / `has_blob_urls`.
  `download_dataset._parse_env_config` now rejects payloads that
  survive `json.loads` but contain only empty / `null` entries
  (previously accepted as "non-empty list", then crashed deep inside
  `parse_blob_url`).
- **`training/il/scripts/lerobot/_validate_blob_urls.py`** *(new)*:
  extracts the previously-inline OSMO blob-URL validation into a
  shared module wired into both submission scripts. Also rejects
  URLs containing userinfo (`user:pass@host`) and ASCII control
  characters (C0 + DEL). `urlparse(...).hostname` silently strips
  userinfo, so without this check
  `https://creds@acct.blob.core.windows.net/c/p` would pass host
  validation and propagate verbatim into workflow params, pod env,
  and MLflow lineage. Control-character rejection closes the parallel
  log-injection vector in the same surfaces.

#### Streamed blob downloads

- **`download_dataset.py`**: replaces `download_blob().readall()` with
  a streamed `readinto()` into a sibling tmp file plus `os.replace()`.
  The old code buffered each blob into a `bytes` object, OOM-killing
  the training pod on multi-GB MP4 episodes. The new path streams in
  chunks, parallelises via `max_concurrency=4`, verifies
  `written == blob.size`, and is crash-safe (no half-written files at
  the final path).

#### Documentation

- **`docs/operations/troubleshooting.md`**: documents the KAI binder
  TLS x509 drift (#794) with a tiered manual recovery procedure.
- **`docs/training/lerobot-training.md`**,
  **`docs/training/experiment-tracking.md`**,
  **`docs/recipes/training/your-first-lerobot-training-job.md`**,
  **`docs/recipes/data-collection/preparing-datasets-for-training.md`**,
  **`docs/reference/scripts.md`**,
  **`docs/reference/scripts-examples.md`**,
  **`docs/infrastructure/cluster-setup.md`**: updated for the new
  `--blob-url` CLI, the removed `--mlflow-enable` flag, and the new
  MLflow lineage tag values.

## Type of Change

- [x] 💥 Breaking change (CLI surface of `submit-osmo-lerobot-training.sh`
  and `run-lerobot-pipeline.sh`; legacy `STORAGE_ACCOUNT` data-source
  convention; MLflow `dataset.source` tag values)
- [x] 📚 Documentation update

## Breaking changes

- `submit-osmo-lerobot-training.sh`: removed `--from-blob`,
  `--storage-account`, `--storage-container`, and `--blob-prefix`. Use
  `--blob-url URL` (repeatable) for Azure Blob datasets.
- `submit-osmo-lerobot-training.sh`: unknown CLI flags now fail fast instead
  of being silently forwarded. Pass-through to `osmo workflow submit` only
  via `--`.
- `run-lerobot-pipeline.sh`: removed the `--mlflow-enable` flag. The flag
  only forwarded to `submit-osmo-lerobot-training.sh`, which logs MLflow
  unconditionally, so the flag was a no-op. Azure subscription / resource
  group / workspace are now always required (previously only when
  `--mlflow-enable` or `--register-model` was set).
- `train.py` / `checkpoints.py`: the legacy `STORAGE_ACCOUNT` /
  `STORAGE_CONTAINER` / `BLOB_PREFIX` env-var convention is no longer
  read anywhere. Submissions that still set these (without `BLOB_URLS`)
  fall through to the HuggingFace path. The OSMO workflow yaml no
  longer forwards them.
- MLflow `dataset.source` tag values: `osmo-lerobot-training` /
  `osmo-azure-data-training` are replaced by
  `{osmo,azureml}-lerobot-{hf,blob}`. Downstream MLflow filters must be
  updated.
- OSMO user pods: every pod now has a `/dev/shm` `Memory`-medium
  `emptyDir` of `USER_SHM_SIZE` (default `8Gi`). Pools that previously
  ran without explicit shared memory will see `8 GiB` deducted from
  pod memory accounting.
- `download_dataset._parse_env_config`: `BLOB_URLS='[""]'` / `'[null]'` /
  any list with no usable URL strings now raise `ValueError` at parse
  time rather than reaching `parse_blob_url`. Operator impact is limited
  to surfacing the same failure earlier with a clearer message.
- `register_checkpoint.py` deleted: any external caller invoking
  `python -m training.il.scripts.lerobot.register_checkpoint` must
  switch to `training.il.scripts.lerobot.checkpoints`.

## Component(s) Affected

- [x] `infrastructure/setup/` - OSMO control plane / pod template / Helm
- [x] `workflows/` - Training workflow (Python 3.12 venv, BLOB_URLS)
- [x] `training/` - LeRobot training scripts and shared helpers
- [x] `docs/` - Operations and training documentation

## Testing Performed

- [x] `pytest training/tests/` — 480+ passed, 1 skipped (pyarrow), 0
  failures attributable to this PR (two pre-existing
  `TestDownloadDatasetFromUrl` failures on `main` are unrelated)
- [x] `shellcheck` clean on `submit-osmo-lerobot-training.sh`,
  `submit-azureml-lerobot-training.sh`, `run-lerobot-pipeline.sh`,
  `03-deploy-osmo-control-plane.sh`, and `04-deploy-osmo-backend.sh`
- [x] `ruff check training/il/scripts/lerobot/` — clean (new helpers at
  100% line coverage)
- [x] `npm run lint:md`, `npm run spell-check`, `npm run lint:py` — clean
- [x] OSMO LeRobot training submitted and completed end-to-end from an
  Azure Blob dataset (validates fix for #790)
- [ ] Smoke tests passed (`smoke_test_azure.py`) — N/A, no infra changes

## Documentation Impact

## Bug Fix Checklist

- [x] Linked to issue being fixed (#790)
- [x] Regression test included
  (`test_lerobot_train.py::TestMain::test_blob_urls_use_blob_source_tag`,
  `test_azureml_run_id_uses_azureml_source_tag`,
  `test_azureml_run_id_with_blob_urls_uses_azureml_blob_source_tag`,
  `test_lerobot_env.py::TestParseUrlListEnv` /
  `TestHasBlobUrls`,
  `test_lerobot_validate_blob_urls.py::TestValidateAccepts` /
  `TestValidateRejects` (including `test_userinfo_in_netloc` and
  `test_control_characters`),
  `test_lerobot_download_dataset.py::TestParseEnvConfig`,
  `test_lerobot_dataset_assets.py::TestRegisterModelLineage`
  empty-entry / mixed-list / pretty-print cases)

## Checklist


training/il/scripts/lerobot/checkpoints.py
```diff
@@ -63,6 +90,29 @@
         return None
 
 
+def _create_model_with_retry(client: Any, model: Any, checkpoint_name: str) -> Any:
```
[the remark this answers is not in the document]
algattik: Fixed

training/il/scripts/lerobot/_validate_blob_urls.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    if not urls:
+        raise SystemExit("--blob-url: at least one Blob URL is required.")
+    if len(urls) > _MAX_BLOB_URLS:
+        raise SystemExit(f"--blob-url: too many Blob URLs ({len(urls)}); maximum is {_MAX_BLOB_URLS}.")
+
+    for url in urls:
+        if any(c < " " or c == "\x7f" for c in url):
+            raise SystemExit(
+                f"--blob-url must not contain control characters (CR, LF, NUL, …); "
```
rezatnoMsirhC, line 41: The control character check rejects ordinals `< 32` and `== 127`, but a single quote (`'`, ordinal 39) passes. The OSMO workflow YAML uses a YAML single-quoted string for the Jinja substitution:

```yaml
BLOB_URLS: '{{ blob_urls }}'
```

Azure Blob prefix paths can legitimately contain `'` (for example `datasets/it's-a-subset`). If one does, the Jinja expansion produces:

```yaml
BLOB_URLS: '["https://account.blob.core.windows.net/c/it's-a-subset"]'
```

This is invalid YAML (the inner `'` terminates the string early), and the submission fails with a YAML parse error that gives no hint the blob URL is the cause.

Suggested: add a guard immediately after the existing control character check:

```python
        if "'" in url:
            raise SystemExit(
                f"--blob-url must not contain single-quote characters; "
                f"they corrupt the YAML single-quoted string that expands blob_urls "
                f"in the OSMO workflow template: {url!r}"
            )
```

And a corresponding test in `TestValidateRejects`:

```python
    def test_single_quote_in_path(self):
        with pytest.raises(SystemExit, match="must not contain single-quote"):
            _validate("https://acct.blob.core.windows.net/c/it's-path")
```

training/il/scripts/lerobot/checkpoints.py
```diff
@@ -31,6 +41,23 @@ def _source_list_tag(values: list[str], plural: str) -> str:
...
+    """
+    if not raw:
+        return []
+    try:
+        json.loads(raw)
+    except (json.JSONDecodeError, TypeError) as exc:
+        print(f"[AzureML] Failed to parse {var_name}: {exc}. Falling back to other dataset source metadata.")
+        return []
+    return parse_url_list_env(raw)
+
```
rezatnoMsirhC, line 59: Non-blocking nit: `_parse_with_diagnostic` calls `json.loads(raw)` once to detect malformed JSON and emit the diagnostic, then calls `parse_url_list_env(raw)` which calls `json.loads(raw)` again. The double parse is harmless but redundant.

Suggested: parse once and use the result:

```python
def _parse_with_diagnostic(raw: str, var_name: str) -> list[str]:
    if not raw:
        return []
    try:
        parsed = json.loads(raw)
    except (json.JSONDecodeError, TypeError) as exc:
        print(f"[AzureML] Failed to parse {var_name}: {exc}. Falling back to other dataset source metadata.")
        return []
    if not isinstance(parsed, list):
        return []
    return [item for item in parsed if isinstance(item, str) and item.strip()]
```

This inlines the four-line filtering logic from `parse_url_list_env` and eliminates the redundant parse with identical behaviour.

training/il/scripts/lerobot/_validate_blob_urls.py
```diff
@@ -0,0 +1,85 @@
...
+_MAX_BLOB_URLS = 64
+_BLOB_HOST_SUFFIX = ".blob.core.windows.net"
+
+
+def validate(urls: list[str]) -> None:
+    """Raise :class:`SystemExit` on the first malformed blob URL.
+
+    Mirrors the constraints applied by
+    :func:`training.il.scripts.lerobot.download_dataset.parse_blob_url` so a
+    URL accepted here will also be accepted by the downloader at run-time.
```
rezatnoMsirhC: Non-blocking nit: the docstring says "Mirrors the constraints applied by `parse_blob_url`" but does not mention that this function applies additional checks. A future maintainer may read "mirrors" as meaning the two are kept exactly in sync and skip adding a new constraint to one of them.

Suggested rewording:

```python
    """Raise :class:`SystemExit` on the first malformed blob URL.

    Mirrors the constraints applied by
    :func:`training.il.scripts.lerobot.download_dataset.parse_blob_url`,
    so a URL accepted here will also be accepted by the downloader at
    run-time. Additional checks (control characters, explicit ports,
    query strings, userinfo, and AzureML identifiers) catch misuse at
    submission time rather than container start-up.
    """
```

training/il/scripts/lerobot/_validate_blob_urls.py
algattik: Implemented the single-quote rejection guard and added the regression test in `training/tests/test_lerobot_validate_blob_urls.py`. Validation passed with the targeted test/lint run.

training/il/scripts/lerobot/checkpoints.py
algattik: Updated `_parse_with_diagnostic()` to parse JSON once and reuse the parsed result, which removes the redundant second parse while preserving the existing malformed-JSON fallback behavior.

training/il/scripts/lerobot/_validate_blob_urls.py
algattik: Clarified the docstring to state that the validator applies additional submission-time checks beyond the downloader’s URL parsing constraints, which makes the intent of the guard explicit for future maintenance.

bindsi (approved):
Automated review: Comprehensive fix for OSMO LeRobot blob submissions. Key improvements: (1) streaming blob download with `readinto()` instead of `readall()` prevents OOM on multi-GB episodes, (2) atomic tmp-file + `os.replace()` prevents partial/corrupt downloads, (3) `validate_content=True` ensures data integrity, (4) short-read detection. The blob URL validation module and checkpoint handling improvements are clean supporting additions. LGTM.

bindsi (approved):
Automated batch review: no actionable findings.

katriendg:
@rezatnoMsirhC I believe all comments have been addressed by @algattik - can we get a quick re-review from you and if all good merge this one? Thanks

rezatnoMsirhC (approved):

PR #857, opened 2026-06-03
chore(ci): remove JaCoCo parser config from codecov.yml

Jah-yee:
Remove JaCoCo parser configuration from codecov.yml as it is not needed.


bindsi (approved):
Automated review: Clean removal of unused JaCoCo parser config from codecov.yml. This repo is Python-only (no Java/Kotlin), so the JaCoCo parser section is indeed dead config. LGTM.

PR #870, opened 2026-06-03
feat(training): add GR00T VLA fine-tuning workflow for OSMO

akhanattentive:
# Pull Request

## Description

Adds GR00T-N1.5 / N1.7 vision-language-action fine-tuning as a first-class training path on OSMO, alongside the existing LeRobot ACT / Diffusion flows. Introduces a dedicated submission script, OSMO workflow, embodiment-agnostic config templates, and documentation covering both GR00T codebase versions.

**What's included:**

- `training/vla/scripts/submit-osmo-lerobot-vla-fine-tuning.sh` — submission CLI with `--vla-version {1.5,1.7}`, dataset blob URL injection, auto-resolution of data/modality config files, optional Azure ML checkpoint mirror, optional ACR OCI push, and resume support.
- `training/vla/workflows/osmo/groot-train.yaml` — self-contained OSMO workflow that clones Isaac-GR00T at the version-pinned ref, materializes the data/modality config from env-injected base64 payloads, and runs fine-tuning on `h100gpu_platform`.
- `training/vla/configs/groot/examples/` — embodiment-agnostic reference templates: `data_config.py` (N1.5 `DATA_CONFIG_MAP["example"]`), `modality_config.py` (N1.7 `ModalityConfig`), and a README with adapt-for-your-embodiment guidance.
- `docs/training/lerobot-training.md` — adds GR00T-N1.7 row to the policy architectures table, renames the GR00T section to cover both N1.5/N1.7, splits the Quick Start into per-version examples, and extends the GR00T parameter table with `--vla-version`, `--data-config-file`, `--modality-config-file`, and `--acr-registry`.
- `docs/training/osmo-training.md` — adds a `🤖 GR00T VLA Fine-Tuning` section with per-version submit examples and cross-links to the lerobot doc.
- `training/vla/README.md` — directory tree, Quick Start, config layout.
- `pyproject.toml` — per-file-ignore for the example `data_config.py` fragment (F821, RUF012 — file references symbols only resolvable after runtime append).

Closes #

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change fixing an issue)
- [x] ✨ New feature (non-breaking change adding functionality)
- [ ] 💥 Breaking change (fix or feature causing existing functionality to change)
- [x] 📚 Documentation update
- [ ] 🏗️ Infrastructure change (Terraform/IaC)
- [ ] ♻️ Refactoring (no functional changes)

## Component(s) Affected

[...]

PR #916, opened 2026-06-08
feat(training): add VLA pi0 AzureML training stack

### LINKED ISSUE #929: feat(training): add VLA pi0 AzureML training
## Problem

`training/vla/scripts/` and `training/vla/workflows/` are empty. There is no AzureML path to train pi0 / pi0_fast policies on the toolchain, despite LeRobot upstream supporting them via `--policy.type=pi0`.

## Proposed solution

Add a pi0 AzureML training submit script + workflow YAML, forked from the IL LeRobot training submission. Key constraint: **do not vendor the OpenPI fork**. LeRobot / HuggingFace already support pi0 upstream — passing `--policy.type=pi0` to `lerobot-train` is sufficient.

## Acceptance criteria

- [ ] `training/vla/scripts/submit-azureml-vla-training.sh` submits a pi0 training CommandJob
- [ ] `training/vla/workflows/azureml/vla-training.yaml` defines the job
- [ ] No OpenPI fork code vendored
- [ ] README documents pi0 / pi0_fast usage

## Related

- PR: #916 (draft)

junkataoka:
## Description

Add an Azure ML training stack for the `pi0`, `pi0_fast`, and `pi05` vision-language-action policy family. The stack uses LeRobot's upstream `[dataset,pi]` extras rather than vendoring OpenPI and reuses the IL LeRobot Azure ML entrypoint.

The shared entrypoint accepts a `LEROBOT_PROJECT` directory, defaulting to the IL dependency project. VLA jobs select `training/vla/lerobot`, whose Linux x86_64 `uv.lock` provides a reproducible runtime dependency set.

Add Azure ML E2E coverage that stages a synthetic LeRobot dataset, submits a short expert-only pi0 training job, and validates the code snapshot, MLflow tracking, checkpoint output, and registered model.

Closes #929

## Type of Change

- [x] ✨ New feature

## Components Affected

- [x] `tests/e2e/` - Azure ML end-to-end coverage

## Testing Performed

- Azure ML pi0 E2E test completed successfully with synthetic LeRobot data and `--log-freq 1`
- Live expert-only pi0 training completed two GPU steps, emitted MLflow metrics, and produced checkpoint `000002/pretrained_model`
- `bash -n` syntax checks passed for the modified shell entrypoint and VLA submission script
- `--config-preview` smoke tests passed for the default configuration, advanced policy and precision options, custom dependency project, and invalid policy rejection
- Markdown lint, spell check, and table formatting checks passed for `training/vla/README.md`

One validation workspace could not upload outputs or register the model because its storage account blocks public network access and its existing GPU compute is not network-integrated. The training path completed before that environment-specific failure.

## Documentation Impact

`training/vla/README.md` documents the directory structure, supported policies, submission examples, checkpoint registration, gated Hugging Face model access, and E2E invocation.

## Checklist

## Files

| File | Action |
|------|--------|
| `.cspell/general-technical.txt` | Add technical terms used by the VLA dependency and submission flow |
| `.github/workflows/dependency-review.yml` | Document and allow advisories constrained by the LeRobot runtime |
| `tests/e2e/_aml.py` | Add VLA submission and Azure ML job/model cleanup helpers |
| `tests/e2e/conftest.py` | Add gated Hugging Face token validation |
| `tests/e2e/test_e2e_aml_vla_pi0_training.py` | Add the Azure ML pi0 training E2E test |
| `training/il/lerobot/pyproject.toml` | Align the shared stable `wrapt` pin |
| `training/il/scripts/lerobot/azureml-train-entry.sh` | Select a locked LeRobot project and authenticate before gated model downloads |
| `training/il/scripts/lerobot/train.py` | Forward expert-only training configuration |
| `training/vla/README.md` | Document the implemented pi0-family Azure ML workflow and E2E test |
| `training/vla/lerobot/pyproject.toml` | Define pinned `lerobot[dataset,pi]` Azure ML runtime dependencies |
| `training/vla/lerobot/uv.lock` | Lock the VLA runtime for Linux x86_64 |
| `training/vla/scripts/submit-azureml-vla-pi0-training.sh` | Add pi0-family Azure ML configuration, asset registration, and job submission |
| `training/vla/workflows/azureml/vla-pi0-train.yaml` | Add the pi0-family Azure ML command job templates |


bindsi (approved):
Automated batch review: no actionable findings.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,674 @@
...
+    --environment-version VER     Environment version (default: 1.0.0)
+    --image IMAGE                 Container image (default: pytorch/pytorch:2.11.0-cuda12.8-cudnn9-runtime)
+    --assets-only                 Register environment without submitting job
+
+TRAINING OPTIONS:
+    -w, --job-file PATH           Job YAML template (default: training/vla/workflows/azureml/vla-pi0-train.yaml)
+    -p, --policy-type TYPE        Policy architecture: pi0, pi0_fast, pi05 (default: pi0)
+    -j, --job-name NAME           Job identifier (default: vla-pi0-training)
+    -o, --output-dir DIR          Container output directory (default: /workspace/outputs/train)
+        --policy-repo-id ID       Pre-trained policy for fine-tuning (HuggingFace repo)
```
dwhiteddsoft: The help text says this flag is the fine-tuning source, but the current training path appears to treat it as policy.repo_id (repo naming/push semantics), while warm-start behavior is tied to policy.path.

References:
- https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/il/scripts/lerobot/train.py#L488-L514
- https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/il/scripts/lerobot/azureml-train-entry.sh#L58-L85

Can we either wire this flag to actual warm-start semantics, or reword this help text so it is not described as fine-tuning input?

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,674 @@
...
+version: $version
+image: $image
+EOF
+
+  info "Publishing AzureML environment ${name}:${version}"
+  az ml environment create --file "$env_file" \
+    --name "$name" --version "$version" \
+    --resource-group "$rg" --workspace-name "$ws" \
+    --subscription "$sub" >/dev/null 2>&1 || \
+    warn "Environment ${name}:${version} already exists or registration failed; continuing"
```
dwhiteddsoft: This swallows all environment create failures and continues. That can make a run look successful while still using a stale environment version (old image/dependencies), which is a correctness and security risk.

Would you consider failing on non already-exists errors instead of always warning and continuing?

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,674 @@
...
+# Register Environment
+#------------------------------------------------------------------------------
+
+register_environment "$environment_name" "$environment_version" "$image" \
+  "$resource_group" "$workspace_name" "$subscription_id"
+
+info "Environment: ${environment_name}:${environment_version}"
+
+if [[ "$assets_only" == "true" ]]; then
+  info "Assets prepared; skipping job submission per --assets-only"
```
dwhiteddsoft, line 544: Because environment-create errors above are downgraded to warnings, assets-only can return success even if environment publication failed.

Could we gate this success path on confirmed environment registration success so assets-only is reliable in CI and automation?

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,674 @@
...
+                                  model. Accepted forms:
+                                    azureml:NAME:VERSION       (numeric version required)
+                                    azureml://.../models/NAME/versions/VERSION
+                                    https://...blob.core.windows.net/...
+                                  Shorthands like azureml:NAME or azureml:NAME@latest
+                                  are rejected to keep runs reproducible.
+                                  Optimizer, scheduler, and step counter start fresh.
+                                  Mutually exclusive with --policy-repo-id.
+        --lerobot-version VER     Specific LeRobot version or "latest" (default: latest)
+        --lerobot-requirements PATH
```
dwhiteddsoft, line 83: The flag is documented as a generic PATH override, but current behavior appears repo-relative only. Validation prepends REPO_ROOT in https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/vla/scripts/submit-azureml-vla-pi0-training.sh#L436, then the raw value is passed to runtime via env in https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/vla/scripts/submit-azureml-vla-pi0-training.sh#L591, where the entrypoint expects a file path inside the container code asset in https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/il/scripts/lerobot/azureml-train-entry.sh#L29-L32.

Could we either:
- explicitly define this flag as repo-relative and reject absolute paths with a targeted error, or
- stage external files into the submitted code asset when absolute paths are provided?

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,674 @@
...
+AZURE CONTEXT:
+        --subscription-id ID      Azure subscription ID
+        --resource-group NAME     Azure resource group
+        --workspace-name NAME     Azure ML workspace
+        --compute TARGET          Compute target override
+        --instance-type NAME      Instance type (default: gpu). pi0 (~3B params)
+                                  does NOT fit reliably on spot instances; default
+                                  uses on-demand GPU. Only honoured on AzureML-on-
+                                  Kubernetes compute. On AzureML managed AmlCompute
+                                  the script detects the compute type and drops the
```
dwhiteddsoft: This help text says the script detects managed AmlCompute and drops `resources.instance_type`, but I do not see detection logic in this script.

Current behavior appears unconditional when set:
- help claim: https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/vla/scripts/submit-azureml-vla-pi0-training.sh#L94-L99
- submission path: https://github.com/microsoft/physical-ai-toolchain/blob/991bddf2d122083d2a67dedf96b863f628ebf672/training/vla/scripts/submit-azureml-vla-pi0-training.sh#L533-L534

Could we either implement compute-type detection and conditional omission, or adjust help text to match actual behavior?

training/vla/lerobot/requirements.txt
```diff
@@ -0,0 +1,672 @@
...
+    #   wandb
+google-auth==2.53.0
+    # via databricks-sdk
+gymnasium==1.3.0
+    # via lerobot
+h11==0.16.0
+    # via
+    #   httpcore
+    #   uvicorn
+hf-xet==1.5.1.dev1
```
dwhiteddsoft: This lockfile currently pins a dev prerelease (`hf-xet==1.5.1.dev1`). Since this is production training infra, could we prefer a stable release pin (or document why a prerelease is required)? Allowing prereleases globally can increase supply-chain and reproducibility risk.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Good catch — added an example `--policy-repo-id lerobot/pi0` line to the help text in commit 4ff0838 so users know this points at the LeRobot HF policy used for the run.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Fixed in 4ff0838. `register_environment()` now captures stderr to a temp file and only treats the `already exists / 409` case as benign (warn + continue). Any other failure logs the captured output and `return 1`s, so the run aborts before submission.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Transitively addressed by the `register_environment()` fix in 4ff0838 — since `register_environment` now returns non-zero on real failures and `set -o errexit` is enabled, the assets-only path will no longer reach the "success" exit when env registration fails.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Added an absolute-path guard in 4ff0838 — values starting with `/` are now rejected with a `fatal` message before the `REPO_ROOT/` prepend, so the flag clearly behaves as a repo-relative override.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Expanded the `--instance-type` help in 4ff0838 to list the available cluster instance types (`defaultinstancetype`, `gpuspot`, `gpu`) and to call out that for managed AmlCompute the script auto-strips `resources.instance_type` from the job spec.

training/vla/lerobot/requirements.txt
junkataoka: Pinned `hf-xet==1.5.1` (stable) in 4ff0838 — added a surgical override in `training/vla/lerobot/pyproject.toml` and regenerated the lockfile. The global `[tool.uv.pip] prerelease = "allow"` flag stays (still required by transitive Azure/LeRobot deps), and the comment there now mentions `hf-xet` alongside `sentry-sdk` so the precedent is documented.

bindsi (approved):
Approved: the VLA pi0 AzureML training stack remains internally consistent after re-review. I did not find actionable blockers in the current diff.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,718 @@
...
+
+AZURE CONTEXT:
+        --subscription-id ID      Azure subscription ID
+        --resource-group NAME     Azure resource group
+        --workspace-name NAME     Azure ML workspace
+        --compute TARGET          Compute target override
+        --instance-type NAME      Instance type for AzureML-on-Kubernetes compute
+                                  (default: gpu). pi0 (~3B params) does NOT fit
+                                  reliably on spot instances; default uses on-
+                                  demand GPU. The value is forwarded to the job
```
akzaidi: This has less to do with spot GPUs and more to do with GPU memory size. Perhaps we reocmmend that full finetuning, an a100 or h100 is necessary, and for smaller gpus they should fine-tune the expert action head only? Something like this:

```python
    --instance-type NAME      Instance type for AzureML-on-Kubernetes compute
                              (default: gpu). pi0 (~4B params) full fine-tuning
                              needs a high-memory GPU (~40 GB+, A100/H100-class). For
                              smaller GPUs, use --policy.train_expert_only to
                              freeze the VLM backbone. The value is forwarded to
                              the job as resources.instance_type whenever non-empty.
```

training/vla/workflows/azureml/vla-pi0-train.yaml
```diff
@@ -0,0 +1,75 @@
...
+  type: managed_identity
+
+# All values below are placeholders and will be overridden by submit-azureml-vla-pi0-training.sh
+code: .
+environment: azureml:vla-pi0-training-env:latest
+compute: azureml:cpu-cluster
+
+inputs:
+  dataset_repo_id: none
+  policy_type: pi0
```
akzaidi, line 36: pi0 loads a gated backbone (`google/paligemma-3b-pt-224`), but there's no `HF_TOKEN`
forwarded here (and no `--hf-token` option above). On the `--dataset-asset` / `--blob-url`
path the entry script never logs into HF (it only does so in the HF-Hub-dataset branch,
`azureml-train-entry.sh:163-173`), so the gated backbone download fails in a fresh
AzureML container with no cached creds. Suggest adding an `--hf-token` option (default
`${HF_TOKEN:-}`) and forwarding it as `environment_variables.HF_TOKEN` here, mirroring #870 .

training/vla/lerobot/pyproject.toml
```diff
@@ -0,0 +1,121 @@
...
+    "accelerate==1.13.0",
+    "torchcodec==0.14.0; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
+    # Pin gitpython to address GHSA-mv93-w799-cj2w (transitive via mlflow-skinny and wandb).
+    "gitpython==3.1.50",
+    # Pin urllib3 to address GHSA-qccp-gfcp-xxvc and GHSA-mf9v-mfxr-j63j (transitive via many HTTP clients).
+    "urllib3==2.7.0",
+    # Pin wrapt to the latest stable in the 1.x line. opentelemetry-instrumentation
+    # constrains wrapt below 2.0 and above 1.0, but with prerelease resolution
+    # enabled below uv would otherwise pick the 1.17.4 release candidate over
+    # 1.17.3 stable.
```
rezatnoMsirhC: The comment says "Pin wrapt to the latest stable in the 1.x line" and cites avoiding `1.17.4rc1` as the rationale, but the actual pin is `wrapt==2.2.1` (the 2.x line). The comment also states "opentelemetry-instrumentation constrains wrapt below 2.0," yet 2.2.1 is in the lockfile, which suggests that constraint no longer applies at `opentelemetry-instrumentation==0.63b1`.

The comment appears to be a carry-over from an earlier iteration of the pin that was not updated when the version moved to 2.x. A maintainer bumping lerobot will read this and have incorrect expectations about both the version in use and the constraint landscape.

Suggested update:

```toml
    # Pin wrapt to a known-good stable release. With prerelease resolution enabled
    # uv would otherwise select a release candidate. 2.x is compatible with the
    # opentelemetry-instrumentation version currently in use; re-check upper-bound
    # compatibility when bumping opentelemetry-instrumentation.
    "wrapt==2.2.1",
```

training/vla/lerobot/pyproject.toml
```diff
@@ -0,0 +1,121 @@
...
+    "datasets==5.0.0",
+]
+package = false
+
+[tool.uv.pip]
+# Required for prerelease transitives pulled by the resolved tree, including
+# Azure/OpenTelemetry instrumentation betas, wrapt release candidates, and
+# safetensors release candidates. After regenerating requirements.txt, audit
+# the full set with:
+#   grep -E '[0-9][ab][0-9]+|rc[0-9]+' requirements.txt
```
rezatnoMsirhC: The audit instruction here says to run `grep -E '[0-9][ab][0-9]+|rc[0-9]+' requirements.txt` after each lockfile regeneration. Running that against the current lockfile reveals three release-candidate versions beyond the explicitly documented ones:

- `msal==1.38.0rc1` (Azure authentication library, handles token acquisition for all AzureML operations)
- `safetensors==0.8.0rc1` (model weight loading)
- `tokenizers==0.23.0rc0` (HuggingFace tokenizer)

The existing comment mentions `pydantic`, `sentry-sdk`, and `hf-xet` as the packages requiring stable pins, but does not cover these three. A future maintainer regenerating the lockfile will discover them via the audit command but have no record of whether they are intentional or new drift. `msal` is the most sensitive given it controls Azure credential acquisition.

Two options to address this:

**Option A** — add direct stable pins, consistent with the `hf-xet` fix in this PR (verify stable availability before applying):

```toml
    "msal==1.38.0",        # stable; RC otherwise selected by prerelease resolution
    "safetensors==0.8.0",  # stable; RC otherwise selected by prerelease resolution
    "tokenizers==0.23.0",  # stable; RC otherwise selected by prerelease resolution
```

**Option B** — document the accepted RCs in the `[tool.uv.pip]` comment block so the audit result is self-explanatory:

```toml
[tool.uv.pip]
# Required for prerelease transitives. Known RCs at lerobot 0.5.1 resolution time
# (re-verify on each lockfile regeneration with the grep command above):
#   msal==1.38.0rc1        - no stable patch available at resolution time
#   safetensors==0.8.0rc1  - idem
#   tokenizers==0.23.0rc0  - idem
# ...
prerelease = "allow"
```

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Reworded the help text in 8310e89 — it now frames the flag as the AzureML **InstanceType CRD** name (`gpu`, `gpuspot`, etc.) rather than the spot/non-spot axis, and notes that the CRD itself selects the underlying SKU/memory tier.

training/vla/lerobot/pyproject.toml
junkataoka: Rewrote the comment in 8310e89 to match what's actually happening:

- The 1.x pin is no longer about "avoiding deepspeed-style upstream incidents" — it's because `wrapt==2.0.0rc1` is the only 2.x release on PyPI and is a prerelease.
- The comment now states that we pin to the latest **stable** 1.x line (`1.17.3`) to keep the lockfile free of unintended prereleases, with a forward-looking note to revisit once `wrapt 2.x` ships a stable release.

Same rewrite was applied to `training/il/lerobot/pyproject.toml` for sibling consistency.

training/vla/lerobot/pyproject.toml
junkataoka: Addressed both halves of this feedback in 8310e89:

**1. Pinned the two RCs called out as resolvable.**
- `safetensors==0.8.0rc1` → `0.8.0` (stable released)
- `tokenizers==0.23.0rc0` → `0.23.1` (stable released; required adding `tokenizers==0.23.1` to `[tool.uv].override-dependencies` because `transformers==5.3.0` declares an upstream cap of `tokenizers>=0.22.0,<=0.23.0`)

`requirements.txt` regenerated — both packages now resolve to the stable releases (no `rc`/`a`/`b` suffix).

**2. Rewrote the audit instruction comment in `[tool.uv.pip]`** to be honest about which prereleases remain and why:
- `msal==1.38.0rc1` — Microsoft auth library RC, no stable replacement on PyPI
- `azure-core-tracing-opentelemetry==1.0.0b13`, `azure-monitor-opentelemetry-exporter==1.0.0b45` — Azure OT packages, latest published are betas
- 13 `opentelemetry-instrumentation-*` packages at `*b1` — entire OT instrumentation family is still in beta upstream

The comment now enumerates the accepted prereleases explicitly and explains the gating policy, so the audit check is interpretable rather than a tautological grep.

training/vla/workflows/azureml/vla-pi0-train.yaml
junkataoka: Added HF_TOKEN plumbing end-to-end in 8310e89:

1. **Submit script** — new `--hf-token` flag (also honors `HF_TOKEN` env var); value forwarded into the AzureML job's environment via the existing env-var passthrough.
2. **YAML** — commented `HF_TOKEN:` placeholder in `environment_variables` documenting the wiring.
3. **Entry script** (`training/il/scripts/lerobot/azureml-train-entry.sh`) — hoisted the HF login step to run after `uv pip install --no-deps` but **before** data-source branching, since pi0 init pulls the gated `google/paligemma-3b-pt-224` backbone on all data paths (asset/blob/Hub).
4. **Secret hygiene** — submit script prints `HF Token: <set>` / `<none>` and never the value itself.

training/il/scripts/lerobot/azureml-train-entry.sh
```diff
@@ -46,6 +46,18 @@ uv venv --python 3.12 "${LEROBOT_VENV}"
 source "${LEROBOT_VENV}/bin/activate"
 uv pip install --no-cache-dir --no-deps --requirement "${LEROBOT_REQUIREMENTS}"
```
rezatnoMsirhC, line 47: The changes to this file are now in conflict with `main`. The PR parameterizes a flat `LEROBOT_REQUIREMENTS` file path, but `main` has since moved to a project-directory approach using `uv export --frozen`:

```bash
# current main
LEROBOT_PROJECT="training/il/lerobot"
…
uv export --frozen --no-hashes --no-emit-project --project "${LEROBOT_PROJECT}" \
  | uv pip install --no-cache-dir --no-deps -r -
```

This creates a merge conflict (the PR's `mergeable_state` is `dirty`). There are two additional problems that follow from the flat-file approach:

1. `training/il/lerobot/` on `main` has `uv.lock` but no `requirements.txt`, so the PR's default value `training/il/lerobot/requirements.txt` would cause IL jobs to fail at runtime after merge.

2. `training/vla/lerobot/` has `requirements.txt` (from `uv pip compile`) but no `uv.lock`. The codebase convention is "every Python subproject carries a committed `uv.lock` next to its `pyproject.toml`" and "never run `uv pip compile` to produce a committed flat file." The IL lerobot directory follows this correctly.

Suggested path forward after rebasing:

1. Replace `training/vla/lerobot/requirements.txt` with a `uv.lock` (generated via `uv lock`).
2. Add `[tool.uv] environments` to `training/vla/lerobot/pyproject.toml` to constrain the lock to the AzureML container platform, e.g.:
   ```toml
   [tool.uv]
   environments = ["python_version >= '3.12' and sys_platform == 'linux' and platform_machine == 'x86_64'"]
   ```
3. Parameterize the entry script on a `LEROBOT_PROJECT` directory env var (mirroring what main already has) rather than a flat file path.
4. Update the VLA submit script to pass `LEROBOT_PROJECT=training/vla/lerobot`.

The `uv export --frozen --project` call in the container handles the platform constraint automatically when the lock was generated with the matching `environments` entry.

rezatnoMsirhC (approved):

algattik:
I would suggest adding a test for this path under `tests/e2e`. Otherwise it is very difficult for maintainers to assess future changes.

training/il/scripts/lerobot/azureml-train-entry.sh
junkataoka: Addressed in 804b8840. I removed the committed VLA `requirements.txt`, added a Linux x86_64 `uv.lock` generated from `training/vla/lerobot/pyproject.toml`, configured `[tool.uv].environments`, and parameterized the shared Azure ML entrypoint through `LEROBOT_PROJECT` (defaulting to the IL project). The VLA submit script now sets `LEROBOT_PROJECT=training/vla/lerobot`, so both paths use the same `uv export --frozen` installation flow. All PR checks are green.

junkataoka:
@algattik  I added `tests/e2e/test_e2e_aml_vla_pi0_training.py` plus an Azure ML submission helper in `tests/e2e/_aml.py`. The test stages a synthetic LeRobot dataset, submits a short pi0 job, waits through its lifecycle, and validates the registered model, uploaded training snapshot, MLflow metrics, and checkpoint output. All PR checks are green.

junkataoka:
@algattik  Live Azure ML follow-up:

- Ran the PR branch on managed `Standard_NC8as_T4_v3` GPU compute with `lerobot/pusht`.
- The locked VLA environment restored successfully, including the `dataset` extra (`av`, `jsonlines`, and related runtime packages).
- Expert-only pi0 training completed 2 GPU steps, produced checkpoint `000002/pretrained_model`, and completed the MLflow run.
- The Azure ML job then failed during output upload/model registration because this validation workspace's default storage is policy-enforced with public network access disabled, while its existing GPU cluster is not VNet-attached. RBAC was present; a compliant replacement cluster could not be created because this subscription currently has zero available NCasT4_v3 family quota.

So the pi0 training path is live-validated. Full artifact-upload/model-registration completion remains blocked by my MCAPS workspace's network/quota configuration, not by the training implementation. The final PR revision also passes all 53 reported checks with zero failures.

tests/e2e/_aml.py
```diff
@@ -263,6 +264,56 @@ def submit_aml_lerobot_training(
...
+    aml_workspace: AzureMLWorkspace,
+    *,
+    blob_url: str,
+    training_steps: int,
+    save_freq: int,
+    batch_size: int,
+    register_model_name: str,
+) -> AzureMLJob:
+    experiment_name = e2e_name("vla-pi0-training-e2e-aml")
+    instance_type = os.environ.get("E2E_AML_INSTANCE_TYPE", "gpu")
```
algattik: Could we avoid defaulting this to `gpu` for every compute target? The submission script documents that managed AmlCompute must omit `resources.instance_type` by passing an empty value, while `aml_compute_target` only verifies that the target exists and is provisioned—it does not require Kubernetes compute. As written, a valid managed-compute E2E environment will submit with an incompatible Kubernetes-only instance type. Please detect the compute kind and omit the argument for AmlCompute, or default this override to empty and require Kubernetes environments to opt in explicitly.

tests/e2e/test_e2e_aml_vla_pi0_training.py
```diff
@@ -0,0 +1,80 @@
...
+        repo_root,
+        aml_workspace,
+        blob_url=dataset.blob_url,
+        training_steps=2,
+        save_freq=1,
+        batch_size=1,
+        register_model_name=register_model_name,
+    )
+    request.addfinalizer(lambda: cancel_aml_job(job, repo_root))
+    request.addfinalizer(lambda: archive_all_model_versions(repo_root, aml_workspace, register_model_name))
```
algattik: Pytest executes finalizers in LIFO order, so this archival finalizer runs before the cancellation registered above it. If the test fails while training is still active, archival can find no model, after which the job may register a version before cancellation takes effect and leave it behind. Please use one ordered cleanup routine that cancels the job, waits for a terminal state, and then archives all model versions.

algattik (commented):
Two E2E reliability issues should be addressed before relying on this test across supported AzureML environments.

tests/e2e/test_e2e_aml_vla_pi0_training.py
junkataoka: Fixed in `6da4688f`. The test now registers one cleanup finalizer backed by `cleanup_aml_job_and_model_versions()`. It cancels an active job, polls until Azure ML reports a terminal state, then archives every registered model version. Ruff and explicit E2E collection passed on the validation VM.

tests/e2e/_aml.py
junkataoka: Fixed in `6da4688f`. `E2E_AML_INSTANCE_TYPE` now defaults to an empty value, so managed AmlCompute omits the Kubernetes-only `resources.instance_type`; AzureML-on-Kubernetes environments can opt in explicitly (for example, `E2E_AML_INSTANCE_TYPE=gpu`).

tests/e2e/test_e2e_aml_vla_pi0_training.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.e2e
+@pytest.mark.usefixtures("aml_compute_target")
+def test_aml_vla_pi0_training_e2e(
+    request: pytest.FixtureRequest,
+    aml_workspace: AzureMLWorkspace,
+    repo_root: Path,
+    storage_account: str,
+) -> None:
+    hf_token = env_value("HF_TOKEN")
```
algattik: we could remove these 3 lines and instead:
- annotate the method with
    ```
    @pytest.mark.requires_hf_token
    ```
- add to conftest.py
    ```
    def pytest_configure(config: pytest.Config) -> None:
         config.addinivalue_line(
             "markers",
             "requires_hf_token: marks tests that require HF_TOKEN for gated Hugging Face model access",
         )
 
 
     @pytest.hookimpl(tryfirst=True)
     def pytest_runtest_setup(item: pytest.Item) -> None:
         if item.get_closest_marker("requires_hf_token") is None:
             return
         if os.environ.get("HF_TOKEN", "").strip():
             return
 
         pytest.fail(
             f"{item.nodeid} requires HF_TOKEN for gated Hugging Face model access",
             pytrace=False,
         )

    ```

Advantage: this would fail right before test suite is started, rather than during execution.

algattik: Could we document this requirement? E.g. in `README.md`, assuming you take up my `conftest.py` based suggestion:

```
## 🧪 End-to-End Test

`tests/e2e/test_e2e_aml_vla_pi0_training.py` initializes pi0 from the gated `google/paligemma-3b-pt-224` backbone. Export a Hugging Face token with access to that model before selecting this test directly or as part of a broader E2E run:

```bash
export HF_TOKEN="$(< /secure/path/to/hf-token)"
uv run pytest -vv -s -m e2e tests/e2e/test_e2e_aml_vla_pi0_training.py
```

Pytest fails during client-side setup before resolving Azure fixtures or submitting a job when `HF_TOKEN` is unset or empty. Other E2E tests do not require this variable unless they explicitly carry the `requires_hf_token` marker.
```

We should also document the requirement to accept the gated model, and explain how to do so.

File: tests/e2e/_aml.py [hunk 15]
```diff
@@ -263,6 +264,56 @@ def submit_aml_lerobot_training(
...
+) -> AzureMLJob:
+    experiment_name = e2e_name("vla-pi0-training-e2e-aml")
+    instance_type = os.environ.get("E2E_AML_INSTANCE_TYPE", "")
+    log_e2e(
+        "Submitting AzureML VLA pi0 training job "
+        f"for dataset={blob_url}, training_steps={training_steps}, "
+        f"save_freq={save_freq}, batch_size={batch_size}, experiment={experiment_name}, "
+        f"instance_type={instance_type or '<managed-compute>'}"
+    )
+    result = run_command(
```
algattik, line 286: could we add a `--log-freq` parameter, and set its value to `1` in the e2e test?

Otherwise the test doesn't see any logging, and fails.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
...
+  --set "inputs.mixed_precision=$mixed_precision"
+)
+
+[[ -n "$policy_repo_id" ]]      && az_args+=(--set "inputs.policy_repo_id=$policy_repo_id")
+[[ -n "$lerobot_version" ]]     && az_args+=(--set "inputs.lerobot_version=$lerobot_version")
+[[ -n "$training_steps" ]]      && az_args+=(--set "inputs.training_steps=$training_steps")
+[[ -n "$batch_size" ]]          && az_args+=(--set "inputs.batch_size=$batch_size")
+[[ -n "$eval_freq" ]]           && az_args+=(--set "inputs.eval_freq=$eval_freq")
+[[ -n "$register_checkpoint" ]] && az_args+=(--set "inputs.register_checkpoint=$register_checkpoint")
+
```
algattik, line 588: could we add support for `environment_variables.AZURE_CLIENT_ID` as was added to sister script in commit 5339ca96?

```
[[ -n "$managed_identity_client_id" ]] && az_args+=(--set "environment_variables.AZURE_CLIENT_ID=$managed_identity_client_id")
```

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
...
+    # Register environment only (no job submission)
+    submit-azureml-vla-pi0-training.sh -d placeholder --assets-only
+EOF
+}
+
+#------------------------------------------------------------------------------
+# Helpers
+#------------------------------------------------------------------------------
+
+ensure_ml_extension() {
```
algattik: here we could reuse require_az_extension ml in scripts/lib/common.sh

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
...
+  az extension show --name ml &>/dev/null ||
+    fatal "Azure ML CLI extension not installed. Run: az extension add --name ml"
+}
+
+register_environment() {
```
algattik: here we could reuse register_azureml_environment() in common.sh

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
...
+target.write_text("\n".join(lines[: inputs_index + 1] + inserted + lines[inputs_index + 1 :]) + "\n", encoding="utf-8")
+PY
+}
+
+#------------------------------------------------------------------------------
+# Defaults
+#------------------------------------------------------------------------------
+
+environment_name="vla-pi0-training-env"
+environment_version="1.0.0"
```
algattik: Instead of a fixed version, we could call derive_azureml_environment_version_from_image. See submit-azureml-lerobot-training.sh

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
...
+image="${IMAGE:-pytorch/pytorch:2.11.0-cuda12.8-cudnn9-runtime}"
```
algattik: Could we use DEFAULT_LEROBOT_TRAIN_IMAGE here? See submit-azureml-lerobot-training.sh

training/vla/scripts/submit-azureml-vla-pi0-training.sh
```diff
@@ -0,0 +1,741 @@
+#!/usr/bin/env bash
+# Submit LeRobot VLA pi0 training to Azure ML
+# Installs LeRobot[dataset,pi] dynamically and trains pi0/pi0_fast/pi05 vision-language-action
+# policies from datasets in Azure Blob Storage (canonical) or HuggingFace (legacy
+# fallback via --hf-dataset). Reuses the policy-agnostic IL training entry script
+# (training/il/scripts/lerobot/azureml-train-entry.sh) with a VLA-specific
+# dependency project injected via LEROBOT_PROJECT.
+# cspell:ignore alreadyexists paligemma
```
algattik: nit: consider putting paligemma into .cspell/general-technical.txt instead

tests/e2e/_aml.py
algattik: Confirmed that the e2e test is successful with this parameter value.

algattik (changes requested):

tests/e2e/test_e2e_aml_vla_pi0_training.py
junkataoka: Implemented in `4b9ab48a`. The test now uses `@pytest.mark.requires_hf_token`, registered centrally in `conftest.py`, with a `tryfirst` setup hook that fails on an unset or blank token. A validation run with a sentinel `az` executable confirmed the hook fails before any Azure CLI or fixture setup.

junkataoka: Documented in `4b9ab48a`. The VLA README now covers accepting the gated PaliGemma model conditions, exporting an authorized read token, and the exact E2E invocation. It also explains that the marked test fails during client-side setup before Azure fixture resolution.

tests/e2e/_aml.py
junkataoka: Implemented in `4b9ab48a`. `--log-freq` is now forwarded through the E2E helper and VLA submission script to `LOG_FREQ`; the two-step E2E sets it to `1` so MLflow metrics are emitted for its assertions.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. The VLA script now resolves the compute identity with `resolve_azureml_compute_identity_client_id()` and forwards a non-empty client ID as `environment_variables.AZURE_CLIENT_ID`, matching the sister IL script.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. Removed the local Azure ML extension check and now call shared `require_az_extension ml`.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. Added `paligemma` to `.cspell/general-technical.txt` and removed the script-local suppression. Spell-check passes.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. Removed the duplicate environment publisher and now use shared `register_azureml_environment()`, including its exact image-match validation for immutable environment versions.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. The environment version is now derived with `derive_azureml_environment_version_from_image()` unless `ENVIRONMENT_VERSION` or `--environment-version` explicitly overrides it.

training/vla/scripts/submit-azureml-vla-pi0-training.sh
junkataoka: Implemented in `4b9ab48a`. The default now uses shared `DEFAULT_LEROBOT_TRAIN_IMAGE`, which is digest-pinned in `scripts/lib/common.sh`, matching the sister IL script.

algattik (approved):

algattik:
Updated PR description following updates

Issue #873, opened 2026-06-03
build(build): fail CI when a dependency update doesn't resolve or regenerate


algattik:
## Summary
Dependabot can open a PR that edits `pyproject.toml` into a state that **does not
resolve**, or that resolves to an import-incompatible set — and today that PR
passes CI green, because no workflow re-resolves the manifest or regenerates the
generated dependency files. The breakage surfaces later, at container build or
training runtime, instead of on the PR.

PR #779 mitigates the *churn* from such updates; this follow-up makes the
*breaking* ones fail CI. It covers both generated-file mechanisms in the repo.

## The core problem
Dependabot edits `pyproject.toml` directly. Two failure modes, neither caught
today:

1. **Unresolvable / incompatible manifest.** A bump can leave the constraint set
   with no valid solution, or one that breaks at import time.
   *Real example:* #743 bumped `marshmallow` to 4.x against
   `azure-ai-ml==1.33.0`, which imports symbols 4.x removed. Caught only by
   manual review, hand-pinned back to `marshmallow==3.26.2`. CI was green
   throughout.
2. **Stale generated file.** Even a valid bump leaves the generated dependency
   file out of step, because no Dependabot ecosystem regenerates a
   `uv pip compile` `requirements.txt`, and `uv sync` in CI re-resolves silently
   rather than asserting the existing `uv.lock`.

### A. `uv pip compile` → `requirements.txt` (committed, installed at runtime)
- `training/rl/requirements.txt` ← `training/rl/pyproject.toml`
- `training/il/lerobot/requirements.txt` ← `training/il/lerobot/pyproject.toml`

Installed at runtime via `uv pip install --no-deps --requirement requirements.txt`
in `training/rl/scripts/train.sh` and
`training/il/scripts/lerobot/azureml-train-entry.sh`, so they must stay
committed. CI never touches them today (`pytest-training.yml` installs from
`pyproject.toml`), so they are neither validated nor exercised.

**Gate — recompile in place, then diff:**
```bash
uv pip compile pyproject.toml -o requirements.txt \
  --python-version <3.11 rl | 3.12 il> --python-platform manylinux_2_28_x86_64  # fails if manifest unresolvable
git diff --exit-code requirements.txt                                           # fails if generated file stale
```
- Compile **into the committed file** so `uv` reads existing pins as
  preferences; without `--upgrade`, routine transitive minor releases produce no
  diff (verified empirically on `rl`/`il`). The diff fires only when
[...]

PR #779, opened 2026-05-27
build(deps): align dependabot config with uv ecosystem and pin caps

WilliamBerryiii:
## Summary

Aligns Dependabot configuration with the repository's actual package managers and pins upper-bound caps so future Dependabot PRs stop proposing bumps that violate the existing `override-dependencies` constraints. Also bumps the training pytest workflow's torch test dependency to match the new pin.

### `.github/dependabot.yml`

- Switch ecosystem from `pip` → `uv` for the `uv`-managed packages:
  - `/training/rl`
  - `/training/il/lerobot`
  - `/evaluation`
- Add new entries:
  - `/data-pipeline` (`uv`)
  - `/evaluation/sil/docker` (`pip`, SIL container image)
- Add upper-bound `ignore` caps matching the override-dependencies pins so Dependabot does not re-propose known-incompatible bumps:
  - `marshmallow >= 4.0.0` (transitive cap from `azure-ai-ml<2`)
  - `torch >= 2.12.0` (pin is `torch>=2.11.0,<2.12.0`)
  - `numpy >= 2.3.0`
  - `packaging >= 26.0`
- Group names preserved (`training-dependencies`, `lerobot-dependencies`, `inference-dependencies`) so downstream reviewer workflows (PR #498) keep working.

### `.github/workflows/pytest-training.yml`

- Bump torch test dependency `2.10.0` → `2.11.0` to track the override pin.

## Why now

Recent Dependabot waves (#739, #740, #743, #744, #776, etc.) introduced bumps that fight the override pins; capping them at the actual ceiling stops the churn at the source instead of closing each PR by hand.

## Out of scope

- The pre-existing `marshmallow==4.3.0` vs `azure-ai-ml<4.0.0` conflict in `training/rl/pyproject.toml` (introduced by #743) is not fixed here. That belongs in a separate PR focused on resolving the lockfile.

PR #995, opened 2026-06-17
feat(ci): Standardize uv.lock as the single dependency resolution source of truth

### LINKED ISSUE #994: feat(ci): Standardize uv.lock as the single dependency resolution source of truth
### Component

Development Environment

### Problem Statement

Several Python subprojects rely on hand-managed runtime-flat `requirements.txt` files for dependency resolution. These files drift from their `pyproject.toml` manifests, require manual reconciliation, and produce non-reproducible resolution across training, evaluation, and OSMO/AzureML jobs. Dependabot updates and SBOM generation lack a single canonical artifact to operate on, increasing maintenance burden and supply-chain risk.

### Proposed Solution

Adopt committed `uv.lock` files as the authoritative dependency resolution for every Python subproject, and stop committing hand-managed runtime-flat `requirements.txt` files. Each subproject carries a `uv.lock` next to its `pyproject.toml`; any runtime-flat dependency list is derived at build/submit time from the lock rather than maintained by hand.

Work items:

- Add `uv.lock` + `pyproject.toml` for the subprojects that lack them (e.g. `training/rl`, `training/il/lerobot`, `evaluation/sil/docker`, `workflows/osmo`).
- Remove committed runtime-flat requirements files and derive them at build/submit time via `uv export --frozen --no-hashes --no-emit-project` piped into `uv pip install --no-deps`.
- Constrain universal locks to supported platforms via `[tool.uv] environments` (e.g. linux x86_64 for GPU/Isaac subprojects).
- Add a read-only `uv lock --check` consistency gate (PowerShell check + `lint:uvlock` script with Pester coverage) wired into PR and main CI with a summary and annotations.
- Switch Dependabot to native `uv lock` regeneration and update ignore rules to match manifest version caps.
- Extend SBOM sparse-checkout to cover all committed locks.
- Update contributor docs to point dependency guidance at `uv.lock` instead of `requirements.txt`.

**Value**

- **Reproducible builds:** A committed lock pins the full resolved dependency graph, so training, evaluation, and OSMO/AzureML jobs resolve identical versions everywhere.
- **Lower maintenance:** Removes drift-prone hand-edited flat requirements files and the manual reconciliation they require.
- **Safer dependency updates:** Dependabot regenerates affected locks natively, and a read-only CI gate fails any PR whose lock drifts from its manifest — no manual `uv lock` step on Dependabot PRs.
- **Consistent supply-chain posture:** SBOM generation and dependency review operate on one canonical artifact per subproject.

**Acceptance criteria**

- Every Python subproject has a committed `uv.lock` and no committed runtime-flat requirements file.
- CI fails when a lock drifts from its `pyproject.toml`.
- Training/evaluation submission scripts derive runtime dependencies from the lock.
- Dependabot opens lock-regenerating update PRs that pass the consistency gate.

### Alternatives Considered

- Continue using `uv pip compile` to produce committed flat requirements files — rejected because it preserves the drift and dual-source-of-truth problems.
- Pin dependencies only in `pyproject.toml` without a lock — rejected because it does not pin transitive dependencies and yields non-reproducible resolution.

### Additional Context

Reference implementation lives on branch `feat/uv-lock-standardization`.

katriendg:
This branch makes the committed `uv.lock` the single dependency resolution source of truth for every Python subproject and stops committing runtime-flat `requirements.txt` files. Each subproject now keeps a `uv.lock` next to its `pyproject.toml`, and every runtime install path — shell scripts, the SIL Dockerfile, and the OSMO/AzureML workflow YAML — derives its requirement set at build or submit time with `uv export --frozen --no-hashes --no-emit-project` rather than reading a checked-in flat file. A new PowerShell linter and reusable workflow enforce lock/manifest consistency in CI, and several security-driven dependency pins are added through `override-dependencies` and `constraint-dependencies` with the locks regenerated to match.

> The lock is the resolution contract. Removing the committed flat files eliminates the drift surface between a hand-maintained `requirements.txt` and the manifest it was compiled from, and lets Dependabot regenerate locks natively while a read-only `uv lock --check` gate blocks any PR whose lock drifts from its `pyproject.toml`.

### Runtime install paths now derive from the lock

The four runtime entrypoints and two OSMO workflows were rewritten to export from the committed lock instead of reading a flat file:

- **evaluation/sil/evaluation.sh** and **evaluation/sil/infer.sh** export from *training/il/lerobot/uv.lock* into a `mktemp` file and install with `--project` context (full resolution, no `--no-deps`), with the lockfile-presence guard and `trap` cleanup replacing the old requirements checks.
- **training/rl/scripts/train.sh** and **training/il/scripts/lerobot/azureml-train-entry.sh** pipe `uv export --frozen ... | uv pip install --no-deps`.
- **training/il/workflows/osmo/lerobot-train.yaml** and **evaluation/sil/workflows/osmo/lerobot-eval.yaml** replaced their flat-file install steps with the same piped export, guarding on `uv.lock` presence.
- **evaluation/sil/docker/Dockerfile.lerobot-eval** now copies *pyproject.toml* and *uv.lock* and installs via `uv export --frozen | uv pip install --system --no-deps`.
- **training/utils/replay-azureml.sh** adds `uv` to `require_tools` and base64-encodes a submit-time export from *workflows/osmo/uv.lock* instead of a committed *requirements-aml-mirror.txt*.

The committed flat files were removed: *training/il/lerobot/requirements.txt* (651 lines), *training/rl/requirements.txt* (519 lines), *workflows/osmo/requirements-aml-mirror.txt* (511 lines), and *evaluation/sil/docker/requirements-lerobot-eval.txt* (9 lines).

### New manifests and lock-target constraints

Two new manifests back previously flat-only runtimes: *evaluation/sil/docker/pyproject.toml* (SIL LeRobot eval image) and *workflows/osmo/pyproject.toml* (AzureML replay mirror), both with `package = false` and a `[tool.uv] environments` marker constraining the universal lock to the deployment target so `uv export` yields a single-marker, runtime-flat set. The existing *training/il/lerobot/pyproject.toml* and *training/rl/pyproject.toml* gained the same `environments` constraint; the IL manifest also moved `prerelease = "allow"` from `[tool.uv.pip]` to `[tool.uv]` so `uv lock` honors it. The IL cuda-bindings comments were corrected from torch 2.11.0 to **torch 2.10.0**, matching lerobot's `torch<2.11` cap.

### Security-driven pins

`override-dependencies` and `constraint-dependencies` were extended with security-patched versions, documented inline with GHSA identifiers: **cryptography 48.0.1** (GHSA-537c-gmf6-5ccf) across the root, RL, IL, SIL, OSMO, and viewer projects; **starlette 1.3.1** in the viewer backend (SSRF, form DoS); and **diffusers 0.38.0** in the SIL image (GHSA-98h9-4798-4q5v). The affected `uv.lock` files were regenerated, including cryptography 46.0.7 → 48.0.1.

### CI enforcement

A new **scripts/linting/Invoke-UvLockConsistencyCheck.ps1** discovers every directory containing a `uv.lock` (excluding *.venv/*, *external/*, *node_modules/*, *.git/*, *.copilot-tracking/*, *docs/docusaurus/*), runs `uv lock --check` per project, writes JSON results plus a step summary, and exits non-zero on drift. It is covered by *Invoke-UvLockConsistencyCheck.Tests.ps1* (Pester), exposed as `lint:uvlock` in *package.json* (also appended to `lint:all`), and wrapped by the reusable **.github/workflows/uv-lock-consistency.yml**. The *pr-validation.yml* orchestrator runs it with `changed-files-only: true`, and *main.yml* expanded its SBOM file list to cover the new manifest/lock pairs and pinned `syft-version: v1.20.0`.

### Dependabot and documentation alignment

*.github/dependabot.yml* switched the evaluation entry to the `uv` ecosystem, added an `av >= 16.0.0` ignore and a `torch` floor change, and added a new `uv` entry for */workflows/osmo*. The Dependabot reviewer agent and the `aw-dependabot-pr-review` workflow now reference *training/rl/uv.lock* instead of *requirements.txt*. Documentation across *copilot-instructions.md*, *CONTRIBUTING.md*, *component-updates.md*, *pull-request-process.md*, *lerobot-training.md*, *scripts-examples.md*, and *infrastructure/setup/README.md* was updated to describe the lock-as-source-of-truth flow, the `uv lock` regen command, and the new CI gate.

Closes #994

## Type of Change

- [x] 💥 Breaking change (fix or feature causing existing functionality to change)
- [x] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation

## Testing Performed

> Lock consistency is enforced by the new `npm run lint:uvlock` gate and Pester tests in *Invoke-UvLockConsistencyCheck.Tests.ps1*. Runtime install paths were rewritten but require Isaac Sim / OSMO / AzureML execution environments to validate end-to-end.

## Documentation Impact

## Bug Fix Checklist

## Checklist

- [ ] I have performed a self-review
- [ ] No new linting warnings introduced

## Related Issues

- Closes #994
- Related to #873 (this PR addresses part of that issue)


evaluation/sil/docker/Dockerfile.lerobot-eval
nguyena2, line 1: Docker's default RUN shell at evaluation/sil/docker/Dockerfile.lerobot-eval is /bin/sh -c, so a failing uv export --frozen would be masked by uv pip install -r - exiting 0 on empty stdin, yielding a silently broken image. The shell scripts and OSMO YAML paths all run under set -euo pipefail; please add SHELL ["/bin/bash", "-o", "pipefail", "-c"] so the Dockerfile path gets the same guarantee.

osv-scanner.toml
nguyena2, line 1: osv-scanner.toml drops the PYSEC-2026-161 (starlette < 1.0.1) suppression, but the PR bumps starlette to 1.3.1 only in rl and the viewer, while the root uv.lock and evaluation/uv.lock stay at 1.0.1. Confirm those newer starlette advisories don't apply to the root/evaluation surface (or that lint:vuln still passes), so you don't merge a partially-patched transitive.

.github/dependabot.yml
rezatnoMsirhC, line 1: The `evaluation/sil/docker` dependabot entry (directory `/evaluation/sil/docker`, currently line ~148) still uses `package-ecosystem: pip`. This PR deletes `requirements-lerobot-eval.txt` and adds `pyproject.toml` and `uv.lock` in that directory, so Dependabot will no longer find a pip-resolvable file there and will silently stop producing dependency update PRs for the SIL evaluation image lock.

The main `/evaluation` entry was correctly updated to `package-ecosystem: uv` in this PR. The `/evaluation/sil/docker` entry needs the same change:

```yaml
  directory: /evaluation/sil/docker
  groups:
    lerobot-eval-image-dependencies:
      patterns:
      - "*"
  ignore:
  - dependency-name: "av"
    versions:
    - ">=16.0.0"
  labels:
  - dependencies
  - python
  - evaluation
  open-pull-requests-limit: 5
  package-ecosystem: uv
  schedule:
    day: monday
    interval: weekly
```

The `torch >=2.12.0` ignore can be dropped since `lerobot` caps torch transitively in the lock. The `av >=16.0.0` ignore mirrors the pattern added to the main evaluation entry in this PR.

rezatnoMsirhC (approved):

algattik:
I believe this also  Resolves #873 (as it addresses its acceptance criterial; with a superior implementation to the one proposed in that issue)

.github/dependabot.yml
```diff
@@ -83,7 +83,7 @@ updates:
     - ">=4.0.0"
   - dependency-name: "torch"
     versions:
-    - ">=2.12.0"
+    - ">=2.11.0"
```
algattik, line 86: **CI force-installs the exact torch version this `ignore` excludes**

This `ignore` rule tells Dependabot not to propose `torch` ≥ 2.11.0 (holding it at 2.10.x), because 2.11.0 pulls CUDA-13 bindings and breaks `libcudart` — and the lerobot lock pins `torch==2.10.0` (with `cuda-bindings==12.9.0` in `pyproject.toml` to match the CUDA-12 wheels).

But `.github/workflows/pytest-training.yml:41` does this, outside the lock and after `uv sync`:

```yaml
- name: Install Torch test dependency
  run: uv pip install torch==2.11.0
```

So the training tests run against precisely the version this ignore — and the lock — exclude. That seems at odds with this PR's "uv.lock as the single source of truth" goal: CI validates a torch the project deliberately won't ship. (The workflow isn't in this PR's diff, so flagging here for visibility.)

Is the `2.11.0` install a deliberate CPU-runner workaround? If so, could we add a one-line comment explaining it and reconcile it with this ignore. If not, dropping the step so the tests use the locked torch via `uv sync` would close the gap.

.github/dependabot.yml
katriendg: The `/evaluation/sil/docker` entry is already on `package-ecosystem: uv` in the current revision, so Dependabot will keep producing lock updates there. I've also dropped the `torch >=2.11.0` ignore per your suggestion since `lerobot` caps torch transitively in the lock; the `av >=16.0.0` ignore stays to mirror the main `/evaluation` entry.

evaluation/sil/docker/Dockerfile.lerobot-eval
katriendg: Good catch. Added `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` after the `FROM` so the `uv export --frozen | uv pip install -r -` pipe now fails the build if `uv export` errors, matching the `set -euo pipefail` guarantee the shell-script and OSMO YAML paths already have.

osv-scanner.toml
katriendg: Confirmed safe to drop the suppression. There is no root `uv.lock`, and every lock that pulls in starlette is now on `1.3.1`: `evaluation/uv.lock`, `evaluation/sil/docker/uv.lock`, `workflows/osmo/uv.lock`, `data-management/viewer/uv.lock`, `data-management/viewer/backend/uv.lock`, and `training/rl/uv.lock`. No surface remains on `1.0.1`, so the PYSEC-2026-161 entry is no longer needed and `lint:vuln` passes without it.

.github/dependabot.yml
katriendg: Good catch — this is a real cross-surface inconsistency. Three torch versions are currently in play: `2.10.0` (lerobot/eval locks, CUDA-12), `2.11.0` (this CI step), and `2.12.0` (root `pyproject.toml` dev group). The `2.11.0` step pre-dates this PR and isn't in its diff, but it does cut against the single-source-of-truth goal.

@WilliamBerryiii you introduced the `torch==2.11.0` test pin in #779 — could you confirm the intent? My read is it's a CPU-runner workaround so the unit tests don't try to install the GPU/CUDA wheels that `uv sync` (dev group `2.12.0`) or the lerobot lock (`2.10.0`) would pull.

For now I've added an inline comment on the step documenting that it's a deliberate CPU-only pin held outside the lock. If `2.11.0` isn't actually required, we can drop the step entirely and let the tests use the locked/dev-group torch via `uv sync`, which would close the gap. I can open a follow-up issue to reconcile the three versions either way.

katriendg:
> I believe this also  Resolves #873 (as it addresses its acceptance criterial; with a superior implementation to the one proposed in that issue)

Thanks for reviewing, indeed the idea is to supersede this one, there is a linked comment in the PR description, and we'll close down a PR that was addressing part of this as well.

.github/dependabot.yml
katriendg: Created a backlog issue #1023 for tracking this discussion, as we want to merge this PR in due to impact on the projects and PRs coming in. Closing this one here.

PR #996, opened 2026-06-17
build(build): fail CI when a dependency update doesn't resolve or regenerate

### LINKED ISSUE #873: build(build): fail CI when a dependency update doesn't resolve or regenerate

musahabibulloh:
Resolves #873.

Implemented automated CI validations for dependency update PRs to fail when a dependency manifest doesn't resolve or when lockfiles are not correctly regenerated. 

### Changes
- Created .github/workflows/dependency-lock-check.yml to:
  - Assert that uv pip compile requirements output is up to date for 	raining/rl and 	raining/il/lerobot.
  - Asserts that all five uv.lock files are up-to-date across the repository via uv lock --check.
- Added the new job dependency-lock-check to the pr-validation pipeline and the aggregate pr-validation-summary gate.
- Documented the manual command for 
equirements.txt regeneration inside docs/contributing/component-updates.md.

### DIFF
--- a/.github/workflows/dependency-lock-check.yml
+++ b/.github/workflows/dependency-lock-check.yml
@@ -1,0 +1,98 @@
+name: Dependency Lock Check
+
+on:
+  workflow_call:
+    inputs:
+      soft-fail:
+        description: 'Whether to continue on dependency lock issues'
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  dependency-lock-check:
+    name: Dependency Lock Check
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+
+      - name: Check rl requirements.txt
+        id: rl-reqs
+        run: |
+          cd training/rl
+          uv pip compile pyproject.toml -o requirements.txt --python-version 3.11 --python-platform manylinux_2_28_x86_64
+          if ! git diff --exit-code requirements.txt; then
+            echo "RL_REQS_FAILED=true" >> $GITHUB_ENV
+          fi
+        continue-on-error: true
+
+      - name: Check il/lerobot requirements.txt
+        id: il-reqs
+        run: |
+          cd training/il/lerobot
+          uv pip compile pyproject.toml -o requirements.txt --python-version 3.12 --python-platform manylinux_2_28_x86_64
+          if ! git diff --exit-code requirements.txt; then
+            echo "IL_REQS_FAILED=true" >> $GITHUB_ENV
+          fi
+        continue-on-error: true
+
+      - name: Check uv.lock projects
+        id: uv-locks
+        run: |
+          failed=false
+          for project in . data-management/viewer data-management/viewer/backend data-pipeline evaluation; do
+            echo "Checking uv.lock in $project"
+            if ! (cd $project && uv lock --check); then
+              failed=true
+            fi
+          done
+          if [ "$failed" = true ]; then
+            echo "UV_LOCKS_FAILED=true" >> $GITHUB_ENV
+          fi
+        continue-on-error: true
+
+      - name: Add job summary
+        if: always()
+        run: |
+          echo "## Dependency Lock Check Results" >> $GITHUB_STEP_SUMMARY
+          failed=false
+          
+          if [ "$RL_REQS_FAILED" = "true" ]; then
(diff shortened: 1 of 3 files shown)


musahabibulloh:
@microsoft-github-policy-service agree

katriendg:
Hello @musahabibulloh thank you for taking on this one. I am in the process of also addressing his initial problem in the draft PR #995 - will come back on this one and see if we can align. It is broader, and a different approach, still working on it so let me get back to you.

musahabibulloh:
Hi @katriendg, thanks for letting me know! I'll leave this PR open for now so you can review it or pull any useful pieces if needed. Just let me know how you'd like to proceed once you're ready. Happy to help align with your broader approach!

katriendg:
Thanks again for this one @musahabibulloh - we will be closing this one in favor of the #995 changes.

PR #856, opened 2026-06-02
fix(training): wire AzureML named checkpoints output to runtime env var

algattik:
# Pull Request

## Description

Fixes the AzureML `checkpoints` *uri_folder* output for both RL and IL training jobs (issue #855), then adds the surrounding pieces required for that fix to be exercisable end-to-end on a scale-to-zero GPU cluster.

The root cause was wiring the named output through `TRAINING_CHECKPOINT_OUTPUT: ${{outputs.checkpoints}}` inside `environment_variables:`. Azure ML only substitutes `${{outputs.X}}` template expressions inside `command:`; inside `environment_variables:` the container receives the literal template string, writes into a directory of that name, and the data-capability sidecar uploads nothing to the named output. Training entry points now read `$AZURE_ML_OUTPUT_CHECKPOINTS` directly (the per-output runtime path Azure ML exports for every `uri_folder` output) and mirror their local *checkpoints/* tree into it at finalization. The broken `${{outputs.checkpoints}}` line is removed from both job YAMLs.

A real AzureML LeRobot e2e test was added alongside the existing SKRL one, and `assert_job_has_checkpoint` was upgraded to enumerate the *checkpoints* output blobs via the **azure-storage-blob SDK** and assert non-empty contents — the previous declaration-only check happily passed against the buggy code. The LeRobot test forced two related issues to the surface and they are fixed in the same branch: a *torchcodec* / *torch* C++ ABI mismatch from a stale override pin, and missing MLflow metrics on short runs because lerobot's `cfg.log_freq` defaults to 200. Finally, three independent defaults blocked scale-from-zero on the GPU pool the new test targets — *aml-operator* resource validation, a missing static `accelerator=nvidia` node label, and Volcano's enqueue-time capacity gate — and all three are corrected and documented.

Closes #855

### Checkpoint output wiring

- Replaced the `TRAINING_CHECKPOINT_OUTPUT` env-var indirection with a direct read of **`$AZURE_ML_OUTPUT_CHECKPOINTS`** in *training/rl/scripts/skrl_training.py* and a new `_sync_checkpoint_output` helper in *training/il/scripts/lerobot/train.py*. Both entry points mirror local *checkpoints/* into the runtime path on the SIGTERM handler and on normal finalization.
[...]

PR #997, opened 2026-06-17
feat(infrastructure): support NVIDIA OSMO 6.3 and scale-from-zero GPU pools

### LINKED ISSUE #938: feat(terraform): support infrastructure provisioning for NVIDIA OSMO 6.3
Component: Infrastructure (Terraform/Azure)

Problem Statement:
Current infrastructure provisioning does not target NVIDIA OSMO 6.3. OSMO 6.3 is required for workload identity support and is the minimum version supported by the latest NVIDIA Physical AI Data Factory blueprint.

Success Criteria:
- Add explicit OSMO 6.3 version support in infrastructure provisioning and setup scripts.
- Make OSMO control plane/backend deployment default to 6.3-compatible chart/image versions.
- Update docs/config defaults to reflect OSMO 6.3 as the supported target.

algattik:
# Pull Request

## Description

Adds support for provisioning NVIDIA OSMO 6.3 on AKS, with **scale-from-zero GPU workflow scheduling** as the headline capability (the motivating issue: GPU submissions failed with `There are no resources in platform a100gpu_platform and pool default`). Consolidates the previous two-script deploy/uninstall flow into a single `03-deploy-osmo.sh` + `cleanup/uninstall-osmo.sh`, moves OSMO configuration from the deprecated `osmo config` API to ConfigMap mode via Helm values, and prevents a dev-mode operator token-expiry failure that would otherwise 401 all Kubernetes API calls roughly an hour after operator startup.

Closes #938

## Type of Change

## Component(s) Affected

- [x] `docs/` - Documentation

### Scale-from-zero GPU workflows

Mirrors upstream PR #856 (which enabled scale-from-zero GPU pools for Azure ML) for the OSMO submission path:

- New `infrastructure/setup/values/osmo-platforms.yaml` defines the `gpu_platform` platform under pool `default` with a `gpu_tpl` pod template. OSMO injects this template into every GPU workflow pod via `override_pod_template`.
- The template supplies the three things the cluster autoscaler needs to provision a GPU node for a pending pod: the GPU `nodeSelector` (`node.kubernetes.io/instance-type`), **tolerations** matching the GPU node-pool taints (`nvidia.com/gpu:NoSchedule` and `kubernetes.azure.com/scalesetpriority=spot:NoSchedule`), and the `nvidia.com/gpu` resource request/limit.
- With those in place, a GPU submission against an idle (zero-node) autoscaling Spot GPU pool becomes schedulable, so AKS scales the pool up on demand and back down when idle — instead of rejecting the workflow with `no resources in platform ... and pool default`.
- A `gpu_gpu_required` resource validation enforces `USER_GPU >= 1` for the GPU platform. Pod-template selectors/taints are kept in sync with the Terraform GPU node pool (`vm_size`, `node_taints`, `priority = "Spot"`, `should_enable_auto_scaling = true`).

### Other changes

- **Script consolidation** — merge `03-deploy-osmo-control-plane.sh` + `04-deploy-osmo-backend.sh` into `03-deploy-osmo.sh`, and the two uninstall scripts into `cleanup/uninstall-osmo.sh`.
- **ConfigMap config mode** — remove `config/*.template.json`; OSMO platform/pool/service config now lives in Helm values (`osmo-platforms.yaml`, `osmo-control-plane.yaml`).
- **Operator token refresh** — the OSMO 6.3 dev-mode operator calls `load_kube_config()`, which caches a static kubeconfig `tokenFile` once and never refreshes it, so the ~1h AKS projected SA token would expire and every K8s call would 401. Its kubeconfig therefore uses an `exec` credential plugin instead of a static `tokenFile`, re-reading the rotated projected token on demand (validated live: client re-invokes the plugin on the first request after expiry).
- **Supply-chain** — verify SHA256 for both service and backend charts before install.
- **Terraform** — migrate federated credentials to azurerm 4.x `user_assigned_identity_id`.
- **Workflow images** — workflow init/client images now honor `--image-version` / `--use-acr`.
- **Tooling/docs** — bump OSMO CLI + devcontainer to 6.3.0; refresh docs.

## Testing Performed

- [x] OSMO workflow submitted successfully (operator authenticates; backend jobs reach `SUCCESS`)
- Scale-from-zero validated live on AKS: a GPU workflow scheduled onto a freshly-provisioned GPU node from an idle pool (no manual pre-scaling).
- Operator exec-credential refresh validated live: the client re-invokes the plugin on the first request after token expiry, and worker/listener calls authenticate past the projected-token TTL.
- `bash -n` and `shellcheck -x` clean on all modified/added shell scripts.
- YAML values files parse-validated.

## Documentation Impact

## Checklist


scripts/update-chart-hashes.sh
```diff
@@ -81,6 +82,32 @@
   echo "${version#v}"
 }
 
+# Query the latest semver release tag from a ghcr.io OCI repository.
+# Paginates through the tags list API and returns the highest v* tag.
+oci_latest_version() {
+  local repo="$1"
+  local token last_tag tags all_tags=""
+
+  token=$(curl -sf "https://ghcr.io/token?service=ghcr.io&scope=repository:${repo}:pull" | jq -r '.token // empty')
```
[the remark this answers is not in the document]
algattik: Addressed in 4f2dd6d.

This is a false positive: the flagged line fetches a short-lived GHCR pull **token** (piped to `jq`) — a registry API metadata call, not an artifact download — so there is nothing to checksum. The same applies to the subsequent tags-list query. Chart integrity *is* verified: `pull_chart_sha` computes the SHA256 that gets pinned in `defaults.conf`, and at deploy time `pull_and_verify_chart` verifies the downloaded chart against that pin.

Since the custom `dependency-not-pinned` analyzer is a line-regex with no inline-suppression mechanism, the fix builds the token URL into a variable (matching the existing tags-URL style on the line below) so the `curl` invocation no longer carries a literal URL, and adds a comment documenting why no checksum applies. Verified locally with `Test-ShellDownloadSecurity` → 0 violations.

bindsi (dismissed):
Approved: the OSMO 6.3 and scale-from-zero GPU pool support is implemented with appropriate validation and security controls. I did not find actionable issues.

docs/contributing/component-updates.md
```diff
@@ -18,17 +18,17 @@ For quick dependency commands, see the [Component Updates](pull-request-process.
...
-| Python Packages           | pip/uv    | `pyproject.toml`, `requirements.txt`                               | Mixed           | Dependabot       |
-| GitHub Actions            | GitHub    | Workflow YAML (18 files)                                           | SHA-pinned      | Dependabot       |
+| Component                 | Source    | Version Location                                               | Current Version | Update Method    |
+|---------------------------|-----------|----------------------------------------------------------------|-----------------|------------------|
+| NVIDIA GPU Operator       | Helm      | `infrastructure/setup/defaults.conf` → `GPU_OPERATOR_VERSION`  | v25.3.4         | Manual           |
+| KAI Scheduler             | Helm      | `infrastructure/setup/defaults.conf` → `KAI_SCHEDULER_VERSION` | v0.5.5          | Manual           |
+| OSMO Chart                | Helm      | `infrastructure/setup/defaults.conf` → `OSMO_CHART_VERSION`    | 1.0.1           | Manual           |
+| OSMO Image                | Container | `infrastructure/setup/defaults.conf` → `OSMO_IMAGE_VERSION`    | 6.0.0           | Manual           |
```
katriendg: **Component Inventory "Current Version" is stale (Medium)** 🔍

The Component Inventory table still lists the pre-6.3 versions even though this PR bumps them. As written, the canonical version table contradicts `defaults.conf` and `devcontainer.json` after merge.

Please update the "Current Version" column to match this PR:

* OSMO Image `6.0.0` → `6.3.0`
* OSMO Chart `1.0.1` → `1.3.0`
* GPU Operator `v25.3.4` → `v26.3.2`
* KAI Scheduler `v0.5.5` → `v0.20.1`

docs/infrastructure/cluster-setup.md
```diff
@@ -60,43 +60,21 @@ kubectl cluster-info
...
 # - AzureML: ./02-deploy-azureml-extension.sh
-# - OSMO:    ./03-deploy-osmo-control-plane.sh && ./04-deploy-osmo-backend.sh
+# - OSMO:    ./03-deploy-osmo.sh
 ```
 
 > [!IMPORTANT]
-> **Do not re-run `03-deploy-osmo-control-plane.sh` against a Postgres database that already holds OSMO state from a previous AKS cluster.** Script 03 mints a fresh Master Encryption Key on every run; the new key cannot decrypt rows wrapped by the previous one, and OSMO will fail with `jwcrypto` `InvalidJWEData` / `InvalidTag` errors on login and on workflow submission.
+> **Do not re-run `03-deploy-osmo.sh` against a Postgres database that already holds OSMO state from a previous AKS cluster.** Script 03 mints a fresh Master Encryption Key on every run; the new key cannot decrypt rows wrapped by the previous one, and OSMO will fail with `jwcrypto` `InvalidJWEData` / `InvalidTag` errors on login and on workflow submission.
 >
 > If you destroyed and re-created AKS while preserving the Postgres flexible server, first drop and re-create the `osmo` database (or `TRUNCATE` the `configs`, `credential`, `ueks`, and `backends` tables) before running script 03 again.
```
katriendg, line 69: **Upgrade path for existing installs is undocumented (Medium-High — possible blocker)** ⚠️

OSMO 6.3 folds the standalone `router` and `web-ui` charts into the `service` chart, and this PR installs a single Helm release named `osmo` (the old scripts installed three: `service`, `router`, `ui`). For anyone with an existing 6.2 deployment, an in-place rerun will (1) not adopt the old `service` release because of the rename → `osmo`, risking resource collisions, and (2) leave orphaned `router` / `ui` releases. NVIDIA's [6.3.0 release notes](https://github.com/NVIDIA/OSMO/blob/main/releases/6.3.0.md) make these hard requirements:

> "Existing router resources (`osmo-router`, `osmo-router-headless`) continue to work, but **you must remove the separate router Helm release before upgrading**."
> "Set `ui.enabled: true` in service values to deploy the UI alongside the API. **Remove the separate `web-ui` release before upgrading**."

Two more 6.3 changes affect existing users: the squid-proxy sidecar was removed from the backend operator (network policies now restrict pod-to-pod access directly), and ConfigMap mode (`services.configs.enabled: true`, this PR's default) makes CLI/API config writes return **HTTP 409** — so any database-stored config from 6.2 must be migrated, not edited live.

Suggested short migration note to add (sourced from the release notes + NVIDIA's 6.2 → 6.3 upgrade guide):

> **Upgrading from OSMO 6.2?** A direct rerun is not supported — the deployment now installs a single Helm release named `osmo`. Before deploying 6.3:
>
> 1. Export any database-stored config to Helm values with NVIDIA's `deployments/upgrades/export_configs_to_helm.py`, then fold it into `infrastructure/setup/values/osmo-platforms.yaml` (ConfigMap mode replaces the `osmo config` API).
> 2. Remove the legacy Helm releases so the new `osmo` release installs cleanly:
>    `helm uninstall web-ui router service -n osmo-control-plane` (adjust names/namespace to your install).
> 3. Run `infrastructure/setup/03-deploy-osmo.sh`.

If we'd rather not own a migration path, a one-line "fresh install only — remove legacy `service`/`router`/`web-ui` releases first" note plus the upstream link is the minimum. This is the one item I'd like to resolve before merge.

> Note: the upstream 6.2 → 6.3 upgrade-guide page returned 404 at review time (likely not yet published); the breaking-change requirements and the `export_configs_to_helm.py` tool are confirmed directly in the [6.3.0 release notes](https://github.com/NVIDIA/OSMO/blob/main/releases/6.3.0.md). Worth confirming the final link before merging.

docs/infrastructure/cluster-setup.md
```diff
@@ -60,43 +60,21 @@ kubectl cluster-info
...
 
 ## 🔐 Deployment Scenarios
```
katriendg, line 88: **State the supported OSMO version and support policy (Medium)** 💡

There's no statement that the repo targets a single current OSMO version (now 6.3) and that support tracks the current release and may change in future. A one-line note near the version inventory or this OSMO deploy step would set the right expectation for users, per the intent behind #938.

docs/operations/troubleshooting.md
```diff
@@ -341,19 +341,19 @@ RuntimeError: unable to allocate shared memory(shm) for file </torch_...>: Succe
...
 **Resolution:**
 
-OSMO mounts `/dev/shm` according to the `POD_TEMPLATE` config registered on the control plane at deploy time, with `USER_SHM_SIZE` rendered from the pool config (default `8Gi`). The mount applies only to pods created *after* the config is registered — restarting the training pod is not enough; a stuck workflow must be cancelled and resubmitted.
+OSMO mounts `/dev/shm` according to the pod template in Helm values, with `USER_SHM_SIZE` rendered from the platform config (default `16Gi` for GPU platforms). The mount applies only to pods created *after* the config is deployed — restarting the training pod is not enough; a stuck workflow must be cancelled and resubmitted.
 
 Verify the live cluster matches the deploy script's intent:
 
 ```bash
-osmo config get POD_TEMPLATE --output yaml | grep -A 3 dshm
+osmo config show POD_TEMPLATE | grep -A 3 dshm
```
katriendg: **Stale `osmo config show/update` CLI references (Medium)** 🔍

A few references to the deprecated `osmo config show/update SERVICE|POD_TEMPLATE` API remain after the move to ConfigMap mode, where those commands return HTTP 409 when a ConfigMap config is active (this `osmo config show POD_TEMPLATE` is one).

* `docs/operations/troubleshooting.md` — ~6 occurrences (e.g. lines 349, 384, 390, 395, 417, 422)
* `docs/infrastructure/manage-node-pools.md` — line 153 (`osmo config show POOL`)

Please update these to the ConfigMap/`kubectl`-based equivalents (or point at editing `osmo-platforms.yaml` + rerun, consistent with the other sections this PR added).

docs/training/osmo-training.md
```diff
@@ -126,21 +126,20 @@ osmo info
...
-# Login and verify
-osmo login http://localhost:9000 --method=dev --username=testuser
-osmo info
+# Login and configure default pool
+osmo login http://localhost:9000 --method=dev --username=admin
+osmo profile set pool default
+osmo version
 ```
 
 > [!NOTE]
```
katriendg, line 211: **Surface the headline "scale-from-zero GPU pools" feature (Medium)** 💡

The PR title leads with "scale-from-zero GPU pools", and the OSMO GPU platform plumbing (`osmo-platforms.yaml`: `gpu_tpl`, `gpu_platform`, `gpu_gpu_required` validation) is solid — but it's only documented incidentally (operational "edit and rerun" notes plus a "platform config" table cell). The AzureML scale-from-zero path is well documented at `docs/training/azureml-training.md#-scale-from-zero-gpu-pools`, but the OSMO/KAI path this PR introduces isn't explained as a feature.

A short narrative — how OSMO schedules a GPU workflow onto a pool that scales from zero, and what `gpu_platform` / `gpu_gpu_required` do — would surface the capability for users instead of leaving it in the PR description.

scripts/lib/common.sh
```diff
@@ -243,49 +234,6 @@ detect_service_url() {
...
-  local svc="azureml-ingress-nginx-controller"
-  if kubectl get svc "$svc" -n "$ns" &>/dev/null; then
-    echo "http://${svc}.${ns}.svc.cluster.local"
-  fi
-}
-
-# Validate that the OSMO service URL is reachable from the current host.
-# Internal load balancer IPs are only accessible from within the VNet; when running
-# from a devcontainer or codespace, users must port-forward instead.
-validate_service_url_reachable() {
```
katriendg, line 258: **Devcontainer/codespace port-forward hint dropped (Low)** 💡

`validate_service_url_reachable()` (which printed the devcontainer/codespace port-forward guidance) was removed here, and the quickstart `[!IMPORTANT]` callout went with it. Quickstart Step 8 now uses `kubectl port-forward` by default and the advanced doc keeps the table, so this isn't a hard regression — but the script's login-failure warning now mentions only the VPN, not port-forwarding. A one-line "if you're in a devcontainer/codespace, port-forward the gateway" hint in that warning would keep devcontainer users unblocked.

infrastructure/README.md
```diff
@@ -26,8 +26,7 @@ infrastructure/
 ├── setup/                             # Post-deploy configuration scripts
 │   ├── 01-deploy-robotics-charts.sh   # GPU Operator, KAI Scheduler
 │   ├── 02-deploy-azureml-extension.sh # AzureML K8s extension
-│   ├── 03-deploy-osmo-control-plane.sh# OSMO control plane
-│   ├── 04-deploy-osmo-backend.sh      # OSMO backend services
+│   ├── 03-deploy-osmo.sh# OSMO control plane
```
katriendg: **Directory-tree formatting typo (Low)** 🔍

```text
│   ├── 03-deploy-osmo.sh# OSMO control plane
```

Missing the spaces before the `#` comment, and the comment now only says "OSMO control plane" though the consolidated script also deploys the backend operator. Suggest: `03-deploy-osmo.sh        # OSMO service + backend operator`.

katriendg (commented):
Thank you for this one @algattik, I love you went further than just the upgrade, and simplified, cleaned up and completed the work.

The 6.3 migration is broad and carefully executed: consolidating `03-deploy-osmo-control-plane.sh` + `04-deploy-osmo-backend.sh` into a single, well-structured `03-deploy-osmo.sh`, folding router/web-ui into the service chart, moving to ConfigMap-based config, and the workload-identity / `DefaultAzureCredential` storage wiring are all exactly the right direction. The completeness is impressive — per-chart SHA256 pinning, the KAI operator wait fix, the exec-credential token-refresh workaround for the dev-mode operator, and the scale-from-zero GPU platform plumbing show a lot of care. The simplification of the shell scripts in particular is a real improvement in readability and maintainability.

I've left a set of comments below. They're mostly documentation follow-ups around the 6.3 version pin, the upgrade path for existing installs which I think we need to potentially bring out a bit more into view so that we have a vision for future upgrades (often breaking?). Also some minor stale references. Flagging the upgrade-path item as the one worth deciding on before merge.

infrastructure/setup/03-deploy-osmo.sh
```diff
@@ -0,0 +1,596 @@
...
+        info "Generating and applying MEK ConfigMap..."
+        generate_mek_config | kubectl apply -n "$NS_OSMO_CONTROL_PLANE" -f -
+        warn "Back up MEK for production: kubectl get configmap $SECRET_MEK -n $NS_OSMO_CONTROL_PLANE -o yaml > mek-backup.yaml"
+
+        # Clear service_auth from DB — it was encrypted with the old MEK and is now
+        # undecryptable. The service regenerates a fresh keypair on next start.
+        info "Clearing stale service_auth from database..."
+        kubectl run osmo-clear-auth --rm -i --restart=Never -n "$NS_OSMO_CONTROL_PLANE" \
+            --image=postgres:16 \
+            -- psql "host=${pg_fqdn} port=5432 dbname=osmo user=${pg_user} password=${pg_password} sslmode=require" \
```
rezatnoMsirhC: The `password=` field in the psql connection string is passed as a command argument to the `kubectl run` pod. Kubernetes stores the full pod spec, including all command args, in etcd and emits it in the audit log. Anyone with `kubectl get pod -o yaml` access in the namespace can read the password for the duration of the run.

Use `PGPASSWORD` as a `--env` argument instead, which does not appear in the pod spec command args:

```bash
kubectl run osmo-clear-auth --rm -i --restart=Never -n "$NS_OSMO_CONTROL_PLANE" \
    --image=postgres:16 \
    --env="PGPASSWORD=${pg_password}" \
    -- psql "host=${pg_fqdn} port=5432 dbname=osmo user=${pg_user} sslmode=require" \
    -c "DELETE FROM configs WHERE key='service_auth' AND type='SERVICE'" 2>/dev/null || \
    warn "Could not clear service_auth (DB may not be initialized yet -- safe on first deploy)"
```

infrastructure/setup/03-deploy-osmo.sh
```diff
@@ -0,0 +1,596 @@
...
+        --set "services.redis.storageClassName=default"
+        --set "services.redis.storageSize=1Gi"
+    )
+fi
+
+if [[ "$use_acr" == "false" && -n "${OSMO_SERVICE_CHART_SHA256:-}" ]]; then
+    service_chart_ref=$(pull_and_verify_chart "$service_chart_ref" "$chart_version" "$OSMO_SERVICE_CHART_SHA256" "$_WORK_DIR/service-chart")
+fi
+
+helm upgrade --install osmo "$service_chart_ref" "${service_helm_args[@]}"
```
rezatnoMsirhC, line 384: Nit: the backend chart deployment uses `--rollback-on-failure` but the service chart does not. If this upgrade fails mid-way, Helm 4 leaves the `osmo` release in a failed state, making reruns conflict without a manual `helm rollback`. Adding the flag keeps the two deployments consistent:

```bash
helm upgrade --install osmo "$service_chart_ref" \
    --rollback-on-failure \
    "${service_helm_args[@]}"
```

algattik:
@rezatnoMsirhC @katriendg thanks for your great feedback. I've addressed all comments.

katriendg (approved):
Thank you @algattik - solid addition. Good to merge from my end.

rezatnoMsirhC (approved):

PR #1000, opened 2026-06-18
feat: VLM as a Judge Dataviewer support

akzaidi:
# Pull Request

**PR Soundtrack**: [Kelela - outta time (feat. A.K. Paul)](https://www.youtube.com/watch?v=1mLNza10dKs)

## Description

Adds a VLM-as-judge evaluation capability for LeRobot episodes and surfaces it
directly inside the dataset viewer, alongside supporting refactors to the
viewer's trajectory-plot and object-detection panels.

The judge is **open-weight first** (Qwen3-VL) and reused through a single
`evaluation.vlm_judge.JudgeService` consumed by two surfaces: a standalone
CLI/policy-evaluation pipeline and the dataviewer's per-episode `JudgePanel`.
It scores outcome (MCQ self-consistency), dense process reward (GVL
shuffle-and-rank), milestones, and failure attribution, with a SHA256-keyed
disk cache so re-runs are free.

Closes #438 , #437 

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `evaluation/` - new `evaluation/vlm_judge/` harness + `evaluation/metrics/` scripts
- [x] `data-management/viewer/` - backend judge router/service, frontend `JudgePanel`, trajectory-plot and object-detection refactors
- [x] `.github/` - dataviewer agent, skill, and prompts
- [x] `docs/` - dataviewer README + harness README

## Key Changes

**VLM-as-judge harness (`evaluation/vlm_judge/`)**
- New package: backends (`qwen3-vl`, `openai-compat`, `echo`), multi-step
  `JudgeAgent`, frame extraction/tiling, disk cache, dataset adapter
  (LeRobot v2.1 + v3.0), FastAPI service, and CLI (`run.py`).
- Per-dataset result storage and policy-rollout evaluation (`policy_eval.py`).
- Harness README, evaluation shell scripts, and unit tests
  (`agent`, `api`, `cache`, `prompts`).

**Dataviewer integration (`data-management/viewer/`)**
- Backend: `routers/vlm_judge.py` (`GET`/`POST /judge`, mounted only when
  `VLM_JUDGE_ENABLED=true`), lazy-singleton `services/vlm_judge_service.py`,
  config + `.env.example` knobs.
- Frontend: `JudgePanel` (auto-hides when disabled), `use-vlm-judge` hooks,
  types, and API client; wired into the annotation workspace.
- Refactors: trajectory plot split into chart/controls/state modules;
  object-detection panel consolidated into a single `ObjectDetectionWidget`.

**Tooling & docs**
- Dataviewer agent/skill/prompts; dataviewer README env-var reference.

## Testing Performed

- [x] Backend `pytest` — VLM-judge router tests rewritten against a synthetic,
  self-generated LeRobot dataset (no pre-downloaded dataset required) plus new
  `vlm_judge_service` factory unit tests; both run in CI.
- [x] Frontend `vitest` — `JudgePanel` and `api-client` tests (incl. the
  disabled-state 404 → hidden-panel mapping).
- [x] Harness `pytest` — `evaluation/tests/vlm_judge/` suite.
- [x] Linters clean on touched files — `ruff`, `tsc`, `eslint`,
  `markdownlint-cli2`, `cspell`.

## Documentation Impact

## Checklist

## ⚠️ Reviewer Notes

- **Scope is broad.** Beyond the VLM judge, this branch also bundles an
  object-detection panel refactor and a trajectory-plot refactor. Consider
  whether to split these for easier review.
- **Pre-existing test failures** exist on the branch in unrelated subsystems
  (`detection` router, `lerobot_handler`, `hdf5_handler` — ~14 tests). These
  should be triaged before merge.
- **Runtime wiring:** the dataviewer backend reaches `evaluation.vlm_judge` via
  `PYTHONPATH` exported by `start.sh` when `VLM_JUDGE_ENABLED=true`.


.github/agents/dataviewer-developer.agent.md
```diff
@@ -224,6 +232,48 @@ Follow these codebase conventions:
...
+### Phase 5: VLM-as-Judge Evaluation
+
+Drive the VLM judge harness against the loaded datasets, either through the dataviewer's `JudgePanel` (UI verification) or the `evaluation.vlm_judge` CLI (bulk batches). Both share one disk cache, so CLI primes UI and vice versa.
+
+Full surface reference and prompt schema live in the dataviewer skill ("VLM-as-Judge Workflow" and "VLM-as-Judge Endpoints"). The agent steps below stay focused on orchestration and reporting.
+
+#### Step 1: Confirm the judge is enabled
+
+1. Read `data-management/viewer/backend/.env`.
+2. Verify `VLM_JUDGE_ENABLED=true` and that `VLM_JUDGE_BACKEND` matches user intent (`echo` for wiring, `qwen3-vl` for local GPU, `openai-compat` for remote endpoints).
```
algattik, line 244: Could we document the settings in `data-management/viewer/README.md` as well?

Could we make sure we document in detail, for non-expert users, what the feature is for, how to use it and what the settings mean, and what relation there is between "Run judge" and "Language instruction"?

evaluation/vlm_judge/dataset.py
```diff
@@ -0,0 +1,331 @@
...
+    return ""
+
+
+def _load_tasks_parquet(path: Path) -> dict[int, str]:
+    if not path.exists():
+        return {}
+    import pyarrow.parquet as pq
+
+    df = pq.read_table(path).to_pandas()
+    return {int(r["task_index"]): str(r["task"]) for r in df.to_dict(orient="records")}
```
algattik: I ran the viewer like this:
```
VLM_JUDGE_ENABLED=true VLM_JUDGE_BACKEND=echo ./start.sh
```

Selected an episode, and then Run judge. This yielded:
```
INFO:     127.0.0.1:60163 - "POST /api/datasets/hack__2026_05_20_merged/episodes/0/judge HTTP/1.1" 500 
KeyError: 'task'
```

data-management/viewer/backend/src/api/services/dataset_service/lerobot_handler.py
```diff
@@ -104,6 +123,7 @@ def discover(self, dataset_id: str, dataset_path: Path) -> DatasetInfo | None:
                 features[name] = FeatureSchema(
                     dtype=feat.get("dtype", "unknown"),
                     shape=feat.get("shape", []),
+                    names=feat.get("names"),
```
algattik: Running the backend yields:

```text
2026-06-19 14:01:12,583 - src.api.services.dataset_service.lerobot_handler - WARNING - Failed to discover LeRobot dataset ***: 1 validation error for FeatureSchema
names.0
  Input should be a valid string [type=string_type, input_value=['RIGHT_..._POSITION'], input_type=list]
    For further information visit https://errors.pydantic.dev/2.13/v/string_type
INFO:     127.0.0.1:61102 - "GET /api/datasets HTTP/1.1" 200 OK
```

That dataset's `meta/info.json` stores `names` as a list-of-lists (the error shows `names.0` is itself a list). `FeatureSchema.names` is typed `list[str] | None`, so Pydantic rejects the nested list: `discover()` raises and that one dataset is skipped (warning logged) — others still load.

`service.py`'s parallel discovery path already guards this exact case via `_normalize_feature_names()` (it flattens list-of-lists). Suggest lifting that helper into the shared `base.py` and calling it here too.

evaluation/vlm_judge/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+The tests use a `ScriptedBackend` (canned responses keyed by prompt
+fragment) and a `StubService` to validate prompt parsers, cache
+behaviour, the multi-step agent controller (success / low-confidence /
+failure flows + format-violation tolerance), and the FastAPI router. No
+GPU or network is required.
+
+## 🗺️ Roadmap
+
+- **Dataviewer integration.** Mount the HTTP router inside the dataviewer
```
katriendg: This README still lists dataviewer integration as roadmap even though this PR implements the dataviewer `JudgePanel` and backend route. Could we update this section to reflect the current feature state and link to the dataviewer setup guidance? This is the docs-side complement to algattik's request to document the VLM judge settings and non-expert usage in the dataviewer README.

data-management/viewer/backend/tests/api/test_vlm_judge_router.py
```diff
@@ -0,0 +1,132 @@
...
+
+import pytest
+from fastapi.testclient import TestClient
+
+REPO_ROOT = Path(__file__).resolve().parents[5]
+DATASETS_DIR = REPO_ROOT / "datasets"
+LEISAAC_PRESENT = (DATASETS_DIR / "leisaac-pick-orange" / "meta" / "info.json").exists()
+
+
+pytestmark = pytest.mark.skipif(
```
katriendg: The PR description says these router tests use a synthetic, self-generated LeRobot dataset and do not require a pre-downloaded dataset, but this module skips entirely unless `datasets/leisaac-pick-orange` exists. The later synthetic no-instruction case also copies an MP4 from that same fixture. Could we make the fixture fully self-contained, or update the PR description/CI expectations? As written, the main dataviewer router coverage can silently disappear in clean checkouts.

data-management/viewer/frontend/src/components/annotation-workspace/AnnotationWorkspaceContent.tsx
```diff
@@ -71,7 +72,11 @@ export function AnnotationWorkspaceContent({ shell }: AnnotationWorkspaceContent
   )
 
   const trajectoryLabelPanel = <LabelPanel episodeIndex={currentEpisode.meta.index} />
+  const trajectoryJudgePanel = (
+    <JudgePanel datasetId={currentDataset.id} episodeIndex={currentEpisode.meta.index} />
```
katriendg: `JudgePanel` supports an `instruction` override and sends it to the backend, but the workspace mounts it without passing the current `languageInstruction` from the annotation store. A user can edit the adjacent Language instruction widget, then click Run judge and still evaluate against dataset metadata or an empty fallback. Could we wire the current saved/draft `LanguageInstructionAnnotation.instruction` into this prop, or make the UI/docs explicit that Run judge ignores the annotation widget? This also aligns with algattik's docs comment asking to explain the relationship between Run judge and Language instruction.

data-management/viewer/frontend/src/lib/api-client.ts
```diff
@@ -387,3 +390,49 @@ export async function warmCache(datasetId: string, count = 5): Promise<void> {
...
+ */
+export async function fetchVlmJudgeStatus(
+  datasetId: string,
+  episodeIndex: number,
+): Promise<VlmJudgeStatus> {
+  const response = await fetch(
+    `${API_BASE}/datasets/${datasetId}/episodes/${episodeIndex}/judge`,
+    { headers: await requestHeaders() },
+  )
+  const data = await handleResponse<unknown>(response)
```
katriendg, line 440: The comment above says this returns `{ enabled: false }` when `VLM_JUDGE_ENABLED` is not configured, but the backend does not mount this route at all when the judge is disabled, so this `handleResponse()` call turns the expected default state into an error. That means a default dataviewer instance renders the error path instead of the intended disabled/hidden judge panel. Could we either always mount a lightweight status endpoint, or map the expected 404 from this specific route to `{ enabled: false, cached: false, ... }` here and cover it with an API-client/hook test?

evaluation/vlm_judge/openai_shim.py
```diff
@@ -0,0 +1,192 @@
...
+    from PIL import Image
+
+    if url.startswith("data:"):
+        _, _, b64 = url.partition(",")
+        data = base64.b64decode(b64)
+        return Image.open(io.BytesIO(data)).convert("RGB")
+    if url.startswith("http://") or url.startswith("https://"):
+        import urllib.request
+
+        with urllib.request.urlopen(url) as response:
```
katriendg: This shim accepts arbitrary `http://` and `https://` image URLs and fetches them server-side without an allowlist or timeout. Since the shim defaults to binding on `0.0.0.0`, exposing it outside a trusted local/dev network creates SSRF and request-hang risk. Could we either restrict the shim to `data:` URLs only, add explicit network constraints and timeouts, or document/enforce that it must bind only to localhost/private trusted networks?

katriendg (commented):
Thanks for the VLM-as-judge integration. I added inline comments for a few distinct correctness/completeness concerns and intentionally avoided duplicating algattik's existing comments.

High-level items that are not line-specific:

- The PR body still has `Closes #` and GitHub reports no linked issue, so completeness cannot be validated against acceptance criteria. Please link the issue or add explicit success criteria to the PR description.
- GitHub currently reports the branch as conflicting with `main`, so it is not merge-ready yet.
- I agree with algattik's existing runtime findings around the `KeyError: 'task'` path and nested LeRobot `FeatureSchema.names` handling; those should be addressed along with the comments added here.

.github/agents/dataviewer-developer.agent.md
akzaidi: incorporated some documentation and a table of settings with a description:
https://github.com/microsoft/physical-ai-toolchain/tree/feat/vlm-judge-dataviewer/data-management/viewer#vlm-as-judge-experimental

evaluation/vlm_judge/dataset.py
akzaidi: Addressed in 2f1699a

data-management/viewer/backend/src/api/services/dataset_service/lerobot_handler.py
akzaidi: Addressed in 2f1699a

evaluation/vlm_judge/README.md
akzaidi: Updated and addressed 4185d63

data-management/viewer/backend/tests/api/test_vlm_judge_router.py
akzaidi: Addressed in 201d862. These tests now generate a synthetic LeRobot v2.1 dataset on the fly, so they run in CI with no predownloaded dataset or copying.

`leisaac-pick-orange` references are now only in teh CLI harness script under `evaluation/vlm_judge` for example purposes (override with `--dataset <path>`)

data-management/viewer/frontend/src/components/annotation-workspace/AnnotationWorkspaceContent.tsx
akzaidi: Addressed in 4fa2cb4. Docs also updated to reflect the link as requested by algattik

data-management/viewer/frontend/src/lib/api-client.ts
akzaidi: Addressed in 201d862

evaluation/vlm_judge/openai_shim.py
akzaidi: Addressed in 76092c8. New `VLM_SHIM_ALLOW_REMOTE_IMAGES` set to false by default which rejects remote image URLs and bounded by a timeout `VLM_SHIM_REMOTE_IMAGE_TIMEOUT_S`. Documentation also updated: https://github.com/microsoft/physical-ai-toolchain/tree/feat/vlm-judge-dataviewer/data-management/viewer#local-model-via-the-openai-compat-shim

algattik:
nit: the UX is potentially confusing, it shows the list continues but clicking the down button is a no-op:
drop-down

evaluation/vlm_judge/cache.py
```diff
@@ -0,0 +1,101 @@
...
+    """Filesystem-backed JSON cache for ``JudgeResult.to_dict()`` payloads."""
+
+    def __init__(self, root: Path | None) -> None:
+        self._root = Path(root) if root is not None else None
+
+    @property
+    def enabled(self) -> bool:
+        return self._root is not None
+
+    def key(
```
algattik, line 32: in LeRobot v3 chunked layouts multiple episodes share one MP4 file (same path/size/mtime),  distinguished only by their time window. This causes a path collision.

I've confirmed that after running VLM judge on one episode, loading another episode gives a (wrong) cache hit.

data-management/viewer/README.md
```diff
@@ -86,6 +86,183 @@ Expected blob structure:
...
+> Sequential scoring with a local backend is slow: the model loads on the first
+> episode and each subsequent episode runs a full judgment chain. Prefer a hosted
+> `openai-compat` backend for whole-dataset runs, and leave results cached so reruns
+> skip already-scored episodes.
+
+#### "Run judge" vs. "Language instruction"
+
+The judge scores the episode against a **task instruction** — the natural-language
+goal for the episode, such as *"Grab orange and place into plate"*. **Run judge** uses
+the instruction currently shown in the viewer's **Language Instruction** panel (your
```
algattik, line 186: It might seem a more natural UX layout to have *Run judge* below *Language instructions* pane (currently it is rendered above), if users are supposed to cover those items in that order.

data-management/viewer/README.md
```diff
@@ -86,6 +86,183 @@ Expected blob structure:
...
+
+> [!NOTE]
+> Sequential scoring with a local backend is slow: the model loads on the first
+> episode and each subsequent episode runs a full judgment chain. Prefer a hosted
+> `openai-compat` backend for whole-dataset runs, and leave results cached so reruns
+> skip already-scored episodes.
+
+#### "Run judge" vs. "Language instruction"
+
+The judge scores the episode against a **task instruction** — the natural-language
```
algattik, line 184: Looking at the code, my impression is that the task instructions are ignored when clicking "Run all" (as opposed to clicking "Run judge"). Is that intentional? Should we document this behavior?

data-management/viewer/README.md
```diff
@@ -86,6 +86,183 @@ Expected blob structure:
...
+saved or in-progress edit), so refining that text changes what the judge grades against.
+In short:
+
+- **Language Instruction** = *what the robot was asked to do* (the goal the judge grades against).
+- **Run judge** = *grade this episode against that goal* and report the outcome, progress,
+  milestones, and any failure mode.
+
+When the Language Instruction is left empty, **Run judge** falls back to the task
+instruction stored in the dataset's metadata. If neither is available, it returns an
+error (HTTP 422) asking for one to be supplied.
```
algattik: When that happens (422 because no instruction available in dataset metadata), there is no information provided in the UX. Is that intentional?

data-management/viewer/README.md
```diff
@@ -86,6 +86,183 @@ Expected blob structure:
...
+> episode and each subsequent episode runs a full judgment chain. Prefer a hosted
+> `openai-compat` backend for whole-dataset runs, and leave results cached so reruns
+> skip already-scored episodes.
+
+#### "Run judge" vs. "Language instruction"
+
+The judge scores the episode against a **task instruction** — the natural-language
+goal for the episode, such as *"Grab orange and place into plate"*. **Run judge** uses
+the instruction currently shown in the viewer's **Language Instruction** panel (your
+saved or in-progress edit), so refining that text changes what the judge grades against.
```
algattik, line 187: I don't see a way to save the "Language instruction" content. Browsing away clears it out.

data-management/viewer/README.md
```diff
@@ -86,6 +86,183 @@ Expected blob structure:
...
+
+#### Settings
+
+Enable the judge via `start.sh`, or by exporting `PYTHONPATH` to the repository root
+before launching the backend.
+
+| Variable              | Default                     | Description                                                       |
+|-----------------------|-----------------------------|-------------------------------------------------------------------|
+| `VLM_JUDGE_ENABLED`   | `false`                     | Mount the `/judge` router                                         |
+| `VLM_JUDGE_BACKEND`   | `echo`                      | `qwen3-vl` (local HF), `openai-compat` (vLLM, NIM, Azure OpenAI), or `echo` |
```
algattik, line 212: Running this

```
VLM_JUDGE_ENABLED=true VLM_JUDGE_BACKEND=qwen3-vl ./start.sh   
```

Selecting a dataset, clicking Run judge, yields

```
ModuleNotFoundError: No module named 'torch'
```

data-management/viewer/start.sh
```diff
@@ -155,6 +155,18 @@ start_backend() {
     export DATA_DIR
     log_info "Using DATA_DIR=${DATA_DIR}"
 
+    # When VLM_JUDGE_ENABLED=true, expose the evaluation package to the backend
+    # process so its lazy-import of evaluation.vlm_judge succeeds.
```
algattik: This process circumvents uv dependency management.

It seems problematic to make the viewer backend depend on the evaluation package.

The evaluation package introduces a bunch of dependencies (including e.g. `onnxruntime-gpu` which is platform-specific). Just adding to PYTHONPATH forces the dataviewer backend to redeclare parallel dependencies, creating brittle coupling (see also separate comment about `torch` package).

If the intent is to make the evaluation code reusable outside the dataviewer, consider splitting it into its own package, and pip installing it into the dataviewer venv to ensure dependency resolution.

data-management/viewer/backend/src/api/routers/vlm_judge.py
```diff
@@ -0,0 +1,312 @@
...
+        prompt_version=_prompt_version(),
+        agent_config=replace(judge_service.config.agent, process_method=effective_method),
+    )
+    was_cached = not payload.force and cache.get(cache_key) is not None
+
+    try:
+        # Model inference is blocking and GPU-bound; run it in a worker thread so
+        # the single event loop stays free to serve episode/video requests while
+        # a judgment is in flight (otherwise the whole backend stalls per run).
+        result = await run_in_threadpool(
```
algattik: The `/judge` HTTP request stays open for the full model-load + inference, so there's no progress/polling and it's at the mercy of any proxy/ingress read timeout — consider 202 + status-poll (a poll endpoint already exists via  GET …/judge ) or documenting a required proxy timeout.

akzaidi: Addressed via documentation: added a Request timeouts section explaining that /judge is synchronous and that proxy/ingress/gateway read timeouts must exceed expected first-run model-load + inference latency. We did not change the endpoint to a 202/background-job contract in this PR.

katriendg: > **+1 to @algattik — I think we need to treat this as a release blocker, not a docs item.**
>
> Right now `/judge` holds the HTTP connection open for the full model-load + inference. Crediting the recent change: moving the call into `run_in_threadpool` was the right fix for event-loop starvation, so the backend no longer stalls other requests. But the *client-facing contract* is still a single long-lived synchronous request, and that's the part that breaks for end users.
>
> **Why documentation alone isn't enough here:**
> - The first run pays model-load cost on top of inference, so it's routinely multi-minute. Most default ingress/gateway/proxy configs cap read timeouts around 30–60s, so the very first thing a new user does times out — with no error surfaced, just a button that's been disabled "forever." Asking every operator to raise proxy/ingress/gateway timeouts to multi-minute values is both easy to miss and undesirable (it weakens timeout protection for the whole service, not just this route).
> - **It gets worse with "Run all":** that path fires N sequential long synchronous requests, each exposed to the same timeout, and all orchestration state lives in the browser tab. A single timeout, tab close, or network blip mid-run loses progress (the cache softens re-runs but doesn't make it resumable), and there's no server-visible job to recover. So the feature is most fragile exactly where it's most valuable (whole-dataset judging).
> - From a first-time-user lens this reads as "the feature doesn't work," which matches the tester feedback we've gotten.
>
> **Two ways to resolve it — I'd strongly prefer the first:**
>
> **Option A — full fix: `202 + poll` (the small-surface version).** The hard part is mostly already built — there's an existing `GET …/judge` status endpoint and the SHA256 disk cache. Concretely:
> - `POST /judge` enqueues the work (thread/task or a small in-process job table keyed by the cache key) and returns `202 Accepted` immediately with a status of `pending`/`running`.
> - The existing `GET …/judge` returns `pending | running | done | error` (+ result when done); the cache already gives you `done` for free on a hit.
> - Frontend swaps the one-shot mutation for `useQuery` with `refetchInterval` while `running`, so the panel shows real progress and "Run all" becomes a loop over fast status polls instead of long blocking POSTs. Cancellation and progress then come from server state, not just a client ref.
>
> This is the canonical pattern for multi-second/GPU work and it's a contained change because the endpoints and cache already exist.
>
> **Option B — minimum acceptable for this release (if A can't land now):** keep the synchronous endpoint but make the UI honest about it:
> - Replace the indefinite disabled-button state with an explicit in-progress indicator (spinner + "first run loads the model, this can take several minutes").
> - Add a client-side timeout (e.g. `AbortController`) and render a clear, retryable **timeout/failure** state instead of a hung button.
> - Document the required proxy/ingress/gateway read-timeout floor *and* keep it behind the existing `VLM_JUDGE_ENABLED` experimental flag.
>
> Option B is the floor I'd accept to ship; Option A is the right destination and is genuinely not much more work given the GET endpoint + cache are already there. Either way, "documented as synchronous" by itself isn't something I think we can release on.

evaluation/vlm_judge/cache.py
katriendg: > **+1 — flagging this as a correctness blocker.** @algattik confirmed empirically that running the judge on one episode then opening another returns the *wrong* cached verdict. In LeRobot v3 chunked layouts multiple episodes share one MP4 (same path/size/mtime), so a key built from video fingerprints aliases them together. Showing users another episode's result is worse than a hard failure. The cache key needs to include stable episode identity (e.g. `dataset_id` + `episode_idx`, and/or the actual `from_s`/`to_s` window when populated) so shared-file episodes can't collide. Worth a test that judges two episodes backed by the same MP4 and asserts distinct cache entries.

data-management/viewer/backend/pyproject.toml
```diff
@@ -14,6 +14,8 @@ dependencies = [
     "numpy==2.5.0",
     "pyarrow==24.0.0",
     "Pillow==12.2.0",
+    "av==15.1.0",
+    "openai==2.43.0",
```
rezatnoMsirhC: `openai` is not imported by any source file in `src/` directly. The only consumer is `evaluation.vlm_judge.backend` for `VLM_JUDGE_BACKEND=openai-compat`, which already declares it under its own `openai` optional extra in `evaluation/vlm_judge/pyproject.toml`. Installing it unconditionally adds a large SDK (~50 MB, 60+ transitive packages) to every dataviewer deployment, including those with `VLM_JUDGE_ENABLED=false`. Suggest moving it into the `vlm-judge` optional extra:

```toml
vlm-judge = [
    "accelerate==1.13.0",
    "torch==2.10.0",
    "transformers==5.12.1",
    "openai==2.43.0",
]
```

data-management/viewer/backend/src/api/services/vlm_judge_service.py
```diff
@@ -0,0 +1,75 @@
...
+    """Return the singleton ``JudgeService``, building it on first call.
+
+    Returns ``None`` when ``vlm_judge_enabled`` is false so callers can skip
+    mounting the router without conditional imports of optional deps.
+    """
+    global _service
+    if not config.vlm_judge_enabled:
+        return None
+    if _service is not None:
+        return _service
```
rezatnoMsirhC, line 34: The `_service` global uses a double-check pattern but has no `threading.Lock`. Judge inference runs inside `run_in_threadpool`, so two concurrent first-requests can both observe `_service is None` and both construct a `JudgeService`. For the `qwen3-vl` backend this races on GPU memory. A module-level lock removes the window:

```python
import threading
_service_lock = threading.Lock()

def get_vlm_judge_service(config: AppConfig):
    global _service
    if not config.vlm_judge_enabled:
        return None
    if _service is not None:
        return _service
    with _service_lock:
        if _service is None:
            ...
            _service = JudgeService(...)
    return _service
```

data-management/viewer/README.md
```diff
@@ -86,6 +86,202 @@ Expected blob structure:
...
 
+### VLM-as-Judge (experimental)
+
+The viewer can score episodes with a vision-language-model (VLM) judge, reusing the
+`evaluation.vlm_judge` harness. The router mounts only when `VLM_JUDGE_ENABLED=true`;
+the frontend's JudgePanel auto-hides when the backend reports the judge is disabled.
+
+#### What it does
+
+The VLM-as_judge feature samples a fixed number of
```
rezatnoMsirhC: Nit: `VLM-as_judge` should be `VLM-as-judge` (underscore to hyphen).

data-management/viewer/backend/src/api/services/vlm_judge_service.py
```diff
@@ -0,0 +1,75 @@
...
+    cache_dir = Path(config.vlm_judge_cache_dir) if config.vlm_judge_cache_dir else None
+    backend = BackendConfig(
+        kind=config.vlm_judge_backend,
+        model_id=config.vlm_judge_model_id,
+        base_url=config.vlm_judge_base_url,
+        api_key=config.vlm_judge_api_key,
+    )
+    frames = FrameConfig(n_frames=config.vlm_judge_n_frames)
+    method = config.vlm_judge_process_method
+    process_method = method if method in ("gvl", "chronological") else "gvl"
```
rezatnoMsirhC: An invalid `VLM_JUDGE_PROCESS_METHOD` value is silently downgraded to `"gvl"` with no log output, while the same value via the POST body correctly returns HTTP 422. A warning at construction time surfaces misconfiguration early rather than leaving the operator to discover it from unexpected results:

```python
if method not in ("gvl", "chronological"):
    logger.warning(
        "Invalid VLM_JUDGE_PROCESS_METHOD %r; defaulting to 'gvl'. Valid values: gvl, chronological",
        method,
    )
    method = "gvl"
process_method = method
```

rezatnoMsirhC (commented):
[Kelela - outta time (feat. A.K. Paul)](https://www.youtube.com/watch?v=1mLNza10dKs)? Approved.

Leaving a few general comments as I haven't tried the changes locally yet.

data-management/viewer/backend/src/api/services/vlm_judge_service.py
akzaidi: Addressed in 376e7fd

data-management/viewer/README.md
akzaidi: resolved in 376e7fd5f1f695f4c36b8a065708e3366210fde4

data-management/viewer/backend/src/api/services/vlm_judge_service.py
akzaidi: addressed in 376e7fd

data-management/viewer/backend/pyproject.toml
akzaidi: addressed in 376e7fd

data-management/viewer/backend/src/api/routers/vlm_judge.py
akzaidi: Addressed through dd123b7

The synchronous long-lived request was replaced end-to-end with an enqueue + poll job model.

**1. Backend router** (vlm_judge.py)
- A module-level `ThreadPoolExecutor` (`_JUDGE_EXECUTOR`, `max_workers=2`) and an in-process job table `_jobs: dict[str, JudgeJob]` keyed on the SHA256 cache key.
- `POST …/judge` no longer blocks on inference. It resolves the cache key, and:
  - On a cache hit it returns `200` with `job_status="done"` and the result immediately.
  - Otherwise it submits `_run_judgment_job` to the executor, registers a `JudgeJob(status="pending")`, and returns **`202 Accepted`** right away (`response.status_code = status.HTTP_202_ACCEPTED`).
- The background worker transitions the job `pending → running → done | error`, writing the result (or error) back into the shared `_jobs` entry. Backend or model failures are captured as `job.status = "error"`.
- `GET …/judge` reports `idle | pending | running | done | error` plus the result when available. It accepts the `cache_key` query param so the poller targets the exact job.

**2. Frontend polling hook** (use-vlm-judge.ts)
- `useVlmJudgeStatus` uses TanStack Query with `refetchInterval` that polls every `1000ms` while `jobStatus` is `pending` or `running`, and stops (`false`) otherwise — server-driven progress rather than a client ref.

**3. API client** (api-client.ts)
- `runVlmJudge` returns the result directly on a cache hit; otherwise it extracts the `cacheKey` from the `202` response and enters a bounded poll loop (`VLM_JUDGE_POLL_INTERVAL_MS`, capped by `VLM_JUDGE_POLL_TIMEOUT_MS`), calling `fetchVlmJudgeStatus` with the cache key until it sees `done` (returns result), `error` (throws `VLM_JUDGE_FAILED`, 502), or the timeout (throws `VLM_JUDGE_TIMEOUT`, 408).

data-management/viewer/start.sh
akzaidi: addressed in 267919a. `vlm_judge` separated into standalone package with its own requirements and optional dependencies installed via uv.

data-management/viewer/README.md
akzaidi: addressed in 5d2ec45

data-management/viewer/README.md
akzaidi: addressed in 5d2ec45.

data-management/viewer/README.md
akzaidi: addressesd via 8a530f5

evaluation/vlm_judge/cache.py
akzaidi: resolved via 5d2ec45

agreaves-ms (approved):

PR #687, opened 2026-05-12
feat!(infrastructure): support configurable Azure ML compute clusters

fbeltrao:
## Description

Partially addresses #384 (flexible AML compute clusters)

Replace the single Azure ML compute-cluster interface with a flexible `aml_compute_clusters` map. The Terraform root module now passes a map of named compute cluster definitions into `module.platform`, and the platform module creates zero, one, or many Azure ML compute clusters with `for_each` instead of a single optional resource.

Each compute cluster now supports independent settings for VM size, priority, autoscaling bounds, idle scale-down, subnet attachment, public IP exposure, SSH public access, identity type, and optional location. The module continues to block subnet attachment when AML managed network isolation is enabled, but now applies that validation to every configured cluster instead of one shared object.

This PR also migrates outputs from a singular `aml_compute_cluster` value to an `aml_compute_clusters` map, updates example tfvars and Terraform reference docs to the new configuration shape, and expands Terraform tests to cover defaults, conditionals, validation, and outputs for the multi-cluster deployment path.

This change is breaking for Terraform consumers because the previous `should_deploy_aml_compute` and `aml_compute_config` inputs are replaced by `aml_compute_clusters`, and callers must update any code that reads the old singular output.

## Migration Guide

Existing deployments that used `should_deploy_aml_compute = true` and `aml_compute_config` must migrate to `aml_compute_clusters` before applying this change.

```hcl
# Before
should_deploy_aml_compute = true
aml_compute_config = {
  vm_size               = "Standard_NC4as_T4_v3"
  vm_priority           = "LowPriority"
  min_node_count        = 0
  max_node_count        = 1
  scale_down_after_idle = "PT5M"
  cluster_name          = "gpu-cluster"
}

# After
aml_compute_clusters = {
  "gpu-cluster" = {
    vm_size               = "Standard_NC4as_T4_v3"
    vm_priority           = "LowPriority"
    min_node_count        = 0
    max_node_count        = 1
    scale_down_after_idle = "PT5M"
    identity_type         = "SystemAssigned"
  }
}
```

Set `identity_type = "SystemAssigned"` to preserve the previous default compute-cluster identity behavior. Omit `identity_type` to use the new default, `UserAssigned`, with the platform managed identity.

Move existing Terraform state before applying to avoid destroying and recreating the cluster:

```bash
terraform state mv \
  'module.platform.azurerm_machine_learning_compute_cluster.gpu[0]' \
[...]

PR #1035, opened 2026-06-24
fix(infrastructure): align Go output contract and enforce go-tests in CI

### LINKED ISSUE #1034: fix(ci): Go output-contract test skipped in PR CI and soft-failed on main
## Summary

`TestTerraformOutputsContract` (`infrastructure/terraform/e2e`) fails on `main`. The root Terraform output was renamed `aml_compute_cluster` → `aml_compute_clusters` (a map) in #687, but the Go contract struct `InfraOutputs` still references the old singular name. Two CI gaps allowed this to merge and persist undetected for over a month.

## The failing test

- **Command:** `npm run test:go`
- **Failure:** `missing 1 required outputs: [aml_compute_cluster] -- declare these in infrastructure/terraform/outputs.tf or remove from InfraOutputs struct`
- **Root cause:** #687 (commit `7053232`, `feat!`) migrated the singular `aml_compute_cluster` output to the `aml_compute_clusters` map but did not update `infrastructure/terraform/e2e/outputs.go:35`.
- **Fix:** rename the struct field and tag to the plural map shape:

  ```go
  AmlComputeClusters  map[string]any  `output:"aml_compute_clusters"`
  ```

## Why CI did not catch it

1. **PR gate skipped it entirely.** `pr-validation.yml` runs `go-tests` with `changed-files-only: true`. `Invoke-GoTest.ps1` only executes when `*.go` / `go.mod` / `go.sum` change, otherwise it short-circuits to an empty green result. #687 changed only `.tf`, `.hcl`, `.md`, and `.example` files, so the contract test never ran. The test's sole purpose is to catch `outputs.tf` ↔ `outputs.go` drift, yet it is gated on `.go` changes — structurally blind to a `.tf`-only break.
2. **`soft-fail` hides failures.** Both `pr-validation.yml` and `main.yml` invoke `go-tests` with `soft-fail: true` (→ `continue-on-error`). `main.yml` runs the full Go suite on every push to `main` with no path filter, so the contract test *did* run there and failed — but the soft-fail reports the job green, so it never blocks release-please and produces no red signal.

## Proposed remediation

- [ ] Update `InfraOutputs` to match `aml_compute_clusters` (unblocks the failing test).
- [ ] For the `go-tests` job, remove `changed-files-only` (or extend its trigger set to include `infrastructure/terraform/**/*.tf`). The suite runs in under a second with no Azure auth, so gating it saves nothing.
- [ ] Remove `soft-fail: true` for `go-tests` in both `pr-validation.yml` and `main.yml` so contract and test failures block.

## Reproduction

```bash
npm run test:go
```

## Affected files

- `infrastructure/terraform/e2e/outputs.go`
- `.github/workflows/pr-validation.yml`
- `.github/workflows/main.yml`
- `scripts/linting/Invoke-GoTest.ps1`

algattik:
## Summary

`TestTerraformOutputsContract` (`infrastructure/terraform/e2e`) was failing on `main`. #687 renamed the root Terraform output `aml_compute_cluster` → `aml_compute_clusters` (a map) but left the Go contract struct `InfraOutputs` referencing the old singular name. Two CI gaps allowed this to merge and persist undetected.

Closes #1034.

## Changes

**Test fix** — `infrastructure/terraform/e2e/outputs.go`

- `InfraOutputs.AmlComputeClusters map[string]any` tagged `output:"aml_compute_clusters"`, matching the current output.

**CI hardening** — `.github/workflows/pr-validation.yml`, `.github/workflows/main.yml`

- Remove `changed-files-only` from the `go-tests` job so the contract test runs on `.tf`-only changes. The test exists to catch `outputs.tf` ↔ `outputs.go` drift, so gating it on `.go` changes left it structurally blind to the break in #687. The suite runs in under a second with no Azure auth.
- Set `soft-fail: false` so Go test failures block CI instead of reporting green. Previously `main.yml` ran the full suite and the contract test failed there, but `soft-fail: true` swallowed the result.

## Testing

- `npm run test:go` → 19 passed, 0 failed.
- `gofmt` and `go vet` clean.
- Both workflow files validated as well-formed YAML.


katriendg (approved):
Thanks for fixing this! One thing is in future as we eventually do Azure sub testing, we want to use the changed_files setting scoped to the folder changes, but it's OK as is for now.

bindsi (approved):

PR #1038, opened 2026-06-24
security(scripts): scan workflow YAML for unpinned inline pip installs

### LINKED ISSUE #678: security(scripts): extend Test-DependencyPinning.ps1 to scan OSMO workflow YAML for inline pip install
## Problem

`Test-DependencyPinning.ps1` currently scans `requirements*.txt`, `pyproject.toml`, `Pipfile`, `setup.py`, and GitHub Actions workflows for pinning compliance. It does **not** scan OSMO workflow YAML files (`workflows/**/*.yaml`) that contain inline `pip install` commands in task `args` strings.

This creates a blind spot where unpinned runtime dependencies can be introduced via OSMO workflows without any CI detection. Examples:

- `evaluation/sil/workflows/osmo/lerobot-eval.yaml` — inline `uv pip install azureml-mlflow ...`
- `workflows/osmo/replay-azureml.yaml` — previously had inline `pip install` with range specifiers (fixed in #677 by switching to a committed lockfile)

## Proposed Solution

Add a new pattern type (e.g. `shell-inline-pip`) to `Test-DependencyPinning.ps1` that:

1. Scans `workflows/**/*.yaml` (and optionally `**/*.sh`) for `pip install` / `uv pip install` invocations
2. Flags any inline package specifiers that are not exact-version pinned (`==`) or not referencing a `-r requirements*.txt` lockfile
3. Reports violations in the same format as existing ecosystem checks

## Context

- PR #677 addressed `replay-azureml.yaml` by introducing `workflows/osmo/requirements-aml-mirror.txt` (a `uv pip compile` lockfile)
- `lerobot-train.yaml` and `lerobot-eval.yaml` already use committed requirements files for most deps but still have some inline installs
- The `lerobot-train.yaml` pattern (base64-encode a lockfile, decode + `pip install -r` in the task) is the recommended approach

## Acceptance Criteria

- [ ] `Test-DependencyPinning.ps1` detects inline `pip install` in OSMO workflow YAML files
- [ ] Unpinned inline installs are reported as violations
- [ ] Existing OSMO workflows either pass (lockfile-based) or are flagged for remediation

algattik:
## Description

Closes #678 by closing the inline-dependency blind spot in CI, plus two small related fixes uncovered while validating it.

**1. Dependency-pinning linter — inline pip detection (closes #678).** `scripts/security/Test-DependencyPinning.ps1` previously scanned manifests (`requirements*.txt`, `pyproject.toml`, `Pipfile`, `setup.py`) and GitHub Actions, but not the inline `pip install` / `uv pip install` commands embedded in workflow YAML task `args` or in shell entry scripts. This adds a `shell-inline-pip` ecosystem that scans workflow YAML (`**/workflows/**/*.yaml` and `**/workflows/**/*.yml`, including the dot-prefixed `.github/workflows/`) **and** `**/*.sh`, flagging unpinned inline installs.

Compliant patterns: exact `==` pins (including shell-variable pins such as `name=="${VAR}"`), `-r <lockfile>` / `-r -` / `--requirement` lockfile installs, `uv export | uv pip install` pipes, editable `-e .` installs, and the `pip`/`setuptools`/`wheel` build-frontend allowlist. Unpinned `uv run --with` / `uvx --with` is also flagged (with `--with-requirements` treated as lock-derived). A `# pinning-ignore` directive (trailing the install line, or on a dedicated comment line directly above) exempts an intentional non-pin, e.g. an ABI-constrained range. Vendored trees (`.venv`, `node_modules`, `external`, `.git`) are never scanned.

Existing violations are remediated: inline installs in `evaluation/sil/workflows/osmo/lerobot-eval.yaml`, `training/vla/workflows/osmo/groot-train.yaml`, and the `uv` bootstrap in `training/il/scripts/lerobot/azureml-train-entry.sh` are pinned to exact versions (sourced from committed `uv.lock`s where the package is already resolved there). The torch/flash-attn build recipe in `groot-train.yaml` keeps its existing `==` pins and install ordering.

**2. OSMO platform default.** The VLA submission script, `groot-train.yaml`, and the docs table defaulted to `h100gpu_platform`, which is not a platform defined by the repo's OSMO config (`infrastructure/setup/values/osmo-platforms.yaml`). Updated the default to `gpu_platform`, matching the repo-provided platform name already used by `training/il/workflows/osmo/lerobot-train.yaml`.

**3. RL `train.sh` cleanup.** Removed a redundant manual numpy/scipy pre-install in `training/rl/scripts/train.sh`; `numpy==1.26.4` is already pinned in the project and installed via the locked `uv export`. Validated end-to-end on an Isaac Lab GPU run.

Closes #678

## Type of Change

- [x] 🐛 Bug fix (non-breaking change fixing an issue) — the OSMO platform-default correction and the `train.sh` cleanup
- [x] ✨ New feature (non-breaking change adding functionality) — the `shell-inline-pip` linter ecosystem (primary; closes #678)
- [ ] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation

Also `scripts/security/` (the dependency-pinning linter and its Pester tests).

## Testing Performed

Details:

- `npm run test:ps` (Pester): 109 passed, 0 failed — `Get-ShellInlinePipViolations` suites cover pinned/unpinned detection in YAML and `.sh`, shell-variable pins, `uv export` pipes, editable installs, the build-tool allowlist, `uv run --with` and `--with=` specs, URL/path/empty/non-package arguments, dangling line continuations, the `# pinning-ignore` directive (same-line and preceding-comment), and the `install .../uvx` false-positive defense; `Get-FilesToScan` suites cover `.yml` and hidden `.github/workflows` discovery, vendored-tree pruning, and `ExcludePatterns`.
- `scripts/security/Test-DependencyPinning.ps1 -IncludeTypes shell-inline-pip`: 100% compliant with the CI `exclude-paths` (test fixtures excluded).
- `npm run lint:ps` (PSScriptAnalyzer): 0 errors / 0 warnings.
- `npm run lint:sh` (ShellCheck): 52 files, 0 errors / 0 warnings.
- `npm run spell-check`: 0 issues.
- End-to-end: a GR00T-N1.5 fine-tuning workflow was submitted to OSMO on a GPU node and completed the dependency-install and training steps, validating the remediated pins on real hardware; and an Isaac Lab RL training run validated the `train.sh` change.

## Documentation Impact

`docs/security/README.md` now documents the inline-pip scanning, the `shell-inline-pip` type, and the `# pinning-ignore` directive (with an example) alongside the operational-scripts inventory.

## Bug Fix Checklist

- [x] Regression test included (Pester fixtures for compliant and unpinned inline installs, in YAML and `.sh`)

## Checklist


scripts/security/Test-DependencyPinning.ps1
```diff
@@ -158,6 +158,14 @@ $DependencyPatterns = @{
         ValidationFunc = 'Test-ShellDownloadSecurity'
         Description    = 'Shell script downloads must include checksum verification'
     }
+
+    'shell-inline-pip' = @{
+        FilePatterns   = @('**/workflows/**/*.yaml', '**/*.sh')
```
katriendg: `shell-inline-pip` misses `.yml` workflow files. `FilePatterns` only matches `.yaml`, but GitHub Actions workflows use the `.yml` extension — so inline `run: uv pip install …` steps in `.github/workflows/*.yml` are never scanned by this ecosystem. I reproduced it locally: a `.github/workflows/ci.yml` with `uv pip install requests` is not scanned, while an equivalent `workflows/ci.yaml` is flagged. Existing `.yml` inline installs (e.g. `validate-config-schema.yml`) are pinned today, so nothing fails now — but a future unpinned `.yml` inline install would pass silently, which is the exact blind spot this PR closes.

Since the interior-`**` handler uses `Get-ChildItem -Filter` (single pattern), prefer a second entry over a brace glob:

```suggestion
    'shell-inline-pip' = @{
        FilePatterns   = @('**/workflows/**/*.yaml', '**/workflows/**/*.yml', '**/*.sh')
```

scripts/security/Test-DependencyPinning.ps1
```diff
@@ -405,6 +413,168 @@ function Get-PipDependencyViolations {
...
+        # or a dedicated `# pinning-ignore` comment line directly above the install. Lets an
+        # intentional non-pin opt out of this check (e.g. an Isaac-ABI numpy range).
+        $exempt = $hasIgnore -or $prevWasIgnoreComment
+        $prevWasIgnoreComment = $hasIgnore -and $lineText.TrimStart().StartsWith('#')
+
+        if ($lineText -notmatch 'install|uv\s+run|uvx|uv\s+tool\s+run') { continue }
+        if ($exempt) { continue }
+
+        # Split into shell commands on &&, ||, ; (NOT single | so 'uv export | uv pip install' stays intact)
+        $commands = $lineText -split '\s*(?:&&|\|\||;)\s*'
```
katriendg, line 526: Minor robustness note (non-blocking): this command split on `&&` / `||` / `;` doesn't respect shell quoting, so a `;` inside a quoted argument would split a single command into two. Real-world risk on `pip install` lines is low and the per-spec tokenization below already handles quotes, so this is fine to leave — flagging only as a known limitation in case a future install line carries a quoted `;`.

katriendg (commented):
Thank you for this great one, it's going to help towards our goal of more dependency pinning. Leaving a few minor comments which would be nice to tackle before we get this one in.

One additional doc note that can't be left inline (the file isn't in this diff): `docs/security/README.md` (line ~54) still describes the scanner as validating "GitHub Actions, Docker images, and package manifests" — it doesn't mention the new inline-pip scanning, the `shell-inline-pip` ecosystem, or the `# pinning-ignore` directive contributors will now encounter. A short addition there (with a one-line `# pinning-ignore` example) would keep the doc in sync with what the scanner now enforces.

scripts/security/Test-DependencyPinning.ps1
algattik: Fixed in 0dca1bc. Added `'**/workflows/**/*.yml'` to `FilePatterns` as suggested, and fixed a second, deeper cause it surfaced: the interior-`**` handler recurses with `Get-ChildItem -Recurse` but without `-Force`, so it never descends into the dot-prefixed `.github` directory — meaning `.github/workflows/*.yml` (and even `*.yaml`) were silently skipped, and the `.yml` glob alone wouldn't have caught `.github/workflows/ci.yml`. Added `-Force` so recursion enters `.github`; vendored hidden trees (`.git`, `.venv`, `node_modules`, `external`) are still pruned by the relative-path filter. Verified: an unpinned `uv pip install` in `.github/workflows/ci.yml` is now discovered and flagged, vendored trees excluded, all Pester tests green.

scripts/security/Test-DependencyPinning.ps1
algattik: Agreed — leaving as a documented limitation. A quote-aware splitter isn't a clean win here: this function scans raw workflow YAML line-by-line (and `.sh` files), so the split text still carries YAML's own quoting around the embedded shell. Shell-quote-aware splitting would mis-read a double-quoted `run:` scalar — e.g. `run: "uv pip install foo==1 && echo done"` — as one quoted span, stop splitting on the inner `&&`, and flag `echo`/`done` as false positives. Doing it correctly needs YAML-decoding the `run:` value first, which is disproportionate for the low-frequency quoted-`;` case. Keeping the simple split.

katriendg:
Nice work, this is in good shape. One ask before merge which I missed before: can we get the **patch coverage over 80%**?
It's currently sitting at ~78%.

A good chunk of the uncovered lines are **pre-existing CLI/report plumbing** (the `param` block,
threshold-exit handling, and `Export-ComplianceReport` around lines ~1190–1294) that this PR didn't
touch — I'm not asking you to backfill those. The new `Get-ShellInlinePipViolations` function is
already almost fully covered.

The cheap wins are a few unit cases in the existing Pester suite that close the gap inside the new
function (Group A):

- **`--with=SPEC` equals form** of `uv run` (lines 567–568): we only test the space form `--with SPEC`
  today. Add `uv run --with=requests` (expect a violation) and `uv run --with=requests==1.0` (expect none).
- **`$testSpec` defensive guards** (lines 485–491): add specs that should be skipped, e.g.
  `pip install ./local` (path) and `pip install https://example/foo.whl` (URL).
- **Trailing-backslash continuation at EOF** (line 471): a `.sh` fixture whose last line ends with a
  `\` continuation.

That should comfortably push patch coverage past the 80% target. The interior-glob `ExcludePatterns`
branch and `catch` path in `Get-FilesToScan` (lines 714–715, 728) need filesystem-integration tests —
optional, leave them if it's more trouble than it's worth.

algattik:
@katriendg added coverage.

katriendg (approved):
Fantastic - 98% patch coverage!

PR #1041, opened 2026-06-24
fix(training): replace base64 OSMO payload with object-storage url: delivery

### LINKED ISSUE #1036: OSMO IL training pod fails to start: ENCODED_ARCHIVE k8s env var exceeds MAX_ARG_STRLEN (128 KiB) after uv.lock (#995)
## Summary

`training/il/scripts/submit-osmo-lerobot-training.sh` zips `training/il` (plus a few `training/` files), base64-encodes it, and passes it to `osmo workflow submit` as `encoded_archive=<…>`. OSMO injects that value into the workflow pod as a **literal container environment variable** (`training/il/workflows/osmo/lerobot-train.yaml` → `environment: { ENCODED_ARCHIVE: "{{ encoded_archive }}" }`).

On Linux, the kernel caps a single `argv`/`env` string at `MAX_ARG_STRLEN` = 131072 bytes (128 KiB) at `execve`. When the kubelet/runc starts the `lerobot-train` container (entrypoint `/osmo/bin/osmo_exec`), it `execve`s with `ENCODED_ARCHIVE` in the environment. Once that value exceeds 128 KiB the container cannot start:

```
exec /osmo/bin/osmo_exec: argument list too long
```

## Trigger: PR #995 (uv.lock)

The env-var delivery mechanism is long-standing (since #270); the only thing that changed is the **payload size**. PR #995 (`feat(ci): Standardize uv.lock as the single dependency resolution source of truth`) added `training/il/lerobot/uv.lock` (363,001 bytes) to the zipped tree, more than doubling the inline base64 payload and pushing it over the 128 KiB cliff.

| Payload (base64) | Bytes | Linux `execve` result |
| --- | --- | --- |
| Pre-#995 (no `uv.lock`) | 119,764 | OK (~11 KB under the 128 KiB cliff) |
| Linux `MAX_ARG_STRLEN` | 131,072 | — hard limit — |
| Post-#995 (with `uv.lock`) | 256,028 | **FAIL: container start `argument list too long`** |

`uv.lock` contributes **+136,264 bytes** (~53% of the current payload).

## Reproduction

Container start with a 256 KB env var fails on Linux:

```bash
docker run --rm -e ENCODED_ARCHIVE="$(head -c 256264 /dev/zero | tr '\0' a)" busybox:latest /bin/true
# -> exec /bin/true: argument list too long
```

Equivalent on a Linux host:

```bash
env a=$(printf 'a%.0s' {1..130000}) sh -c 'true'   # OK
env a=$(printf 'a%.0s' {1..260000}) sh -c 'true'   # Argument list too long
```

## Notes

- The same single-giant-string root cause would also fail the **local** `osmo workflow submit` invocation (the `encoded_archive=<…>` argv) on a Linux submit host. macOS submit hosts (`ARG_MAX` = 1 MiB) pass it through, so the failure they produce is the **pod** env var on the Linux node, not the local submit.

## Affected paths

- `training/il/scripts/submit-osmo-lerobot-training.sh` (builds and passes the inline base64 archive)
- `training/il/workflows/osmo/lerobot-train.yaml` (`ENCODED_ARCHIVE` delivered as a container env var)

## Suggested directions (not implemented)

- Deliver the payload as a **file** rather than an env var: OSMO's `files:` mechanism (written into the pod by `osmo-ctrl`, not via `execve`) or its dataset/folder-injection path (documented as unbounded size). `entry.sh` would read the archive from a file instead of `${ENCODED_ARCHIVE}`.
- Exclude `uv.lock` from the inline archive and deliver it through the file/dataset channel; the container only consumes it as a file for `uv export`.
- Add a preflight guard in the submit script that fails fast with a clear message when the encoded payload approaches `MAX_ARG_STRLEN`.

## Environment

- Linux `MAX_ARG_STRLEN` = 131072 B per string (kernel constant; applies on AKS nodes).
- Trigger commit: `9dfe22a` (PR #995), merged 2026-06-23.

algattik:
## Description

OSMO workflows shipped their packaged code as a base64-encoded `ENCODED_ARCHIVE` **Kubernetes env var**. On Linux a single env/arg string is capped at `MAX_ARG_STRLEN` = 128 KiB, and once the LeRobot `uv.lock` is part of the archive the value exceeds that, so the workflow container fails to start (`execve` → `argument list too long`). This surfaced after the `uv.lock` grew (it more than doubled the payload to ~256 KiB).

This PR replaces inline base64 delivery with **object-storage injection**:

- The submission packages the code, uploads it with `osmo data upload azure://<account>/<container>/osmo-code/<hash>`, and submits the workflow with a `code_url` parameter.
- The workflow declares an `inputs: - url: "{{ code_url }}"` task input; the pod downloads the archive via its **workload identity** to `{{input:0}}` and unpacks it. No payload size limit, no base64 env var, no oversized argv.
- A shared `stage_and_upload_code()` helper is added to `scripts/lib/common.sh`; the IL training, RL training, and SIL eval (×2) submit scripts and workflows are migrated.

Two adjacent issues found while validating are also fixed:

- GPU workflows lacked an explicit `platform: gpu_platform`, so they failed OSMO resource validation against the CPU-only default platform.
- The LeRobot eval workflow installed dependencies into the image's Python 3.11 (`--system`); LeRobot requires `>=3.12`. It now creates a 3.12 venv (mirroring training) and drops redundant dependency installs already provided by the LeRobot lock.

Closes #1036

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation
- [x] `evaluation/` - SIL evaluation workflows and scripts
- [x] `scripts/lib/` - Shared submission library

## Testing Performed

Validated on a live OSMO cluster (OSMO 6.3):

- **IL training** ran end-to-end: code delivered via `url:`, pod started (env var dropped from 256 KB to <0.5 KB), trained, and registered the model to Azure ML.
- **RL training** passed resource validation and scheduled onto the GPU node (the prior failure was at submit-time validation).
- **`url:` input download** verified in-pod via a minimal probe workflow (download via workload identity to `/osmo/data/input/0`).
- `shellcheck`, `markdownlint-cli2`, and `cspell` clean on changed files.

## Documentation Impact

## Bug Fix Checklist

- [x] Justification for no regression test: the failure modes are GPU/runtime/interpreter-specific (container `execve` env-size limit, OSMO server-side resource validation, container Python 3.12) and pass green on CPU-only PR CI; they are exercised by live OSMO submissions rather than unit tests.

## Checklist

- [x] Commit messages follow conventional commit format (two web-editor "Update …" commits aside)

### DIFF
--- a/evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
+++ b/evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
@@ -77,7 +77,7 @@ EOF
 # Defaults
 #------------------------------------------------------------------------------
 
-workflow="$REPO_ROOT/workflows/osmo/lerobot-infer.yaml"
+workflow="$REPO_ROOT/evaluation/sil/workflows/osmo/lerobot-eval.yaml"
 policy_repo_id="${POLICY_REPO_ID:-}"
 policy_type="${POLICY_TYPE:-act}"
 dataset_repo_id="${DATASET_REPO_ID:-}"
@@ -153,8 +153,7 @@ done
 
 [[ "$use_local_osmo" == "true" ]] && activate_local_osmo
 
-require_tools osmo
-require_tools osmo zip base64
+require_tools osmo zip
 
 # Policy source validation
 if [[ "$from_aml_model" == "true" ]]; then
@@ -216,36 +215,20 @@ if [[ "$config_preview" == "true" ]]; then
 fi
 
 #------------------------------------------------------------------------------
-# Package Runtime Payload
+# Package and Upload Runtime Payload
 #------------------------------------------------------------------------------
 
-TMP_DIR="$SCRIPT_DIR/.tmp"
-ARCHIVE_PATH="$TMP_DIR/osmo-lerobot-inference.zip"
-B64_PATH="$TMP_DIR/osmo-lerobot-inference.b64"
 payload_root="${PAYLOAD_ROOT:-/workspace/lerobot_payload}"
+code_storage_account="${AZURE_STORAGE_ACCOUNT_NAME:-$(get_storage_account)}"
+osmo_container="${OSMO_WORKFLOW_BUCKET:-osmo}"
+[[ -z "$code_storage_account" ]] && fatal "Azure storage account required for code upload (set AZURE_STORAGE_ACCOUNT_NAME or deploy infra)"
 
-info "Packaging LeRobot runtime payload..."
-mkdir -p "$TMP_DIR"
-rm -f "$ARCHIVE_PATH" "$B64_PATH"
-
-(cd "$REPO_ROOT" && zip -qr "$ARCHIVE_PATH" training/il evaluation/sil \
-  -x "**/__pycache__/*" \
-  -x "*.pyc" \
-  -x "*.pyo" \
-  -x "**/.pytest_cache/*" \
-  -x "**/.mypy_cache/*" \
-  -x "**/*.egg-info/*") || fatal "Failed to create runtime archive"
-
-[[ -f "$ARCHIVE_PATH" ]] || fatal "Archive not created: $ARCHIVE_PATH"
-
-if base64 --help 2>&1 | grep -q '\-\-input'; then
-  base64 --input "$ARCHIVE_PATH" | tr -d '\n' > "$B64_PATH"
-else
-  base64 -i "$ARCHIVE_PATH" | tr -d '\n' > "$B64_PATH"
-fi
-
-[[ -s "$B64_PATH" ]] || fatal "Failed to encode archive"
-encoded_payload=$(<"$B64_PATH")
+info "Packaging and uploading LeRobot runtime payload..."
+code_url=$(stage_and_upload_code "$REPO_ROOT" \
+  "azure://${code_storage_account}/${osmo_container}/osmo-code" \
+  training/il evaluation/sil) \
+  || fatal "Failed to stage and upload runtime payload"
+info "Runtime payload uploaded: $code_url"
 
 #------------------------------------------------------------------------------
 # Build Submission Command
@@ -254,7 +237,7 @@ encoded_payload=$(<"$B64_PATH")
 submit_args=(
   workflow submit "$workflow"
   --set-string "image=$image"
-  "encoded_archive=$encoded_payload"
+  "code_url=$code_url"
   "payload_root=$payload_root"
   "policy_repo_id=$policy_repo_id"
   "policy_type=$policy_type"
--- a/evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
+++ b/evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
@@ -107,6 +107,9 @@ subscription_id="${AZURE_SUBSCRIPTION_ID:-$(get_subscription_id)}"
 resource_group="${AZURE_RESOURCE_GROUP:-$(get_resource_group)}"
 workspace_name="${AZUREML_WORKSPACE_NAME:-$(get_azureml_workspace)}"
 
+code_storage_account="${AZURE_STORAGE_ACCOUNT_NAME:-$(get_storage_account)}"
+osmo_container="${OSMO_WORKFLOW_BUCKET:-osmo}"
+
 forward_args=()
 
 #------------------------------------------------------------------------------
@@ -210,6 +213,8 @@ if [[ "$config_preview" == "true" ]]; then
   print_kv "Subscription" "${subscription_id:-<not set>}"
   print_kv "Resource Group" "${resource_group:-<not set>}"
   print_kv "Workspace" "${workspace_name:-<not set>}"
+  print_kv "Storage Account" "${code_storage_account:-<not set>}"
+  print_kv "Code Storage" "azure://${code_storage_account}/${osmo_container}/osmo-code"
   print_kv "Workflow" "$workflow"
   exit 0
 fi
@@ -219,8 +224,6 @@ fi
 #------------------------------------------------------------------------------
 
 payload_root="${PAYLOAD_ROOT:-/workspace/lerobot_payload}"
-code_storage_account="${AZURE_STORAGE_ACCOUNT_NAME:-$(get_storage_account)}"
-osmo_container="${OSMO_WORKFLOW_BUCKET:-osmo}"
 [[ -z "$code_storage_account" ]] && fatal "Azure storage account required for code upload (set AZURE_STORAGE_ACCOUNT_NAME or deploy infra)"
 
 info "Packaging and uploading LeRobot runtime payload..."
--- a/evaluation/sil/workflows/osmo/eval.yaml
+++ b/evaluation/sil/workflows/osmo/eval.yaml
@@ -8,6 +8,7 @@ workflow:
       cpu: 120
       memory: 480Gi
       storage: 200Gi
+      platform: gpu_platform
   tasks:
     - name: isaac-inference
       image: "{{ image }}"
--- a/evaluation/sil/workflows/osmo/eval.yaml
+++ b/evaluation/sil/workflows/osmo/eval.yaml
@@ -14,6 +14,9 @@ workflow:
       command:
         - /bin/bash
         - /tmp/entry.sh
+      inputs:
+        - url: "{{ code_url }}"
+          regex: ".*"
       environment:
         NVIDIA_DRIVER_CAPABILITIES: "all"
         TASK: "{{ task }}"
@@ -22,7 +25,6 @@ workflow:
         VIDEO_LENGTH: "{{ video_length }}"
         ACCEPT_EULA: "Y"
         PRIVACY_CONSENT: "Y"
-        ENCODED_ARCHIVE: "{{ encoded_archive }}"
         PAYLOAD_ROOT: "{{ payload_root }}"
         CHECKPOINT_URI: "{{ checkpoint_uri }}"
         INFERENCE_FORMAT: "{{ inference_format }}"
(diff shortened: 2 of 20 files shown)


training/il/workflows/osmo/lerobot-train.yaml
```diff
@@ -97,18 +99,22 @@ workflow:
...
+            # Unpack training scripts delivered via the OSMO url: input. The code
+            # archive arrives as a downloaded object under the input mount rather
+            # than an env var: a single env string is capped at 128 KiB
+            # (MAX_ARG_STRLEN) and the archive (which ships the lerobot uv.lock
+            # for `uv export`) exceeds that, failing the container execve E2BIG.
             PAYLOAD_ROOT="${PAYLOAD_ROOT:-/workspace/lerobot_payload}"
             mkdir -p "${PAYLOAD_ROOT}"
-            if [[ -z "${ENCODED_ARCHIVE:-}" ]]; then
-              echo "ERROR: ENCODED_ARCHIVE is not set or empty; training payload is required." >&2
+            ARCHIVE_PATH="$(find {{input:0}} -maxdepth 2 -name '*.zip' | head -n1)"
```
katriendg: nit (non-blocking): quote `{{input:0}}` to match `train-dataset.yaml`. Same applies to `ls -laR {{input:0}}` on L112.
```suggestion
            ARCHIVE_PATH="$(find "{{input:0}}" -maxdepth 2 -name '*.zip' | head -n1)"
```

evaluation/sil/workflows/osmo/eval.yaml
```diff
@@ -39,22 +42,24 @@ workflow:
...
-            ARCHIVE_PATH="${TEMP_ARCHIVE:-/tmp/isaac_payload.zip}"
             PAYLOAD_ROOT="${PAYLOAD_ROOT:-/workspace/isaac_payload}"
             INFERENCE_ROOT="${INFERENCE_ROOT:-${PAYLOAD_ROOT}/evaluation/sil}"
 
             mkdir -p "${PAYLOAD_ROOT}"
-            printf '%s' "${ENCODED_ARCHIVE}" | base64 -d > "${ARCHIVE_PATH}"
+            # Code is delivered via the OSMO url: input (object storage), not an
+            # env var: a single env string is capped at 128 KiB (MAX_ARG_STRLEN)
+            # and the archive exceeds that, failing the container execve E2BIG.
+            ARCHIVE_PATH="$(find {{input:0}} -maxdepth 2 -name '*.zip' | head -n1)"
```
katriendg: nit (non-blocking): quote `{{input:0}}` to match `train-dataset.yaml`. Same applies to `ls -laR {{input:0}}` on L60.
```suggestion
            ARCHIVE_PATH="$(find "{{input:0}}" -maxdepth 2 -name '*.zip' | head -n1)"
```

training/rl/workflows/osmo/train.yaml
```diff
@@ -43,17 +46,19 @@ workflow:
...
-            ARCHIVE_PATH="${TEMP_ARCHIVE:-/tmp/isaac_payload.zip}"
             PAYLOAD_ROOT="${PAYLOAD_ROOT:-/workspace/isaac_payload}"
             TRAINING_ROOT="${TRAINING_ROOT:-${PAYLOAD_ROOT}/training/rl}"
 
             mkdir -p "${PAYLOAD_ROOT}"
-            printf '%s' "${ENCODED_ARCHIVE}" | base64 -d > "${ARCHIVE_PATH}"
+            # Code is delivered via the OSMO url: input (object storage), not an
+            # env var: a single env string is capped at 128 KiB (MAX_ARG_STRLEN)
+            # and the archive exceeds that, failing the container execve E2BIG.
+            ARCHIVE_PATH="$(find {{input:0}} -maxdepth 2 -name '*.zip' | head -n1)"
```
katriendg: nit (non-blocking): quote `{{input:0}}` to match `train-dataset.yaml`. Same applies to `ls -laR {{input:0}}` on L59.
```suggestion
            ARCHIVE_PATH="$(find "{{input:0}}" -maxdepth 2 -name '*.zip' | head -n1)"
```

evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
```diff
@@ -216,36 +215,20 @@ if [[ "$config_preview" == "true" ]]; then
...
 #------------------------------------------------------------------------------
-# Package Runtime Payload
+# Package and Upload Runtime Payload
 #------------------------------------------------------------------------------
 
-TMP_DIR="$SCRIPT_DIR/.tmp"
-ARCHIVE_PATH="$TMP_DIR/osmo-lerobot-inference.zip"
-B64_PATH="$TMP_DIR/osmo-lerobot-inference.b64"
 payload_root="${PAYLOAD_ROOT:-/workspace/lerobot_payload}"
+code_storage_account="${AZURE_STORAGE_ACCOUNT_NAME:-$(get_storage_account)}"
```
katriendg: nit (non-blocking): `code_storage_account`/`osmo_container` resolve here, *after* the `--config-preview` block (L196), so preview can't surface or validate the upload target. The other three submit scripts print "Storage Account" in preview — consider hoisting this resolution above the preview block and adding a `print_kv` for parity.

evaluation/sil/workflows/osmo/lerobot-eval.yaml
```diff
@@ -76,15 +79,28 @@ workflow:
...
+            # Unpack runtime payload and install dependencies from workflow manifest.
+            # Code is delivered via the OSMO url: input (object storage), not an
+            # env var: a single env string is capped at 128 KiB (MAX_ARG_STRLEN)
+            # and the archive (which ships the lerobot uv.lock for `uv export`)
+            # exceeds that, failing the container execve E2BIG.
             PAYLOAD_ROOT="${PAYLOAD_ROOT:-/workspace/lerobot_payload}"
             mkdir -p "${PAYLOAD_ROOT}"
-            if [[ -z "${ENCODED_ARCHIVE:-}" ]]; then
-              echo "ERROR: ENCODED_ARCHIVE is not set or empty; runtime payload is required." >&2
+            ARCHIVE_PATH="$(find {{input:0}} -maxdepth 2 -name '*.zip' | head -n1)"
```
katriendg: nit (non-blocking): quote `{{input:0}}` to match `train-dataset.yaml`. Same applies to `ls -laR {{input:0}}` on L101.
```suggestion
            ARCHIVE_PATH="$(find "{{input:0}}" -maxdepth 2 -name '*.zip' | head -n1)"
```

scripts/lib/common.sh
```diff
@@ -82,6 +82,63 @@ activate_local_osmo() {
...
+  fi
+
+  if [[ ! -s "$archive" ]]; then
+    rm -rf "$tmp"
+    error "Code archive is empty"
+    return 1
+  fi
+
+  size="$(wc -c < "$archive" | tr -d ' ')"
+  hash="$(calculate_sha256 "$archive" | cut -c1-16)"
```
katriendg: nit (non-blocking): uploads are content-addressed by `sha256[:16]` of the zip, but `zip` embeds mtimes so it isn't reproducible — identical code produces a new object on every submit and the `osmo-code/` prefix grows unbounded with no TTL/cleanup. Worth a storage lifecycle policy on the prefix, or a cleanup note in the docs.

katriendg (approved):
Thanks @algattik — this is a clean, well-scoped fix. A few things I appreciated:

- **Consistent implementation across the stack.** All four OSMO submit scripts and their workflows moved to the shared `stage_and_upload_code` helper with the `url:` task input, and there's no base64 code-delivery path left anywhere in source.
- **Fixed the dead references in the script defaults.** The eval submit scripts pointed their default workflow at non-existent `workflows/osmo/{infer,lerobot-infer}.yaml`, and `eval.yaml` invoked the wrong `scripts/infer.sh` entrypoint — both corrected to the real paths.
- **Thorough, accurate docs.** The base64 → object-storage narrative is updated consistently across the skill, troubleshooting, recipes, reference, and training docs.

Approving. I left a few non-blocking nits inline (quoting ones as suggestions) — take or leave them.

training/il/workflows/osmo/lerobot-train.yaml
algattik: Done in `4013688` — quoted both `find "{{input:0}}"` and `ls -laR "{{input:0}}"` to match `train-dataset.yaml`.

evaluation/sil/workflows/osmo/eval.yaml
algattik: Done in `4013688` — quoted the `find` and `ls -laR` invocations.

training/rl/workflows/osmo/train.yaml
algattik: Done in `4013688` — quoted the `find` and `ls -laR` invocations.

evaluation/sil/workflows/osmo/lerobot-eval.yaml
algattik: Done in `4013688` — quoted the `find` and `ls -laR` invocations.

evaluation/sil/scripts/submit-osmo-lerobot-eval.sh
algattik: Done in `4013688` — hoisted `code_storage_account`/`osmo_container` resolution above the `--config-preview` block and added `print_kv` for `Storage Account` + `Code Storage`, matching the other three submit scripts.

scripts/lib/common.sh
algattik: Good catch on both. `4013688`:

1. The key is now a SHA over the archive **contents** (per-file sha + sorted relative path), not the mtime-bearing zip, so byte-identical code yields a stable `osmo-code/<hash>` — verified locally that mtime-only changes no longer move the hash, while content changes do.
2. Upload is now skipped when the object already exists. Since `osmo data upload` always overwrites, this is also clobber-safe for concurrent submits (the content-addressed TOCTOU window is benign and noted in a code comment).

Distinct code revisions still accumulate, so rather than adding infra here I documented `osmo data list`/`delete` plus an Azure Storage lifecycle-policy pointer in the troubleshooting guide.

Issue #809, opened 2026-06-01
fix(training): RL training requirements.txt drifts from Isaac Lab 2.3.2 runtime, causing ABI and plugin failures


algattik:
## Training Issue

**Framework:** SKRL
**Issue Type:** Isaac Sim/Lab compatibility
**Python Version:** container ships 3.11; `pyproject.toml` was pinned to `>=3.12`
**GPU Model:** NVIDIA A100 (Standard_NC24ads_A100_v4 — `gpu-cluster`)
**Isaac Sim/Lab Version:** Isaac Lab 2.3.2 (`nvcr.io/nvidia/isaac-lab:2.3.2`)

### Issue Description

`training/rl/requirements.txt` was generated without `--python-version` or `--python-platform` constraints, so it drifted from the actual container interpreter (Python 3.11 on `manylinux_2_28_x86_64`). `train.sh` installs it via `uv pip install --no-deps`, so any internal inconsistency lands verbatim in the container and fails at import time. Four cascading failures resulted:

1. **`pydantic-core 2.47.0 ↔ pydantic 2.13.4` ABI mismatch.** Lock was resolved against Python 3.12 wheels; in the 3.11 runtime `pydantic`'s startup guard raised `SystemError: The installed pydantic-core version (2.47.0) is incompatible with the current pydantic version, which requires 2.46.4`.
2. **`Unsupported URI 'azureml://…'` in MLflow.** PR #541 dropped `azureml-mlflow` (the plugin that registers the `azureml://` URI scheme) while resolving an unrelated pin conflict. `bootstrap_azure_ml` then crashed at `mlflow.set_tracking_uri(workspace.mlflow_tracking_uri)`.
3. **`numpy` circular import in Isaac Sim extensions.** `requirements.txt` re-installed `numpy==2.4.6` over the in-script `<2.0` downgrade in `train.sh`, breaking `omni.kit.pip_archive` numpy and silently disabling `isaaclab_assets`/`isaaclab_tasks`.
4. **`TypeError: azureml_artifacts_builder() got an unexpected keyword argument 'tracking_uri'`.** `mlflow==3.13.0` passes `tracking_uri=` to artifact builders, but `azureml-mlflow` (any version) caps `mlflow-skinny<=3.9.0` and its registered builder doesn't accept the kwarg.

### Training Configuration

```yaml
# Submission
./training/rl/scripts/submit-azureml-training.sh \
  --compute gpu-cluster --instance-type "" \
[...]

PR #1045, opened 2026-06-25
feat(build): add GPU-free import smoke gate for training-domain PRs

### LINKED ISSUE #1043: Add a GPU-free smoke gate to catch runtime/interpreter import regressions on every PR
## Problem

PR CI gives a green check without exercising the environment our code actually runs in.

- Every Python job runs on a generic `ubuntu-latest` image under a generic interpreter (for example `pytest-training.yml` pins Python `3.12`), but the training and evaluation domains run inside pinned runtime containers on their own interpreters — RL and evaluation in the Isaac Lab container (`DEFAULT_ISAAC_LAB_IMAGE` in `scripts/lib/common.sh`), LeRobot in its PyTorch runtime image — each on a pinned `3.11.x`. CI resolves CPU wheels; production installs the CUDA graph. An install, import, or interpreter break can pass CI and only surface when the job launches on the cluster.
- Test execution is gated by a path-filter (`changes` job in `pr-validation.yml`). When that filter misfires, jobs are silently skipped and the run reports green while testing nothing.

These are not hypotheticals. Each class below has already shipped a regression that CPU-only CI did not catch:

| Incident | Class | What CPU CI missed |
| --- | --- | --- |
| #809 | Interpreter / ABI | RL requirements drifted from the Isaac Lab 2.3.2 runtime, causing ABI and plugin failures at load time. |
| #790 | Resolution / runtime | A LeRobot submission fix whose interpreter/version-conflict half is gate-catchable (the Azure Blob data-source change is orthogonal). |
| #691 | Gating integrity | A broken regex in the PR path filter gated jobs incorrectly. |
| #547 | Gating integrity | A domain folder restructure left tests unrun until repaired. |
| #958 | Dependency resolution | A `torch` version bump (2.10 → 2.11 security update); the resolution half is cheaply catchable on CPU (the device/ABI half needs a GPU and is out of scope here). |

## Proposal

A two-depth, GPU-free smoke gate that runs on standard runners and blocks merge through the existing single required check (`pr-validation-summary`).

### CPU import smoke — fast, every PR (seconds)

For each changed domain: create a CPU virtual environment, install the CPU `torch` wheel as a discrete step (`uv pip install torch --index-url https://download.pytorch.org/whl/cpu`) so the multi-gigabyte CUDA build is never pulled, then install the rest of the project, import its top-level package, and run its CLI `--help` / `--config-preview`. (Install torch separately rather than folding the index into the project install: a project-wide `--index-url` would replace PyPI and break every non-torch dependency, so where the smoke env needs both, follow the repo convention of an additive `--extra-index-url`.) This reuses checks already in the pipeline (uv-lock consistency, YAML and schema validation, shellcheck) and adds only the import-and-help probe.

> [!NOTE]
> State the limit plainly: this runs on CPU wheels, so it validates a different dependency graph than the production CUDA one. It catches import, resolution, and interpreter-syntax errors — not the production CUDA resolution.

### Runtime-image smoke — path-gated (minutes)

Pull the changed domain's runtime container image — the Isaac Lab container for RL and evaluation (`DEFAULT_ISAAC_LAB_IMAGE`), the PyTorch runtime image for LeRobot, both anonymously pullable — and, inside it, reinstall the PR's committed lock exactly as the job does, then exercise the changed domain on the real interpreter from the checked-out source tree. Mirror production: `--no-emit-project` deliberately omits the first-party package (production runs from a mounted tree, not an installed wheel), so put the source on `PYTHONPATH` rather than importing an installed package:

```bash
uv export --frozen --no-hashes --no-emit-project | uv pip install --system --no-deps -r -
PYTHONPATH=. python -m <changed_domain.import_check>   # dedicated import-check entrypoint, or `--help`
```

This is the depth that mirrors production and deterministically catches the interpreter/ABI-at-import class (the #809 class): importing the domain loads its real dependency graph (for example Isaac / SKRL / gym for RL, the LeRobot stack for IL) on the pinned interpreter. Its limits: it needs enough runner disk to pull a multi-gigabyte image (see implementation notes), and it cannot prove CUDA, Vulkan, MIG, or a real training loop.

### One fail-safe required check

- Add a `smoke` job whose CPU import depth runs **unconditionally** (no path gate) and whose runtime-image depth is path-gated. In `pr-validation.yml`, append `smoke` to the `pr-validation-summary` `needs:` list so it joins the single required check. Skipped needed-jobs count as a pass, so only the runtime-image depth may ever skip — never the baseline.
- Fail-safe gating: the CPU baseline runs on every PR regardless of path classification, and the path classifier must **fail** (not silently skip) on unexpected or empty input. This reduces the silent-skip failure mode behind #691 and #547 — a buggy filter cannot make the whole gate vanish, because the unconditional baseline still runs.

## Scope

In scope:

- CPU import-and-help smoke for changed domains on every PR.
- Runtime-image import smoke, path-gated to the changed domain, with a runtime cap.
- Wiring both into the existing required-check aggregator on PRs and on `main`.

Non-goals:

- Any GPU execution (CUDA, Vulkan, MIG) or a real training/inference loop.
- Building or validating the evaluation container image (separate concern).
- Repairing specific pre-existing dependency desyncs (track separately).
- Dependency-update grouping or merge automation.

## Acceptance criteria

- Every PR runs the CPU import-and-help smoke for the changed domains on a CPU runner; a failure blocks merge via `pr-validation-summary`.
- The runtime-image smoke runs for the changed domain, path-gated, behind the same required check, with a runtime cap.
- A deliberately broken import (an unresolved or incompatible dependency) fails the gate; verified on a throwaway branch against a #809-style break.
- Gating cannot silently skip: the CPU baseline runs unconditionally, and a misconfigured or empty path classification fails rather than skips.
- Runtime budget: the CPU smoke under roughly two minutes; the runtime-image smoke capped (for example fifteen minutes) and triggered only by relevant changes.
- The runtime image pulls and runs within the chosen runner's disk budget (reclaim space first or size the runner accordingly).

## Implementation notes

- New reusable workflow `.github/workflows/smoke-cpu.yml` with two jobs (`import-smoke`, `image-smoke`).
- New script for the runtime-image import (for example under `shared/ci/` or `scripts/`), invoked inside the container.
- Reclaim runner disk before pulling the image (for example a free-disk-space step) or use a larger runner; the image exceeds the free disk on a default `ubuntu-latest` runner.
- Edit `pr-validation.yml`: add the `smoke` caller job, extend the `changes` job outputs for per-domain runtime-image gating, and append `smoke` to `pr-validation-summary.needs`. `main.yml` has no aggregator or path-filter job — add the smoke job there to run unconditionally.
- Select the runtime image per changed domain, each taken from that domain's existing workflow or submit scripts (`DEFAULT_ISAAC_LAB_IMAGE` in `scripts/lib/common.sh` for RL and evaluation; the PyTorch runtime image pinned in `training/il/workflows/osmo/lerobot-train.yaml` for LeRobot) rather than hard-coding tags.

algattik:
# Pull Request

## Description

Adds a GPU-free import smoke gate that catches dependency-resolution, import, and interpreter/ABI regressions on every PR, before they reach a GPU job.

PR CI runs Python jobs on a generic `ubuntu-latest` interpreter, but training and evaluation run inside pinned runtime containers (Isaac Lab on Python 3.11; LeRobot on a PyTorch image). The existing `pytest-training`/`evaluation-pytests` jobs mock the heavy framework imports and install off-lock, so a break that only manifests on the real interpreter or the committed lock passes CI green and surfaces on the cluster.

Two depths, both on standard runners, behind the single required `pr-validation-summary` check:

- CPU import smoke — runs for every domain on every PR (unconditional baseline). Installs the domain's committed lock with CPU torch wheels (`uv --torch-backend cpu`, `--no-deps`) and imports the framework stack plus the first-party entrypoint.
- Runtime-image smoke — path-gated to the changed training domain. Pulls the domain's production container, installs the committed lock exactly as the job does, and imports on the real interpreter.

| Domain | Python | Runtime image | CPU smoke | Runtime-image smoke |
|--------------|--------|-----------------------------------------|-----------|---------------------|
| `rl` | 3.11 | Isaac Lab (`DEFAULT_ISAAC_LAB_IMAGE`) | yes | yes |
| `il` | 3.12 | PyTorch (`lerobot-train.yaml` default) | yes | yes |
| `evaluation` | 3.12 | shares the Isaac Lab SiL runtime | yes | no |

`shared/ci/smoke-import.sh` is the inner probe (linux/x86_64); `shared/ci/smoke-image.sh` runs it in a linux/amd64 container so any host — including macOS — can reproduce CI locally. Usage is documented in `shared/ci/README.md`.

Image references come from their source of truth (`scripts/lib/common.sh`, `lerobot-train.yaml`); GitHub Actions are SHA-pinned; inline installs use the lock-derived `uv export | uv pip install --requirement -` pattern.

Closes #1043

## Type of Change

## Component(s) Affected

`.github/workflows/` (CI) and `shared/ci/` (new scripts).

## Testing Performed

- All five documented commands run locally (Docker, linux/amd64), each green: CPU smoke for `rl`/`il`/`evaluation`, runtime-image smoke for `rl` (Isaac Lab) and `il` (PyTorch).
- Negative tests reproducing the gate going red on the incident classes it targets:
  - Interpreter mismatch: installing the LeRobot lock on Python 3.11 fails resolution (`lerobot==0.5.1` requires `>=3.12`).
  - ABI-at-import: a `--no-deps` install of drifted pins succeeds, while the import fails (`numpy.core.multiarray failed to import`) — confirming the import step, not the install, is what catches ABI drift.
- `shellcheck`, `actionlint`, `markdownlint`, `cspell`, and the markdown table formatter pass on all changed files.

## Documentation Impact

`shared/ci/README.md` documents local and CI usage.

## Checklist

### Scope notes

- No GPU execution (CUDA, Vulkan, MIG) or training loop — the gate proves install, import, and interpreter/ABI, not runtime behavior.
- The RL probe loads the pip-package ABI surface but not Isaac's `omni`/`isaaclab` plugins (deferred in code; they require the Isaac app). Full plugin-load coverage remains a GPU e2e concern.
- Evaluation runtime-image smoke is deferred (no dedicated runtime image); evaluation still gets the CPU import smoke.


.github/workflows/smoke-cpu.yml
```diff
@@ -0,0 +1,100 @@
...
+        description: 'Run the RL runtime-image smoke (training/rl changed)'
+        required: false
+        default: false
+        type: boolean
+      il:
+        description: 'Run the LeRobot runtime-image smoke (training/il changed)'
+        required: false
+        default: false
+        type: boolean
+      soft-fail:
```
katriendg: 💡 **Optional (Low) — `soft-fail` is defined but never wired.**

The reusable workflow threads `soft-fail` through `continue-on-error` on all three step types, but no caller (`pr-validation.yml` / `main.yml`) passes it, so it's dead in CI today. If it's intended for a future soft-launch/canary path, a one-line comment here noting who is expected to set it would stop a later reviewer pruning it as unused; otherwise it could be dropped until a caller needs it. Non-blocking.

.github/workflows/pr-validation.yml
```diff
@@ -50,6 +52,8 @@ jobs:
           echo "Computed pester=$pester_match"
           {
             echo "training=$(match '^training/')"
+            echo "rl=$(match '^(training/rl/|shared/ci/|scripts/lib/common\.sh|\.github/workflows/smoke-cpu\.yml)')"
```
katriendg, line 59: 💡 **Optional (Low) — AC-fidelity note on the path classifier.**

Issue #1043 says an empty/misconfigured path classification should *fail* rather than skip; the `match()` helper (line 50) returns `false` on empty input rather than failing. In practice the unconditional CPU baseline already neutralizes the silent-skip risk — a wrong classifier here only suppresses the path-gated runtime-image depth, never the baseline — so the intent is covered. Flagging only so we can consciously accept the deviation, or add a narrow sanity check if you'd prefer literal AC compliance (watch out for legitimately-empty diffs on some events). Non-blocking.

.github/workflows/smoke-cpu.yml
```diff
@@ -0,0 +1,100 @@
...
+
+jobs:
+  # Unconditional fail-safe baseline: import every domain on CPU on every PR.
+  # DO NOT add an `if:` or path filter here -- a skipped required job counts as a
+  # pass, so path-gating this would let the gate go green while testing nothing
+  # (the silent-skip failure that has shipped before). Only the runtime-image
+  # depth below is path-gated. See shared/ci/README.md (Design) for the rationale.
+  import-smoke:
+    name: CPU Import Smoke (${{ matrix.domain }})
+    runs-on: ubuntu-latest
```
katriendg, line 33: 🔴 **Required before approval (Medium) — add `timeout-minutes` to the smoke jobs.**

Issue #1043 lists a runtime cap as an acceptance criterion (*"the runtime-image smoke capped (for example fifteen minutes)"* plus the CPU/runtime-image runtime budget). None of the smoke jobs set `timeout-minutes`, so they inherit GitHub's **360-minute** default. A hung image pull, a stuck `uv` resolve, or an import that blocks could burn up to 6 hours of runner time per job before the gate reports red — and the multi-gigabyte runtime-image jobs are the highest-risk.

Suggested cap for the CPU baseline here:

```suggestion
    runs-on: ubuntu-latest
    timeout-minutes: 10
```

Please also cap the two runtime-image jobs (they pull large images, so a longer-but-bounded budget):

- `image-smoke-rl` (around line 63): add `timeout-minutes: 15` under `runs-on:`
- `image-smoke-il` (around line 84): add `timeout-minutes: 15` under `runs-on:`

katriendg (commented):
Great work, I want to comment that doing the Copilot PR review, it came out with this comment "PR #1045 cleanly implements the GPU-free smoke gate from issue #1043, stays inside Phase 1 scope (no bleed into Phase 0/2/3), and **is one of the better-engineered CI changes I've reviewed here**". Not sure where it gets it's reviewed here memory but I'll take it :-)

The approach to run the CPU import and the Isaac Lab runtime-image smoke test is really well thought out.

What stood out:

- **The fail-safe design is the right one.** Running the CPU import baseline unconditionally (no path gate) directly closes the silent-skip class behind #691/#547, and you defended the invariant in *both* the code (`DO NOT path-gate`) and the README Design section so it survives future edits.
- **`--no-deps` + load-bearing import is correctly reasoned.** Installing the exact committed lock (so pyproject `override-dependencies` aren't discarded) and letting the *import* catch ABI drift is exactly the mechanism that catches the #809 class — and you proved it with negative tests rather than claiming it.
- **Security hygiene is exemplary.** Every action SHA-pinned with a version comment, `persist-credentials: false`, least-privilege `contents: read`, image refs read from their source of truth (no hard-coded tags), and a checksum-verified `uv` bootstrap.
- **It's locally reproducible.** The `smoke-image.sh` Docker wrapper plus the README make the gate runnable on any host including macOS — a genuinely thoughtful touch.

One required change before approval (runtime cap), plus two small optional notes below.

.github/workflows/smoke-cpu.yml
algattik: Done in `bbfd7c3` — added `timeout-minutes: 10` to the CPU baseline and `timeout-minutes: 15` to both runtime-image jobs (`image-smoke-rl`, `image-smoke-il`), per the #1043 runtime-cap AC.

.github/workflows/smoke-cpu.yml
algattik: Dropped it rather than documenting it. No caller (`pr-validation.yml` / `main.yml`) wires `soft-fail`, and there is no concrete soft-launch scenario in mind, so carrying the input plus its `continue-on-error` wiring is YAGNI. Removed in `bbfd7c3`; it can return when a caller actually needs it.

.github/workflows/pr-validation.yml
algattik: Consciously accepted, and now documented at the `match()` helper in `bbfd7c3`. The comment records that a `false` on an empty/non-matching diff is safe: the unconditional CPU import baseline runs regardless of these outputs, so a misclassification can only narrow the path-gated runtime-image depth, never silently skip the gate.

katriendg (approved):
Thank you, look forward to seeing this one run in the CI phase. It will be a big step already.

PR #1060, opened 2026-06-25
test(training): add live e2e suite for OSMO and Azure ML submission paths

### LINKED ISSUE #1059: Live end-to-end test suite for OSMO and Azure ML submission paths
## Summary

Add a live (cloud-backed) end-to-end test suite that exercises the real submission paths for the training and evaluation systems. Each test submits a real workflow/job, waits for it to complete, and asserts on observable outcomes (task success, MLflow tracking, expected outputs).

## Motivation

End-to-end coverage previously existed only for the RL submission path and depended on external HuggingFace artifacts, so it could not run standalone and did not catch regressions in the IL, VLA, or evaluation paths.

## Coverage

- OSMO: RL training, IL (LeRobot/ACT) training, IL evaluation, VLA (GR00T) fine-tuning
- Azure ML: RL training, IL training

## Goals

- Self-contained: tests generate and stage their own datasets and policy, with no external tokens or pre-staged artifacts required.
- Observable failures: live workflow/pod logs surface in the test output so failures are diagnosable without leaving the run.
- Fast feedback: minimal step counts and synthetic fixtures keep runs short while still exercising the real submission, scheduling, and output paths.

## How to run

```bash
# A single submission path
uv run pytest -vv -s -m e2e tests/e2e/test_e2e_osmo_il_eval.py

# Full e2e suite
uv run pytest -vv -s -m e2e tests/e2e/
```

Requires GPU nodes and OSMO/Azure ML access.

algattik:
## Summary

Adds a live (cloud-backed) end-to-end test suite that exercises the real submission paths for the training and evaluation systems, and bundles the product fixes those live runs surfaced. Each test submits a real OSMO workflow or Azure ML job, waits for completion, and asserts on observable outcomes (task success, MLflow tracking, code snapshot, checkpoint output). Tests generate and stage their own datasets and policy, so the suite runs standalone with no external tokens or pre-staged artifacts.

Closes #1059.

## Why

End-to-end coverage previously existed only for the RL submission path (`tests/e2e/test_e2e_training.py`) and depended on external HuggingFace artifacts, so it could not run standalone and did not catch regressions in the IL, VLA, or evaluation paths. Standing up the IL and VLA paths under real submission also surfaced several latent bugs in the training/evaluation scripts and dependency locks, fixed here so the paths actually pass end to end.

### 1. Live E2E test suite (`tests/e2e/`)

Replaces the single `test_e2e_training.py` with one file per system under test, each marked `@pytest.mark.e2e`:

| Test file | Path exercised |
| --- | --- |
| `test_e2e_osmo_rl_training.py` | OSMO Isaac/SKRL RL training |
| `test_e2e_osmo_il_training.py` | OSMO LeRobot/ACT IL training |
| `test_e2e_osmo_il_eval.py` | OSMO LeRobot IL evaluation |
| `test_e2e_osmo_vla_finetune.py` | OSMO GR00T VLA fine-tuning |
| `test_e2e_aml_rl_training.py` | Azure ML RL training |
| `test_e2e_aml_il_training.py` | Azure ML IL training |

### 2. Shared harness

- `_osmo.py` — OSMO submit/poll/cancel, workflow-task success assertions, and `kubectl logs -f` pod-log streaming for the running task (wired into every OSMO test so live failures are diagnosable in the test output).
- `_aml.py` — Azure ML job submit/poll/cancel, code-snapshot and checkpoint-output assertions.
- `_mlflow.py` — MLflow run resolution and metric/param presence assertions for both training and evaluation.
- `_common.py` / `conftest.py` — shared `e2e_name` helper (deduplicated from `_aml`), polling utilities, fixtures, and logging.

### 3. Self-contained fixtures

- `_lerobot_dataset.py` — generates a minimal LeRobot v3.0 dataset and stages it to blob.
- `_vla_dataset.py` — generates a minimal GR00T dataset and stages it to blob.
- The IL eval path mints its base policy in-container from LeRobot's built-in ACT architecture (sized to the synthetic dataset) via a new `--builtin-policy` source, removing the dependency on an external HuggingFace policy. See `submit-osmo-lerobot-eval.sh` and `lerobot-eval.yaml`.

### 4. Product fixes surfaced by the live runs (review these closely)

These are behavior changes to the shipped scripts/workflows/locks, not just test code:

- **LeRobot `--log-freq` plumbing** (`submit-osmo-lerobot-training.sh`, `lerobot-train.yaml`): LeRobot only prints the parseable metric line every `log_freq` steps (default 200), so short runs logged zero `train/*` metrics to MLflow. A configurable `--log-freq` is now threaded through the submit script and workflow.
- **Always-on `--dataset.video_backend=pyav`** (`lerobot-train.yaml`): previously only set on the blob path; moved to the common args so local/HuggingFace datasets decode consistently.
- **Remove incorrect `cuda-bindings==13.3.1` pin** (`training/il/lerobot/pyproject.toml`, `uv.lock`): the pin pulled a CUDA 13 runtime dependency while the cu12 torch wheels bundle CUDA 12, breaking the resolved environment; removing it also substantially shrinks the lock.
- **HuggingFace token via environment** (`lerobot-eval.yaml`, `lerobot-eval-entry.sh`): the eval auth step reads `HF_TOKEN` from `os.environ` inside Python instead of interpolating it into a `python3 -c "...login(token='${HF_TOKEN}')"` argument, keeping the token out of the process argument list and tolerant of quote characters.
- **Deterministic eval camera-key selection** (`lerobot-eval.yaml`, `run_evaluation.py`, `run-local-lerobot-eval.py`): the eval image key is discovered by feature dtype (`video`/`image`), preferring an `observation.images.*` key so multi-camera datasets resolve deterministically and singular-`observation.image` datasets (the built-in/pusht policy) are handled correctly.

### 5. Build / config

- Dev dependencies: add `imageio` + `imageio-ffmpeg` for synthetic-video encoding in the dataset fixtures (`pyproject.toml`, `uv.lock`).
- `.cspell.json`: allow the technical terms `pixelformat` (imageio writer kwarg) and `getfixturevalue` (pytest API).

### 6. Docs

- `CONTRIBUTING.md` — per-file run instructions for the new suite.
- `docs/evaluation/osmo-evaluation.md`, `docs/reference/workflow-templates-osmo.md` — document the `--builtin-policy` source and the `builtin_policy` workflow parameter.

### 7. E2E robustness on idle (scale-from-zero) clusters

- The e2e RL submission requests `20` CPU / `180Gi` (via `tests/e2e/_osmo.py`) to fit one `Standard_NC24ads_A100_v4` A100 node; the shipped workflow/script defaults are unchanged.
- `ensure_gpu_nodes_available` accepts an autoscaling GPU node pool (`max_count > 0`) in addition to a registered GPU node, so GPU e2e tests are not skipped on scale-from-zero clusters where GPU nodes are provisioned on demand.
- The OSMO "started" timeout is 60 minutes to absorb a cold scale-from-zero path (GPU node provisioning + NVIDIA GPU-operator cold-start + isaac-lab image pull) before the workflow reports `RUNNING`.

## How to test

```bash
# A single submission path
uv run pytest -vv -s -m e2e tests/e2e/test_e2e_osmo_il_eval.py

# Full e2e suite
uv run pytest -vv -s -m e2e tests/e2e/
```

Requires GPU nodes and OSMO/Azure ML access; `kubectl` on the runner enables live pod-log streaming for the OSMO tests.

## Reviewer guide

- Start with the six `test_e2e_*.py` entry points to see the flow and assertions per path.
- Then review section 4 (product fixes) — these affect shipped training/eval behavior independent of the tests.
- The fixtures in `_lerobot_dataset.py` / `_vla_dataset.py` are pure-Python generators (no framework import), so they can run on the host that drives the suite.

## Notes

- The suite makes live cloud submissions; it is gated behind the `e2e` marker and is not part of the default unit-test run.


training/rl/scripts/submit-osmo-training.sh
```diff
@@ -92,8 +92,8 @@ payload_root="${PAYLOAD_ROOT:-/workspace/isaac_payload}"
 backend="${TRAINING_BACKEND:-skrl}"
 
 gpu="${OSMO_GPU:-1}"
-cpu="${OSMO_CPU:-30}"
-memory="${OSMO_MEMORY:-400Gi}"
+cpu="${OSMO_CPU:-20}"
+memory="${OSMO_MEMORY:-180Gi}"
```
katriendg: This lowers the **default** RL resource request for every user of these scripts, not just the e2e smoke: cpu `30 → 20` and memory `400Gi → 180Gi` (same change in `submit-osmo-dataset-training.sh` and the `train.yaml` / `train-dataset.yaml` default-values). The change is justified — `Standard_NC24ads_A100_v4` is 24 vCPU / 220 GiB, so the previous `30`/`400Gi` defaults could never schedule on it — but it's a real reduction in default headroom.

Can you confirm 180Gi is sufficient for a production-scale RL run, not just the 2-iteration smoke? If production training needs more, consider keeping the higher value as the script default and overriding down to `180Gi` only in the e2e submission path, so a real job doesn't silently get less memory than before.

evaluation/sil/workflows/osmo/lerobot-eval.yaml
```diff
@@ -585,14 +638,10 @@ workflow:
...
-                for key, feat in info.get("features", {}).items():
-                    if key.startswith("observation.images."):
-                        image_key = key
-                        break
-                if image_key is None:
-                    image_key = "observation.images.color"
+                # Discover image key from dataset features (camera = video/image feature)
+                features = info.get("features", {})
+                video_keys = [k for k, v in features.items() if v.get("dtype") in ("video", "image")]
+                image_key = video_keys[0] if video_keys else "observation.images.color"
```
katriendg: This rewrites camera-key discovery from a name-prefix match (`observation.images.`) to a `dtype in ("video","image")` match. It's more robust in general, but it's a **silent behavior change to every eval run**, not only the built-in path. For a multi-camera dataset it picks `video_keys[0]`, which is dict-insertion-order dependent.

Can you confirm this resolves to the same key as before for the existing eval datasets (i.e. the first `video`/`image` feature is the one previously selected by the `observation.images.` prefix)? If multi-camera ordering is ever a concern, a deterministic tiebreak (e.g. prefer an `observation.images.*` key, then fall back to first video/image) would remove the ordering dependency.

katriendg (approved):
Thanks for this! The self-contained synthetic fixtures (no external HuggingFace dependency) and the in-container `--builtin-policy` mint are exactly the right call for a standalone suite, and the harness is genuinely clean: the generic `wait_for_status`, the best-effort `TaskPodLogStream` that can't break the test it observes, and consistent `addfinalizer` cleanup for both local work dirs and staged blobs. Graceful `pytest.skip` on missing `az`/`terraform`/`osmo`/`kubectl`/GPU nodes is the correct shape for an opt-in live suite.

I'm approving on the test infrastructure. My only open items are two quick confirmations on the shipped product changes bundled in (the resource-default and image-key changes are user-facing, not test-internal) — details inline. Nothing blocking once those are confirmed.

bindsi (approved):
Thank you for this — it's a genuinely high-quality PR. The honesty in calling out the section 4 product fixes separately made review much easier, the in-container policy minting and the `--builtin-policy` validation are carefully guarded, and standing the suite up as self-contained synthetic fixtures (no external HF tokens or pre-staged artifacts) is a real step up in coverage. Nicely done. 🙌

Approving. One non-blocking, out-of-scope nit to consider while you're in the file:

**`lerobot-eval.yaml` — HF auth token handling (pre-existing, not introduced here):**

The unchanged HF-auth block interpolates `${HF_TOKEN}` directly into an inline `python3 -c "…login(token='${HF_TOKEN}'…)"`, which exposes the token in the process arg list (visible via `ps`) and breaks if the token ever contains a quote character. Since the file is already heavily edited, this is a reasonable moment to switch to passing the token via the environment instead:

```python
login(token=os.environ["HF_TOKEN"], add_to_git_credential=False)
```

Entirely optional and safe to defer to a follow-up — not a blocker for merge.

training/rl/scripts/submit-osmo-training.sh
algattik: Good call — reverted the default change. `submit-osmo-training.sh`, `submit-osmo-dataset-training.sh`, `train.yaml`, and `train-dataset.yaml` are back to `30` CPU / `400Gi` (unchanged from `main`), so no production run silently gets less headroom than before.

The reduced `20` CPU / `180Gi` now lives only in the e2e submission path (`tests/e2e/_osmo.py`), which is the only path this suite exercises, sized to fit one `Standard_NC24ads_A100_v4` GPU node (24 vCPU / 220 GiB). Right-sizing the production default is left as a separate change that warrants a production-scale test. (commit 8b1a29d)

evaluation/sil/workflows/osmo/lerobot-eval.yaml
algattik: Added the deterministic tiebreak you suggested, across all three eval paths that select a single image key (the OSMO inline script here plus `run_evaluation.py` and `run-local-lerobot-eval.py`, kept in sync): prefer an `observation.images.*` key, then fall back to the first video/image feature, then `observation.images.color`.

Confirmed it resolves to the same key as before for existing datasets: for any dataset carrying `observation.images.*` keys the tiebreak selects the identical key the old prefix match did (including the multi-camera first-key case). The dtype-based switch is also required for the built-in/pusht datasets, which expose `observation.image` (singular) — the old `observation.images.` prefix match missed those and fell back to a non-existent `observation.images.color`. (commit b1c3678)

algattik:
I've addressed the comments, and run the e2e OSMO tests to validate the updates. I've also caught a bug in the scale-from-zero pool feature and addressed it.

katriendg (approved):

PR #1066, opened 2026-06-29
fix(data): repair dataviewer compose build and guard it in CI

### LINKED ISSUE #1065: dataviewer: docker compose up --build fails on main (backend + frontend)
## Summary

`docker compose up --build` in `data-management/viewer/` fails on `main` — both service images fail to build, so the dataviewer stack cannot be brought up locally.

## Repro

```
cd data-management/viewer
docker compose up --build
```

## Backend

`python:3.13.14-slim` now ships pip 26, which removed the command-line `--hash` option. The Dockerfile's `pip install --require-hashes uv==0.10.9 --hash=...` aborts:

```
no such option: --hash
```

## Frontend

`npm ci` fails `ERESOLVE` (`eslint@10` vs `eslint-plugin-jsx-a11y@6.10.2` peer ≤9). The repo carries an `.npmrc` with `legacy-peer-deps=true`, but the Dockerfile copies only `package.json`/`package-lock.json`, so the build never sees it.

## Expected

`docker compose up --build` builds both images and serves the dataviewer. No CI rebuilds these images today, so the regression is invisible — a build smoke is warranted.

algattik:
Closes #1065.

`docker compose up --build` in `data-management/viewer/` failed on `main` — both images broke, and no CI rebuilt them, so it shipped silently. This repairs both builds, hardens the images, and adds a build smoke so the regression can't recur.

## Build fixes (the bug)

- **backend** — base `python:3.13.14-slim` now ships pip 26, which removed CLI `--hash`, so `pip install --require-hashes uv==... --hash=...` aborted (`no such option: --hash`). uv now arrives via `COPY --from=ghcr.io/astral-sh/uv` (digest-pinned), and deps install with `uv sync --frozen`.
- **frontend** — `npm ci` failed `ERESOLVE` (eslint 10 vs jsx-a11y peer ≤9); the repo's `.npmrc` (`legacy-peer-deps=true`) wasn't copied. Now copied before `npm ci`.

## Image hardening (prevents the class of failure)

- **non-root** — backend runs as `appuser`; frontend serves via `nginxinc/nginx-unprivileged` on 8080.
- **frozen lock** — `uv sync --frozen` + `UV_COMPILE_BYTECODE`/`UV_LINK_MODE=copy`; runs `.venv/bin/uvicorn` directly (no per-start `uv run`).
- **`.dockerignore`** — added for backend and frontend (no `.venv`/`node_modules`/secrets in context).
- **healthchecks** — frontend gains one; backend retained.

## CI guard

- `shared/ci/smoke-docker.sh` runs `docker compose config` + `build`, exercising both images plus compose syntax.
- Wired into `smoke-cpu.yml` (new `dataviewer` input/job), path-gated on backend/frontend/compose in `pr-validation.yml`, always-on in `main.yml`.

## Compose

- frontend mapped `5173:8080` (unprivileged port); `restart: unless-stopped` on both. Host port unchanged, so CORS still matches.

Validated: actionlint + shellcheck pass; both images build and run non-root (backend `appuser`, frontend serves 200).


data-management/viewer/frontend/Dockerfile
```diff
@@ -23,7 +23,9 @@ ENV VITE_AZURE_TENANT_ID=${VITE_AZURE_TENANT_ID}
 RUN npm run build
 
 # ─── Serve stage ─────────────────────────────────────────────────────────────
-FROM nginx:1.31-alpine@sha256:54f2a904c251d5a34adf545a72d32515a15e08418dae0266e23be2e18c66fefa AS serve
+FROM nginxinc/nginx-unprivileged:1.31-alpine@sha256:a8d5564c3354241473c1e152d5dd3281ab4224edb61b23c291e0bfd9854687a1 AS serve
```
katriendg, line 26: Non-blocking (docs): this PR moves the frontend to `nginx-unprivileged` on port **8080**, but the deployment contract in `data-management/specifications/viewer-deployment.specification.md` still documents the old state — the Architecture table lists the frontend port as `80`, and the Container Images table lists the base as `nginx:alpine`. Worth a follow-up edit there (port → `8080`, base → `nginx-unprivileged`) so the spec stays truthful. Not a blocker for approval.

katriendg (approved):
Approving — clean, well-scoped fix. Both compose build breakages are correctly repaired (backend uv via digest-pinned `COPY --from` + `uv sync --frozen`; frontend `.npmrc` now copied so `npm ci` resolves), the non-root hardening is sound, and the port 80 → 8080 change is propagated consistently across the frontend Dockerfile, nginx template, compose mapping, and the Container App ingress `target_port`. The `smoke-docker.sh` build gate plus the path-gated/always-on CI wiring closes the silent-regression gap.

One non-blocking note left inline: the deployment specification doc still references the old frontend port (80) and base image (`nginx:alpine`) and should be updated in a follow-up.

bindsi (approved):
Approving — a clean, well-scoped fix-plus-hardening PR. 🚀

Both compose-build breakages are correctly repaired: backend gets uv via the digest-pinned `COPY --from=ghcr.io/astral-sh/uv` + `uv sync --frozen` (keeping OSSF Scorecard Pinned-Dependencies satisfied, and self-guarding since a Dependabot digest bump now re-triggers the build smoke), and the frontend copies `.npmrc` before `npm ci` so the eslint-10/jsx-a11y `ERESOLVE` resolves. The non-root hardening (`appuser`, `nginx-unprivileged`) is sound, and the 80 → 8080 change is propagated consistently across the frontend Dockerfile, nginx template, compose mapping, and the Container App `target_port`. The `smoke-docker.sh` gate plus the path-gated/always-on CI wiring closes the silent-regression hole.

katriendg already covered the one stale-doc item (deployment spec still referencing port 80 / `nginx:alpine`), so I won't repeat it. Two additional points katriendg didn't raise, both non-blocking:

1. **The smoke is build-only** — `docker compose config` + `build`, no `up`/healthcheck. That's an explicit, reasonable scope choice (it catches exactly the base-image/dependency regression class behind #1065). A future enhancement could add a `compose up` + healthcheck-wait to also catch *runtime* regressions — e.g. the `nginx-unprivileged` entrypoint's `envsubst` writing `/etc/nginx/conf.d/default.conf` as the non-root `nginx` user, which a build-only gate won't exercise.

2. **Build smoke runs with empty `VITE_AZURE_*` args** (compose `${…:-}` defaults), so it won't surface issues that only appear with real auth config baked in. Fine for a build gate — just noting the coverage boundary.

Neither blocks merge. Nice work shutting the regression gap.

data-management/viewer/frontend/Dockerfile
algattik: Done in acf202bc — updated `data-management/specifications/viewer-deployment.specification.md`: Architecture table frontend port `80 → 8080` and Container Images base `nginx:alpine → nginx-unprivileged`, so the spec matches the Dockerfile.

PR #1068, opened 2026-06-29
feat(build): auto-merge low-risk patch dependency updates

### LINKED ISSUE #1067: Auto-merge low-risk patch dependency updates
## Problem

Dependabot opens many PRs weekly. The trivial tail — patch bumps to build tooling, infra providers, and base images — is low-risk yet still waits on a manual click after review, while genuinely risky updates get the same handling. The toil pushes maintainers to rubber-stamp, which erodes review quality.

## Outcome

Auto-merge the safe, low-risk patch tail once required checks pass, while everything riskier stays in manual review:

- Patch-only; major/minor stay manual.
- Security updates never auto-merge.
- Exclude packages whose breakage CI can't see (GPU/ABI Python runtimes).
- No PR code runs to make the decision; merge stays gated by the existing required checks and approval.
- A clear revert path when a bad patch lands.

Reduce reviewer toil on the trivial majority without lowering the bar.

algattik:
Closes #1067.

Dependabot's patch tail is low-risk but still waits on a manual click after review; this auto-merges the safe slice while leaving anything riskier to maintainers.

## What it does

[.github/workflows/dependabot-auto-merge.yml](.github/workflows/dependabot-auto-merge.yml) runs on dependabot PRs, reads `dependabot/fetch-metadata` (no PR code executed), and arms `gh pr merge --auto --squash` only when **all** hold:

- update-type is patch (`version-update:semver-patch`)
- not a security advisory (`ghsa-id` empty)
- ecosystem in `github_actions`, `terraform`, `npm_and_yarn`, `docker`

Major/minor, uv (GPU/ABI risk CI can't see), and security stay manual. The allowlist is fail-closed: a new ecosystem isn't auto-merged until explicitly added.

## Safety

- Merge is never forced — auto-merge waits for the `pr-validation-summary` required check and the existing approval, then squash-merges. It only removes the post-approval click.
- Pre-enable: if "Allow auto-merge" is off, the step traps the error and exits 0 — a green no-op, nothing merges.
- Magic values are commented as fetch-metadata outputs (e.g. `npm` → `npm_and_yarn`) to prevent future "corrections".

## Docs

`docs/contributing/component-updates.md` gains a Patch Auto-merge section: scope, the one admin setting required ("Allow auto-merge"), and a revert playbook.

Config-only; no application code changes.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -25,6 +25,12 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
+      - name: Checkout trusted base scripts
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.base.sha }}
+
       - name: Enable auto-merge for eligible patch bumps
         # All values below are dependabot/fetch-metadata outputs, not dependabot.yml strings:
         #  - update-type 'version-update:semver-patch' = patch bumps only
@@ -43,12 +49,4 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_URL: ${{ github.event.pull_request.html_url }}
-        run: |
-          if out=$(gh pr merge --auto --squash "${PR_URL}" 2>&1); then
-            echo "Auto-merge armed: ${PR_URL}"
-          elif grep -qi "auto.merge is not allowed" <<<"${out}"; then
-            echo "::notice::'Allow auto-merge' is off; no-op until an admin enables it."
-          else
-            echo "${out}" >&2
-            exit 1
-          fi
+        run: scripts/dependabot-auto-merge.sh --pr-url "$PR_URL"
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,55 @@
+name: Dependabot Auto-merge
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  auto-merge:
+    name: Auto-merge patch bumps
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Fetch Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@25dd0e34f4fe68f24cc83900b1fe3fe149efef98 # v3.1.0
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Enable auto-merge for eligible patch bumps
+        # All values below are dependabot/fetch-metadata outputs, not dependabot.yml strings:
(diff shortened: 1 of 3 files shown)


bindsi (approved):
I love this --> best PR ever...thanks for this PR which will ease our lives a lot

katriendg (changes requested):
Can we hold off this one please? There are a few branch protection rules we need to add before we merge (on us @algattik - let us check and take care of merge).

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,55 @@
...
+  contents: read
+
+jobs:
+  auto-merge:
+    name: Auto-merge patch bumps
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    permissions:
+      contents: write
+      pull-requests: write
```
katriendg, line 19: Required change: drop `contents: write` here. Enabling auto-merge needs only
`pull-requests: write` — the workflow never writes repository contents (no checkout,
no commit, no push). Given this job carries an elevated token on Dependabot PRs,
trimming to least privilege removes write access we don't use and shrinks the blast
radius if the step is ever abused.


```suggestion
    permissions:
      pull-requests: write
```

katriendg (changes requested):
Really solid workflow — the hard parts are already right: it runs on `pull_request`
(not `pull_request_target`), never checks out or executes PR code, never force-merges
(it only arms `--auto` and still waits on `pr-validation-summary` + approval), the
allowlist is fail-closed, and security advisories (`ghsa-id != ''`) stay manual.

Holding approval on one thing, and it's not in this diff: auto-merge is only as safe
as `main`'s branch protection. Supply-chain attacks via the dependency channel have
been rising sharply, and patch-pinned transitive updates are a favored vector — so the guardrails around an automatic merge path need to be airtight before we turn it on.

Two protections must be confirmed/enabled on `main` first:

1. Dismiss stale approvals on new commits (`dismiss_stale_reviews: true`). Without it,
   a commit pushed *after* approval — a Dependabot rebase/version bump, or any
   write-access push — keeps the old approval valid, and auto-merge will ship code no
   human reviewed at its current HEAD. This is the core risk and the reason to wait.
2. Require approval, and specifically the move to **more than one approver on all PRs**.
   We wanted multi-approver review regardless; this auto-merge path makes it a prerequisite rather than a nice-to-have. Tracking that as separate branch-protection work.

Recommend we land those branch-protection changes first, then approve this. Could also
add a one-line prerequisite to the new Patch Auto-merge doc section noting it depends on
`main` having stale-approval dismissal and the required-approver count in place.

Before clicking merge, please confirm `main`'s rules with:
  gh api repos/microsoft/physical-ai-toolchain/branches/main/protection \
    --jq '{required_approvals: .required_pull_request_reviews.required_approving_review_count, dismiss_stale_reviews: .required_pull_request_reviews.dismiss_stale_reviews, strict_up_to_date: .required_status_checks.strict}'
Expect dismiss_stale_reviews: true and the agreed required_approvals count.

@WilliamBerryiii - I'd like your input on this one, and let me know if you'd like us to manage the new rules, or edit them yourself.

.github/workflows/dependabot-auto-merge.yml
algattik: Done in 223ba73f — dropped `contents: write`, leaving `pull-requests: write` only. The job just arms auto-merge via `gh pr merge --auto` (no checkout/commit/push), so write access to contents was unused. actionlint passes.

WilliamBerryiii:
I'll dig into this, but I believe MSFT github org global policy prevents anything other then a human gate.

Issue #1085, opened 2026-07-01
Container base images are not scanned for CVEs


algattik:
## Problem

Our published containers build on public base images (Python, Node, nginx, an Azure ML image). The OS and library layers of those base images are never scanned for vulnerabilities. Existing checks (dependency review, OSV, CodeQL) cover our own language dependencies but not the base-image layers.

A fixable CRITICAL/HIGH CVE can therefore sit undetected in a shipped image. There is no fast path to catch it: base-image OS CVEs do not trigger security dependency updates, and the routine image-refresh cadence is intentionally throttled to weekly.

## Desired outcome

- Base images are scanned for known vulnerabilities; fixable CRITICAL/HIGH findings surface in the Security tab — on pull requests that touch a Dockerfile, and on a weekly re-scan of already-shipped images.
- The intentional weekly cadence for routine, non-security image updates is preserved; security scanning is a separate channel that reacts only to actual vulnerabilities.

## Scope

Scanning is the first of three channels. Enforcement — flipping the gate from report-only to blocking on fixable CVEs — and automated remediation of shipped images are tracked separately in #1118 and #1117.

PR #1088, opened 2026-07-01
security(scripts): harden supply-chain pinning across deps, downloads, and images

### LINKED ISSUE #1080: [Task]: pin unpinned CI, agent, and runtime dependencies; extend the pinning scanner
## Summary

Close four related supply-chain blind spots so every CI, agent, and runtime dependency resolves to
an immutable, verifiable reference, and the existing hard-fail CI gate catches any new unpinned
dependency: floating installs in the Copilot agent / CI environment, mutable runtime GPU image tags,
an unverified binary download executed inside the GR00T training container, and matching gaps in the
dependency-pinning scanner.

## Category

- [x] Improvement suggestion

## Goals

**Pin the agent / CI installs.** The cloud agent environment installs `gh-aw` and `powershell-yaml`
at floating HEAD / latest. Pin both to released versions so every run resolves a fixed, auditable
version, and update the agent guidance to match.

**Digest-pin runtime GPU images.** Training and evaluation run from images referenced by mutable tag
(nvcr.io Isaac Lab, Docker Hub PyTorch); a tag can be overwritten in place. Pin each to an immutable
`@sha256:` digest, keep the human-readable tag for legibility, and document the digest-bump
procedure.

**Verify the ORAS binary download.** The GR00T training container fetches and executes an ORAS
release tarball with no integrity check — code execution if the asset or network path is
compromised. Verify the tarball against a recorded SHA-256 and fail closed on mismatch. Pin the
IntelliSense-only Isaac Lab developer clone to a commit as a low-priority rider.

**Extend the dependency-pinning scanner.** The scanner does not inspect composite actions, downloads
inside workflow `run:` blocks, `gh extension install` without `--pin`, or `Install-Module` without
`-RequiredVersion` — so the pins above could regress unnoticed. Extend coverage so the existing
hard-fail gate flags them.

## Outcome / acceptance

- No CI, agent, or runtime dependency resolves to a floating tag, HEAD, or latest: `gh-aw` and
  `powershell-yaml` are version-pinned; the Isaac Lab and PyTorch images carry an `@sha256:` digest;
  the ORAS download is SHA-256-verified and aborts on mismatch.
- The dependency-pinning scanner flags composite-action, workflow-`run:` download,
  `gh extension install`, and `Install-Module` pinning violations, and the hard-fail compliance gate
  stays green on `main`.
- Image digest-bump and ORAS checksum-bump procedures are documented for maintainers.

## Notes

The scanner extension guards the same install pins it sits alongside; the image-digest and ORAS
hardening share the GR00T training workflow.

### LINKED ISSUE #1081: Harden download-pinning checks repo-wide and require checkpoint integrity verification
## Summary

Strengthens supply-chain integrity for shell-script downloads and model-checkpoint loading across the repository.

## Background

The `shell-downloads` dependency-pinning check only inspected shell scripts directly under the root `scripts/` directory. Provisioning, training, and evaluation scripts elsewhere in the tree (`training/`, `evaluation/`, `infrastructure/setup/`, …) were never checked, so an unverified `curl`/`wget` download could sit outside the scanner's view.

## Changes

**Scanner coverage**

- `shell-downloads` now scans every `.sh` in the repository (vendored/VCS trees — `.git`, `.venv`, `node_modules`, `external` — excluded).
- Recognizes the repository's portable `verify_sha256` helper as valid checksum verification (removes a false positive).
- Supports a `# pinning-ignore: <reason>` annotation — on the download line, or the comment line directly above it — for downloads that have no stable digest to pin.

**Triage of surfaced findings**

- Three trusted apt-repository downloads in the Isaac-Sim dev-VM provisioning script (dynamic, distro-dependent URLs trusted via GPG-signed repositories) are annotated with justified opt-outs.

**Behavior change — checkpoint integrity (reviewer call-out)**

- The SIL inference entry point now **requires** an expected SHA-256 when a checkpoint is fetched over plain `http(s)` — the only unauthenticated download path. Authenticated sources (MLflow `runs:/` / `models:/`, Azure Blob) remain optional. This fails closed rather than loading an unverified checkpoint. No existing in-repo caller uses the plain-HTTP-without-SHA path.

**Documentation & tests**

- Added a download-pinning convention to the shell-script authoring instructions.
- Added unit tests covering repo-wide discovery (nested / non-`scripts` locations), the `verify_sha256` helper, and both forms of the ignore annotation.

## Validation

- PowerShell unit tests, PSScriptAnalyzer, markdown lint, and shellcheck on edited scripts all pass.
- Full dependency-pinning scan reports no violations.

### LINKED ISSUE #1082: Harden container-image supply chain: digest-pin images and enforce pinning in CI
## Problem

Runtime container images across the workflow-orchestration and job-submission surface are referenced by mutable tags, and some float on `:latest`. Tags are mutable, so image pulls are neither reproducible nor tamper-evident, and a moved tag can change what runs without any change in the repository.

Nothing enforces digest pinning on this surface. The supply-chain scanner covers GitHub Actions, package manifests, and inline `pip`/`uv` installs, but not workflow-YAML `image:` references. OpenSSF Scorecard only inspects Dockerfiles, and Dependabot does not parse `image:` fields in workflow YAML — so the workflow image surface has no guard at all.

## Goal

Make every runtime container image reference reproducible and tamper-evident, and prevent regressions automatically:

- Concrete OCI image references on the workflow and submission surface are pinned by immutable `@sha256` digest (the readable tag may be retained alongside the digest).
- An automated check flags any unpinned workflow-YAML `image:` reference and gates it in CI, so a new unpinned image cannot merge.
- A documented procedure refreshes the digests when an image tag is intentionally bumped.

## Coverage

The reproducibility and enforcement goals apply to the container images referenced from:

- OSMO and AzureML workflow templates.
- Kubernetes manifests and Helm values.
- The shared job-submission defaults that downstream scripts inherit.

## Acceptance criteria

- [ ] No tag-only OCI image reference remains on the covered surface.
- [ ] The scanner flags an unpinned workflow `image:` reference and fails CI at the existing compliance threshold.
- [ ] Submission-time templated and shell-variable image references are not false-flagged, and an intentional exception can opt out explicitly.
- [ ] The digest-refresh procedure is documented.

algattik:
# Pull Request

## Description

Hardens the repository's software supply chain across three fronts: dependency pinning enforcement, download integrity, and container-image digest pinning. The dependency-pinning scanner is extended to catch several previously unenforced classes of unpinned dependency, unpinned references across CI workflows and runtime scripts are pinned, and all first-party container images are pinned by `@sha256` digest with tooling to keep those digests fresh.

Closes #1080
Closes #1081
Closes #1082

### What changed, by theme

**Dependency-pinning scanner (`scripts/security/Test-DependencyPinning.ps1`)**

- New validators for `gh extension install`, PowerShell `Install-Module`, and Docker image references in workflows, in addition to the existing shell-download and inline-pip checks.
- Fixes to existing detection logic:
  - The `pinning-ignore` exemption no longer matches the token anywhere on a line; it must be a real inline exemption.
  - Multi-line download/pip commands are now folded via `Join-LineContinuation` before scanning, so continuations are evaluated as one command (bash-context patterns use backslash-only continuation; pwsh contexts also honour a trailing backtick).
  - Coverage gap closed: local `**/*.ps1` and `**/*.sh` scripts are now scanned, not only workflow YAML.
  - Scan entries are de-duplicated by `(path, type)` rather than `path` alone, so a file matched by multiple rule types is run through every applicable validator instead of just one.
  - Stateful inline-exception tracking for `gh extension` and module installs.

**Container-image digest pinning**

- First-party image references are pinned by `@sha256:...` digest. Image defaults are centralised in `scripts/lib/common.sh` so a single source of truth drives training/eval/OSMO workflows.
- New `scripts/update-image-digests.sh` refreshes digests repo-wide (skips `tests/Fixtures` and `*.Tests.ps1`, uses `cp` to preserve target-file permissions, and applies a hyphen-aware `sed` ref boundary).
- `scripts/security/tool-checksums.json` gains the tool digests, and the update procedure for ORAS/Actionlint/Gitleaks is documented rather than left as ad-hoc shell edits.

**AzureML submission scripts**

- Shared AzureML environment-registration helper added; environment versions are now derived from the (digest-pinned) image rather than hand-maintained.
- Explicit `--image` / version overrides are preserved and existing images are verified; the environment version is recomputed after an `--image` override.
- Empty subscription arguments are no longer passed to the AzureML CLI.

**CI workflows**

- Pinned previously unpinned references in `dependency-pinning-scan.yml`, `frontmatter-validation.yml`, `msdate-freshness-check.yml`, and `pester-tests.yml`.

**Documentation**

- `docs/security/README.md`, `docs/contributing/component-updates.md`, and `docs/reference/scripts.md` updated to reflect digest-aware versions and the checksum-registry update procedure.

**Tests**

- New Pester fixtures and cases covering the gh-extension, install-module, and docker-image validators plus the inline-pip compliant path (`scripts/tests/security/Test-DependencyPinning.Tests.ps1`, `.../Fixtures/**`). Test-only additions; no production behaviour depends on them.

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation

## Testing Performed

Dependency-pinning Pester suite run locally: `Invoke-Pester -Path 'scripts/tests/security/Test-DependencyPinning.Tests.ps1' -CI` → 173 passed, 0 failed.

## Documentation Impact

## Bug Fix Checklist

## Checklist


scripts/security/Test-DependencyPinning.ps1
```diff
@@ -575,6 +668,237 @@
...
+}
+
+function Get-PowerShellModuleViolations {
+    <#
+    .SYNOPSIS
+        Detects `Install-Module` invocations that do not pin an exact version with -RequiredVersion.
+    .DESCRIPTION
+        Install-Module without -RequiredVersion resolves the latest gallery version at run time.
+        Only command-position invocations are flagged: a `Mock Install-Module`, a comment, or an
+        Install-Module mention inside a string (e.g. a Write-Warning hint) are ignored. A trailing
```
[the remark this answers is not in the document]
algattik: Fixed in 3a0a9fe. This alert is a self-scan false positive: the flagged line is the `Install-Module` token inside this function's `<# #>` comment-based help, not a live invocation. `Get-PowerShellModuleViolations` now tracks and skips block-comment and here-string regions, so documentation and string content are no longer treated as commands. New unit tests cover both cases; the full Pester suite passes (175/175).

scripts/tests/security/Test-DependencyPinning.Tests.ps1
```diff
@@ -1382,3 +1765,266 @@
...
+name: bad
+on: push
+jobs:
+  x:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: gh extension install github/gh-aw
+      - shell: pwsh
+        run: Install-Module -Name powershell-yaml -Force -Scope CurrentUser
```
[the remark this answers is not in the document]
algattik: Fixed in 3a0a9fe. This alert is a self-scan false positive: the flagged line is a deliberately-unpinned `Install-Module` inside a single-quoted here-string (`@'...'@`) that the test writes out to a temp workflow file — the string is fixture data, not a live invocation, and the test's real assertion comes from scanning that generated workflow. `Get-PowerShellModuleViolations` now skips here-string bodies, so the fixture source is no longer flagged. Added a unit test for the here-string exclusion; Pester passes (175/175).

scripts/security/Test-DependencyPinning.ps1
katriendg, line 1: **Digest enforcement coverage gap (manifests + Helm values)** — Medium

The `docker` validator's `FilePatterns` are `**/workflows/**/*.y*ml` only, but #1082's covered surface also names "Kubernetes manifests and Helm values." Two files this PR digest-pins by hand sit outside that glob and so aren't regression-guarded:

- `infrastructure/setup/manifests/gpu-grid-driver-installer.yaml` (`ubuntu:22.04`, `registry.k8s.io/pause:3.10`)
- `infrastructure/setup/values/osmo-control-plane.yaml` (`backend_images.init` / `.client`)

The Helm values also use `init:` / `client:` keys, which the `image:`-only line regex can't see even if the file were in scope. `update-image-digests.sh` already *discovers* these (it git-greps by digest shape), so refresh and enforcement are asymmetric today: a future edit that drops a digest here would refresh fine but pass CI silently.

Two ways to close it, whichever you prefer:

1. Widen the `docker` `FilePatterns` to include the manifest/values paths and broaden the key match beyond `image:` (e.g. an allowlist covering `init:`/`client:`), or
2. If that's more than you want in this PR, leave a short comment in `Get-DockerImageViolations` documenting the intentional workflow-only scope, and track the manifest/Helm coverage as a follow-up.

Non-blocking, but I'd lean toward landing option 1 here since it closes a stated acceptance-criteria gap in #1082.

scripts/update-image-digests.sh
katriendg, line 1: **"One step further": image-digest freshness/staleness loop** — Low (enhancement)

This PR already applies the "pin + validate against a maintained list" pattern to binaries: `tool-checksums.json` is the pin list, `Test-BinaryFreshness.ps1` re-hashes against upstream, and `Test-SHAStaleness.ps1` checks the releases API — and you've correctly extended that loop to ORAS here.

The one surface that gets pins *without* a validation loop is the container images: digests live across `common.sh`, the workflow YAMLs, a manifest, and Helm values, refreshed only by this manual `update-image-digests.sh`. There's no scheduled check that re-resolves each readable tag to its current registry digest and flags drift — the closing-the-loop step microsoft/edge-ai runs via `Update-DockerSHAPinning.ps1` alongside its staleness checks.

Worthwhile follow-up: extend `Test-BinaryFreshness.ps1` to re-resolve each discovered `tag@sha256` (the registry-metadata logic already exists in this script) and warn on drift, or surface `update-image-digests.sh --dry-run` as a scheduled SARIF job. That gives images the same tamper-evidence **plus** freshness the binaries already have.

Given the size (new check + tests + scheduled workflow), I'd suggest tracking this as a backlog issue rather than expanding this PR — happy to file it. Flagging here so the thread is visible; resolve once a decision/issue is recorded.

katriendg (approved):
Thanks for this, @algattik, great one — detailed and high quality as always. Pinning deps, downloads, and images in one coherent pass, wiring ORAS into the existing `tool-checksums.json` freshness/staleness loop, folding line continuations so the scanner reads whole commands, and the fail-closed touches (checkpoint SHA gate, `register_azureml_environment` verifying the existing image, the clone-to-temp-then-`mv` in `setup-dev.sh`) all show real care. The Pester coverage on the new validators is excellent.

This fully closes #1080 and #1081, and meets #1082 for the workflow surface (verified no tag-only workflow `image:` reference remains).

Approving. I've left two inline comments — both additive rather than corrections. Please resolve each thread before merge:

- **Test-DependencyPinning.ps1** — a digest-enforcement coverage gap on the K8s manifest and Helm values that #1082 also names (Medium; I'd lean toward closing it here).
- **update-image-digests.sh** — the "one step further": images get pins but no freshness/staleness loop like binaries and Action SHAs already have (Low/enhancement; likely a backlog issue).

Neither changes my approval; flagging them so they're tracked and consciously resolved.

scripts/security/Test-DependencyPinning.ps1
algattik: Went with option 1 (86993d1). The `docker` validator now also scans `infrastructure/setup/manifests/*.y*ml` and `infrastructure/setup/values/*.y*ml`, and the key match is broadened from `image:` to `image|init|client`, so the OSMO `backend_images` `init:`/`client:` refs are guarded alongside the manifest `image:` lines. An `init:`/`client:` value is treated as an image only when it carries a registry/namespace path (`/`), so plain scalars like `init: true` / `client: guest` are left untouched.

Verified: a full-repo `docker`-type scan stays clean (0 violations, no new false positives), and stripping the digests from the manifest `image:` and the `init:`/`client:` values now flags all three. Added Pester coverage for the new keys and for manifest/Helm-values discovery; PSScriptAnalyzer clean.

scripts/update-image-digests.sh
algattik: Filed as #1093 rather than expanding this PR, as you suggested. It scopes a scheduled check that re-resolves each pinned image tag to its current registry digest and flags drift, across the workflow YAML / Kubernetes manifest / Helm values / shell-library surfaces, and records the boundary with #1085 (Dockerfile-`FROM` CVE remediation, which already carries tag→digest resolution) so the loop shares that resolution rather than duplicating it. Happy for you to resolve this thread once you've had a look.

scripts/security/Test-DependencyPinning.ps1
katriendg: this is great to have it within the same PR

PR #1089, opened 2026-07-01
chore(build): remediate manifest drift and harden the pinning gate

### LINKED ISSUE #1087: chore(build): pin floating dependency specifiers to eliminate manifest drift
## Problem

Three dependency manifests use floating specifiers. Lock files neutralize the risk
today, but the manifests are the source of truth and permit silent drift on the next
regeneration.

- `pyproject.toml` — `atheris>=3.1.0` (floating lower bound)
- `docs/docusaurus/package.json` — `follow-redirects` override `">=1.16.0 <2.0.0"`
  (range, unlike its sibling exact-pinned overrides)
- `data-pipeline/pyproject.toml` — declares no runtime/test dependencies despite
  `data-pipeline/capture/` importing `pydantic` and the test suite using `pytest`,
  `pytest-cov`, and `hypothesis`

## Proposed change

- Pin `atheris==3.1.0`.
- Pin the `follow-redirects` override to exact `1.16.0`, matching the sibling overrides.
- Declare `pydantic` as a runtime dependency and `pytest`/`pytest-cov`/`hypothesis` as
  dev dependencies in `data-pipeline/pyproject.toml`; regenerate its `uv.lock`.

## Acceptance criteria

- `npm run lint:uvlock` stays green after `uv lock`.
- No floating specifier remains in the named manifests.
- All `data-pipeline` imports are covered by declared deps.

Severity: Low · Effort: S

algattik:
Closes #1087.

## Why

Several dependency manifests carried floating or undeclared specifiers, causing drift between what is declared and what is actually resolved and used. This pins them to exact versions so builds are reproducible and consistent with the repo's dependency-pinning security gate (`Test-DependencyPinning.ps1`).

The gate itself could not have caught two of the drifts this PR fixes: a specifier on a single-line `pyproject.toml` array and a range inside a `package.json` `overrides` block both slipped past the parser entirely. Hardening the parser closes those gaps so the same drift cannot silently regress after this lands.

## Changes

**Pin floating version specifiers to exact pins**

- `pyproject.toml` (root): `fuzz` extra `atheris>=3.1.0` → `atheris==3.1.0`.
- `docs/docusaurus/package.json`: `follow-redirects` override `>=1.16.0 <2.0.0` → `1.16.0`, making it consistent with the other exact-pinned overrides in that block.

**Declare previously-undeclared dependencies in `data-pipeline`**

- Adds runtime `dependencies = ["pydantic==2.13.4"]` — `pydantic` was already imported by `capture/models/config_models.py` but never declared in the manifest.
- Adds `[dependency-groups] dev` (`hypothesis==6.155.7`, `pytest==9.1.1`, `pytest-cov==7.1.0`), matching the sibling `evaluation/` convention and the existing `--cov` test configuration.

**Regenerate locks**

- `uv.lock` (root) and `data-pipeline/uv.lock` regenerated via `uv lock`.

**Close two parser blind spots in the dependency-pinning gate**

- `Test-DependencyPinning.ps1` — `Get-PipDependencyViolations`: parse specifiers that share the array-opener line so single-line arrays such as `fuzz = ["atheris>=3.1.0"]` are checked. The opener branch previously matched `name = [` and skipped straight to the next line, so any specifier on that same line (the exact shape of the root `atheris` pin above) was never inspected. Extraction is factored into a shared `Get-PyprojectSpecViolation` helper reused by the multi-line path.
- `Test-DependencyPinning.ps1` — `Get-NpmDependencyViolations`: also scan the `overrides` and `resolutions` sections, walking them recursively via `Get-NpmOverrideViolation` so nested override objects and the `.` self-pin are covered. Only `dependencies`, `devDependencies`, `peerDependencies`, and `optionalDependencies` were scanned before, so the `follow-redirects` override pinned above sat in an unscanned section.
- Fixtures and Pester coverage for both gaps: `scripts/tests/Fixtures/Pip/single-line-pyproject.toml`, `scripts/tests/Fixtures/Npm/overrides-package.json`, plus new cases under `Get-PipDependencyViolations` and `Get-NpmDependencyViolations`.

## Behaviour impact

The manifest and lock edits have no runtime impact — every pin equals the already-resolved version, so there are no transitive dependency-tree changes. The gate change is detection-only: it is strictly stricter and adds no new violations on the current tree (a full scan of every `pyproject.toml` and `package.json` reports zero), so the pinning CI gate stays green while now covering the two previously-blind shapes.

## Validation

- `uv lock --check` passes for both root (170 packages) and `data-pipeline` (16 packages) → the `lint:uvlock` CI gate stays green.
- `data-pipeline` test suite: **71 passed**, 100% coverage on `config_models.py`.
- `Test-DependencyPinning.Tests.ps1`: **108 unit tests pass** (7 new, covering single-line pyproject arrays and npm overrides/resolutions).
- Full-repo scan with the hardened parser reports **zero violations**, confirming no false positives and no new CI failures.
- PSScriptAnalyzer on the changed script adds no new findings.

### DIFF
--- a/data-pipeline/pyproject.toml
+++ b/data-pipeline/pyproject.toml
@@ -3,6 +3,9 @@ name = "physical-ai-data-pipeline"
 version = "0.1.0"
 description = "Data capture pipeline runtime"
 requires-python = ">=3.12"
+dependencies = [
+    "pydantic==2.13.4",
+]
 
 [tool.uv]
 package = false
@@ -42,3 +45,10 @@ exclude_lines = [
 
 [tool.coverage.xml]
 output = "logs/coverage.xml"
+
+[dependency-groups]
+dev = [
+    "hypothesis==6.155.7",
+    "pytest==9.1.1",
+    "pytest-cov==7.1.0",
+]
--- a/data-pipeline/uv.lock
+++ b/data-pipeline/uv.lock
@@ -3,6 +3,311 @@ revision = 3
 requires-python = ">=3.12"
 
 [[package]]
+name = "annotated-types"
+version = "0.7.0"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/ee/67/531ea369ba64dcff5ec9c3402f9f51bf748cec26dde048a2f973a4eea7f5/annotated_types-0.7.0.tar.gz", hash = "sha256:aff07c09a53a08bc8cfccb9c85b05f1aa9a2a6f23728d790723543408344ce89", size = 16081, upload-time = "2024-05-20T21:33:25.928Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/78/b6/6307fbef88d9b5ee7421e68d78a9f162e0da4900bc5f5793f6d3d0e34fb8/annotated_types-0.7.0-py3-none-any.whl", hash = "sha256:1f02e8b43a8fbbc3f3e0d4f0f4bfc8131bcb4eebe8849b8e5c773f3a1c582a53", size = 13643, upload-time = "2024-05-20T21:33:24.1Z" },
+]
+
+[[package]]
+name = "colorama"
+version = "0.4.6"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/d8/53/6f443c9a4a8358a93a6792e2acffb9d9d5cb0a5cfd8802644b7b1c9a02e4/colorama-0.4.6.tar.gz", hash = "sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44", size = 27697, upload-time = "2022-10-25T02:36:22.414Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/d1/d6/3965ed04c63042e047cb6a3e6ed1a63a35087b6a609aa3a15ed8ac56c221/colorama-0.4.6-py2.py3-none-any.whl", hash = "sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6", size = 25335, upload-time = "2022-10-25T02:36:20.889Z" },
+]
+
+[[package]]
+name = "coverage"
+version = "7.14.3"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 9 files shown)


katriendg (approved):
Nice remediation one, thanks.

While reviewing this manifest-drift fix, I noticed the pins here were the right call, but the pinning gate (`Test-DependencyPinning.ps1`) wouldn't have caught these cases on its own — there are two **manifest-parser blind spots** that let floating specifiers slip past entirely:

1. **Single-line pyproject arrays** — in `Get-PipDependencyViolations`, the array-start branch matches `^\s*(\w[\w-]*)\s*=\s*\[` and immediately `continue`s, so a same-line specifier is never parsed. Example that passes today despite being floating:

   ```toml
   fuzz = ["atheris>=3.1.0"]
   ```

2. **package.json `overrides` / `resolutions`** — `Get-NpmDependencyViolations` only scans `dependencies`, `devDependencies`, `peerDependencies`, `optionalDependencies`. A range in an override like `"follow-redirects": ">=1.16.0 <2.0.0"` isn't checked.

So even after this PR lands, a future regression of the exact drift it fixes would go undetected. Closing the gaps would mean: (a) parse same-line quoted specifiers in the pyproject array branch, and (b) add `overrides`/`resolutions` to the npm section list — each with a fixture.

This is out of scope for this PR. Could you either fold in the two small fixes here, or open a backlog item so we don't lose track of it? Happy to take the follow-up if that's easier. Just tag me so I get a notification - missing a few of these lately.

algattik:
Both blind spots are now closed in this PR (e50029b):

- **Single-line pyproject arrays** — `Get-PipDependencyViolations` now parses specifiers that share the array-opener line, so `fuzz = ["atheris>=3.1.0"]` is checked. Extraction moved into a shared `Get-PyprojectSpecViolation` helper reused by the multi-line path.
- **package.json `overrides` / `resolutions`** — `Get-NpmDependencyViolations` now scans both sections, walking them recursively (`Get-NpmOverrideViolation`) so nested override objects and the `.` self-pin are covered.

Each ships with a fixture and Pester cases (`Fixtures/Pip/single-line-pyproject.toml`, `Fixtures/Npm/overrides-package.json`). 108 unit tests pass and a full-repo scan reports zero violations, so the gate stays green while now covering the two shapes it previously skipped — including the exact `atheris` and `follow-redirects` drifts this PR pins.

PR #1094, opened 2026-07-02
security(build): scan container base images for fixable CRITICAL/HIGH CVEs

### LINKED ISSUE #1085: Container base images are not scanned for CVEs

algattik:
## Summary

Published container images build on public base images (Azure ML, Python, Node, nginx). Existing checks (dependency review, OSV, CodeQL) cover our own language dependencies but never the OS and library layers of those base images, so a fixable CRITICAL/HIGH CVE can ship undetected. Dependabot bumps base-image digests but does not scan behind them, and its routine refresh is intentionally throttled.

This adds base-image CVE **scanning** across two channels — a PR gate and a weekly re-scan — surfacing fixable CRITICAL/HIGH findings to the Security tab.

Closes #1085 — #1085 is scoped to base-image scanning, which this delivers. Two shipped base images currently carry fixable CVEs, so the PR gate runs **report-only** (findings still publish to the Security tab); enforcing the gate (flipping it to blocking) and automated remediation are separate channels, tracked in #1118 and #1117.

### 1. PR gate — `container-scan.yml` (new reusable workflow)

- Discovers every digest-pinned external base image referenced across all `Dockerfile`s and scans each with Trivy (`scanners: vuln`, `os,library` packages, `severity: CRITICAL,HIGH`, `--ignore-unfixed`).
- Wired into `pr-validation.yml`: runs when a PR touches a `Dockerfile`, the scan workflow, or a discovery/slug helper, and is part of the required `pr-validation-summary` gate. It currently runs **report-only** (`soft-fail: true`, mirroring the existing Checkov gate): fixable CRITICAL/HIGH findings are published but do not block the PR. The reusable workflow's own default is `soft-fail: false`; removing the override to flip the gate to blocking is tracked in #1118 (it requires triaging the current findings first).
- Results upload to the Security tab as SARIF (`security-events: write`), one category per image. Fork-PR SARIF upload is `continue-on-error` (expected token limitation on forks), but the gating step still runs.

### 2. Weekly re-scan — `weekly-validation.yml`

- Adds a `container-rescan` job that reuses `container-scan.yml` with `soft-fail: true`: it re-evaluates already-merged base images and surfaces newly-disclosed CVEs to the Security tab without failing the run (the PR gate only sees images at merge time).

## Supporting changes

- Two helpers live in `scripts/security/` (beside `generate-vex.sh`, which sources the same `scripts/lib/common.sh`), the repository's established home for security tooling:
  - `discover-base-images.sh`: extracts the unique, digest-pinned `FROM` base images across Dockerfiles; excludes stage aliases and `ARG`/`scratch` bases (no `@sha256`, nothing to scan).
  - `image-slug.sh`: produces a stable per-image slug (readable tag prefix + hash of the full ref), used to name the scan's per-image SARIF file and category.
- `docs/security/workflow-permissions.md`: enumerates the new `security-events: write` grants for the scan and re-scan jobs; refreshes `ms.date`.
- `.cspell` dictionary: adds technical terms introduced by the new workflows and docs.

## Testing / validation

This is CI/tooling only. The pure, deterministic helpers are covered by Pester. Validation performed:

- `scripts/tests/security/image-slug.Tests.ps1` and `discover-base-images.Tests.ps1` — 16 Pester tests covering slug shape, tag-collision disambiguation and hash provenance, plus the `FROM`-parser edge cases (digest / stage-alias / `ARG` / `scratch`, case-insensitivity, dedup, empty repo). They run the shipped bash helpers on the ubuntu Pester runner and skip cleanly where `bash`/`git` are absent. All green under Pester 5.7.1.
- `shellcheck -x` on the two helpers — clean.
- `actionlint` on all changed/new workflows — clean.
- `cspell` over the repo (716 files) — clean.
- Functional dry-run: `discover-base-images.sh` finds the four digest-pinned bases (Azure ML, `nginx`, `node`, `python`); `image-slug.sh` output is stable and collision-free across discovered images.
- Trivy's live CVE scan runs in the GitHub-hosted job (it needs the scanner plus multi-GB image pulls) and was not executed locally. On this PR it flags pre-existing fixable CVEs in the Azure ML and `node` base images (surfaced to the Security tab); because the gate is report-only, the run stays green.

## Scope and follow-ups

To keep this PR reviewable and unblock the scanning channel of #1085, the two channels that need maintainer / security-owner input are split into follow-ups. Together with this PR they cover the full outcome of #1085:

- **Enforcement** (flip the gate from report-only to blocking) — requires per-CVE security-team triage of the four bases: suppress non-exploitable findings via committed OpenVEX, bump digests where a clean fix exists, then remove the report-only override. Tracked in #1118.
- **Auto-remediation** (daily digest-bump pull request / tracking issue via a scoped App token) — needs an App-identity decision and runtime prerequisites. Tracked in #1117.


katriendg:
@nguyena2 added you as a reviewer as I believe it would be welcome, given some of the work you did on VEX.

scripts/security/image-slug.sh
```diff
@@ -0,0 +1,14 @@
...
+# (registry/repo:tag@sha256:...). A short hash of the full ref (digest included)
+# is appended so two images whose tags collapse to the same alphanumeric slug
+# still get distinct slugs. Shared by container-scan.yml (SARIF file names and
+# categories) and container-cve-remediation.sh (remediation branch names) so
+# both derive identical slugs from one source.
+set -o errexit -o nounset -o pipefail
+
+ref="${1:?usage: image-slug.sh <image-ref>}"
+
+hash="$(printf '%s' "$ref" | sha256sum | cut -c1-12)"
```
katriendg, line 11: **RI-2 / RI-5 — no automated coverage on very testable logic** (Testability, Medium)

The PR notes "no unit tests to add," but that is a consequence of choosing bash. Our security scripts set the opposite precedent: `scripts/security/Test-DependencyPinning.ps1` (our descendant of the hve-core script), `Test-SHAStaleness.ps1`, and `Test-BinaryFreshness.ps1` each have a `*.Tests.ps1` in `scripts/tests/security/`, run via `npm run test:ps`.

The highest-value logic here is pure and deterministic — ideal for Pester:
- `image-slug.sh` — slug disambiguation (`tag-prefix` + `sha256(ref)[:12]`, this line); test collision behavior.
- `discover-base-images.sh` — the awk `FROM` parser (digest / stage-alias / `ARG` / `scratch` edge cases).
- `count_from_report` / `cves_from_report` — jq parsing against fixture Trivy JSON.
- `--plan-output` plan generation — fixture-driven, no network.

If ported to PowerShell, these could also reuse `scripts/lib/Modules/CIHelpers.psm1` (`Write-CIAnnotation`, step summaries) and the SARIF exporters in `scripts/security/Modules/` instead of re-implementing logging on `common.sh`. Pragmatic middle ground if a full port is too heavy: keep the `gh`/`git` apply phase in bash but move the scan/plan/parse phase to Pester-tested PowerShell, or add fixture tests mirroring `shared/ci/tests/Fixtures`. The two tiny helpers are the cheapest, highest-value wins.

.github/workflows/container-cve-remediation.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Generate GitHub App token
+        if: steps.scan.outputs.remediation_needed == 'true'
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3.2.0
+        with:
+          app-id: ${{ vars.RELEASE_APP_ID }}
+          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-issues: write
```
katriendg: **RI-6 — App identity decision + runtime prerequisites before the daily job can remediate** (Correctness / Operations, High — blocks first real remediation)

The security posture is sound — read-only `GITHUB_TOKEN`, conditional App-token minting, explicit `contents/issues/pull-requests: write` downscope (this step), and script-level token isolation. Before the first non-quiet run, though, there is one identity-governance decision and two concrete prerequisites to settle:

**Open question — reuse `RELEASE_APP_ID`, or provision a dedicated security App?** This is the real prerequisite question. The workflow reuses the release-please App identity for a second, unrelated purpose (base-image CVE remediation). Worth a deliberate decision rather than defaulting to reuse:

- *Reusing `RELEASE_APP_ID`* costs nothing to provision and the per-mint downscope keeps the *token* tight — but it widens the *App's* blast radius: one private key now gates both release publishing and security PRs, a compromise affects both, the audit trail collapses to a single bot identity, and the release App must carry `Issues: write` it otherwise would not need. The `RELEASE_APP_*` name also stops describing its scope.
- *A dedicated App* (e.g. `SECURITY_REMEDIATION_APP_ID`) gives least-privilege by *identity* — installed with only `contents`/`issues`/`pull-requests` write, rotated independently, isolated from the release key, and a distinct bot author on security PRs for a clean audit trail. Cost: one more App + secret to provision and rotate.

Given this repo's stated OpenSSF-hardening posture, a dedicated App is the stronger fit, but this is a maintainer/security-owner call. **Which direction do we want here?** Whichever we pick, document the rationale in `docs/security/workflow-permissions.md` alongside the new App-token table.

Two concrete prerequisites regardless of the identity decision:

1. **`security` label must exist.** `open_bump_pr`/`open_or_refresh_issue` call `gh ... --label security --label docker`, and `gh` errors on a missing label (it will not create it). `docker` is confirmed (dependabot uses it); `security` is not referenced anywhere in repo config and is not a GitHub default label. Create it, or the first PR/issue creation fails.
2. **The chosen App must have `Issues: write` installed.** `contents`/`pull-requests` write are proven for `RELEASE_APP_ID` (release-please), but `create-github-app-token` can only narrow permissions, never add them. If the App lacks `Issues: write`, the downscope request is silently dropped and the tracking-issue path 403s. Confirm the installed permission set includes Issues:write (a fresh dedicated App would be provisioned with it explicitly).

Also worth a line in the PR description or `docs/security/workflow-permissions.md`: the "Allow GitHub Actions to create and approve pull requests" repo/org toggle is intentionally **not** required here because an App token is used, and App-token PRs (unlike `GITHUB_TOKEN` PRs) trigger the `container-scan` gate — that is the reason for choosing the App token.

@WilliamBerryiii and @agreaves-ms I would like your input on the decision here please?

.github/workflows/container-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+        run: |
+          set -euo pipefail
+          slug=$(bash .github/scripts/image-slug.sh "$IMAGE")
+          echo "slug=$slug" >> "$GITHUB_OUTPUT"
+        shell: bash
+
+      # Trivy writes the SARIF report before applying exit-code, so a gating
+      # failure here still leaves the file for the always() upload below.
+      - name: Trivy vulnerability scan
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # v0.36.0
```
katriendg, line 97: **RI-3 — no VEX suppression, though we already ship container-image VEX** (Security / Correctness, High — before flipping the gate to blocking)

We already have container-image VEX in this repo, modeled on the hve-core approach: `scripts/security/generate-vex.sh` (Trivy + Grype → OpenVEX) and `security/vex/inference-base.openvex.json`. This Trivy step (and `scan_image` in `container-cve-remediation.sh`) never passes `--vex`, so there is no suppression channel for CVEs the security team has assessed `not_affected`.

Consequences:
- The daily job and PR gate re-surface maintainer-triaged, non-exploitable CVEs as noise.
- Once the gate flips from report-only to **blocking** (the stated follow-up), a `not_affected` CVE could block an unrelated PR with no in-repo waiver — exactly the scanner-noise-vs-real-exposure problem the hve-core VEX model exists to solve.

Suggested changes:
1. Pass `trivy ... --vex <openvex>` in both channels (this `trivy-action` step and `scan_image`) so both honor `not_affected` / `fixed`.
2. Unify the image set: `discover-base-images.sh` is a clean single source of truth — feed the same four real bases (python, node, nginx, azureml) into `generate-vex.sh` triage. Today only the `scratch` inference base has a VEX doc (empty by design); the four scannable bases have none.
3. Keep the human-in-the-loop accountability the hve-core model requires — the daily auto-PR/issue flow (no auto-merge) already does this for remediation; VEX makes the suppression side equally human-owned.

scripts/security/container-cve-remediation.sh
```diff
@@ -0,0 +1,429 @@
...
+# open a digest-bump PR when a newer clean digest exists, otherwise open/refresh
+# a security issue. Invoked by .github/workflows/container-cve-remediation.yml.
+set -o errexit -o nounset -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || (cd "$SCRIPT_DIR/../.." && pwd))"
+gh_token="${GH_TOKEN:-${GITHUB_TOKEN:-}}"
+unset GH_TOKEN GITHUB_TOKEN
+# shellcheck source=../../scripts/lib/common.sh
+source "$REPO_ROOT/scripts/lib/common.sh"
```
katriendg: **RI-1 — `.github/scripts/` is a new, off-convention location** (Conventions / Maintainability, Medium)

`.github/scripts/` is a brand-new directory — the repo has no prior precedent for it. We already have two established homes for exactly this profile, and this script even reaches back into one of them (`source "$REPO_ROOT/scripts/lib/common.sh"` on this line):

- `shared/ci/` holds the bash CI-gate helpers (`smoke-docker.sh`, `smoke-image.sh`, `smoke-import.sh`), which source the same `scripts/lib/common.sh` and ship a `tests/` folder with fixtures.
- `scripts/security/` holds the security tooling (`generate-vex.sh`, `Test-*.ps1`).
- `copilot-instructions.md` defines `scripts/` as "CI/CD scripts, shared libraries, linting, security, and Pester tests"; `.github/scripts` is listed nowhere.

Suggest relocating all three scripts to `shared/ci/` (CI-gate profile) or `scripts/security/` (security profile) so they sit beside `common.sh` and inside the existing test harness. `shellcheck` already covers them either way (the linter globs `*.sh` recursively), so this is purely layout consistency and test-harness reuse. Workflow `run:` paths and the `# shellcheck source=` comment would need updating to match.

.github/workflows/container-cve-remediation.yml
```diff
@@ -0,0 +1,84 @@
...
+  schedule:
+    # Daily at 06:00 UTC
+    - cron: "0 6 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}
```
katriendg: **RI-5b — confirm force-push safety assumptions** (Reliability, Low — non-blocking)

The `git push --force` (not `--force-with-lease`) in `container-cve-remediation.sh` is justified in-comment (shallow checkout has no remote-tracking ref; branch re-derived from `origin/main` each run; only reached after confirming no open PR; serialized by this concurrency group). This is acceptable. Just confirm no other workflow shares the concurrency group name `${{ github.workflow }}` semantics — here it resolves per-workflow so serialization is correct.

.github/workflows/container-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+        with:
+          scan-type: image
+          image-ref: ${{ matrix.image }}
+          # Pin the scanner build; container-cve-remediation.yml pins the same version.
+          version: v0.72.0
+          format: sarif
+          output: trivy-${{ steps.slug.outputs.slug }}.sarif
+          severity: ${{ inputs.severity }}
+          vuln-type: os,library
```
katriendg, line 110: **RI-4 — express Trivy scan scope the same way in both channels** (Maintainability / Correctness, Low)

Both channels pin Trivy `v0.72.0`, but the PR gate uses the older `trivy-action` input `vuln-type: os,library` (this line) while the CLI path (`scan_image` in `container-cve-remediation.sh`) uses the modern `--scanners vuln` (no explicit package-type filter). They are functionally equivalent today, but the PR's own "Trivy parity" claim depends on identical findings, and expressing scope two different ways invites future drift. Suggest aligning both on the modern form (`--scanners vuln` / `--pkg-types os,library`) or adding a one-line comment asserting the equivalence.

katriendg (commented):
Really strong PR — the security engineering here is the highlight. The scan → plan → apply split that keeps the write token out of the scanning phase, the token isolation (`unset` then re-inject only through `gh`/`git` wrappers), the "only bump when the candidate digest itself scanned clean" TOCTOU guard, the conditional App-token minting, and the documented force-push safety are all done well. Trivy is pinned to `v0.72.0` in both channels for parity. None of my comments below are correctness blockers; they are about **repository-fit, reuse of what we already ship, and the runtime prerequisites** that need to be in place before the daily job can actually open a PR or issue.

On the repo's security posture for Action-created PRs: this is correctly modeled. The default `GITHUB_TOKEN` stays read-only and the workflow mints the scoped `RELEASE_APP_ID` App token (the same one release-please already uses in `main.yml`), minted only when remediation is needed and explicitly downscoped to `contents/issues/pull-requests: write` — actually tighter than `main.yml`, which requests the App's full install set. Using the App token (not `GITHUB_TOKEN`) is also the right call because App-token PRs trigger the `container-scan` gate, whereas `GITHUB_TOKEN` PRs would not. Permissions are limited and not over-exposed at the *token* level. The open question is at the *identity* level: should we reuse the release App for a second, unrelated purpose, or provision a dedicated security-remediation App? That plus two concrete prerequisites need settling before the first real remediation (see RI-6).

Summary of items (details inline):
- 🔧 **RI-1** — Move scripts out of the new `.github/scripts/` into an existing home (`shared/ci/` or `scripts/security/`).
- 🧪 **RI-2 / RI-5** — Consider PowerShell + Pester (or at least fixture tests) for the pure logic; it currently has zero coverage.
- 🔒 **RI-3** — Wire in our existing VEX so maintainer `not_affected` findings are suppressed (important before the gate flips to blocking).
- ⚖️ **RI-4** — Align the Trivy scan-scope flags across the two channels.
- ❓ **RI-6** — Decide whether to reuse `RELEASE_APP_ID` or provision a dedicated security App; confirm the `security` label exists and the App has `Issues: write` before merge.
- 🔁 **RI-5b** — Non-blocking: confirm the force-push concurrency assumptions.

Not approving yet — leaving RI-3 and RI-6 as the two items to resolve before the gate flips from report-only to blocking.

scripts/security/container-cve-remediation.sh
algattik: Relocated all three scripts to `scripts/security/` (beside `generate-vex.sh`, which sources the same `common.sh`). Updated the `run:` paths in `container-scan.yml` and `container-cve-remediation.yml`, plus the `pr-validation.yml` path filter that path-gates the `container-scan` job.

`shellcheck` still passes with no `SC1091`: the `# shellcheck source=../../scripts/lib/common.sh` directive resolves unchanged because both the old (`.github/scripts/`) and new (`scripts/security/`) locations are two levels below the repo root. Tests for the two pure helpers now live beside the other security Pester suites in `scripts/tests/security/` (see the RI-2 thread).

Addressed in 6d3791eb.

.github/workflows/container-scan.yml
algattik: Aligned the scan scope so both channels provably report the same findings:

- `container-scan.yml`: added `scanners: vuln` to the `trivy-action` step. Without it the action fell back to Trivy's image default, which also runs the secret scanner — a real (not just cosmetic) divergence from the CLI path's `--scanners vuln`.
- `container-cve-remediation.sh` `scan_image`: added `--pkg-types os,library` so the package scope is explicit on both sides rather than relying on the CLI default.

Both channels now express: `scanners=vuln`, `pkg-types=os,library`. The only remaining naming difference is forced by the action's API — this pinned `trivy-action` version still calls the input `vuln-type` (the upstream rename to `pkg-types` is a TODO in its `action.yaml`), so I added a comment on the step asserting the equivalence.

Addressed in 6d3791eb.

scripts/security/image-slug.sh
algattik: Added the two cheapest / highest-value suites you flagged, beside the other security Pester tests in `scripts/tests/security/`:

- `image-slug.Tests.ps1` — slug shape (punctuation collapse, registry host/port, hyphen trimming), a golden assertion proving the suffix is a SHA-256 of the *full* ref (digest included, so it can't regress to hashing the tag only), determinism, and the tag-collision disambiguation you called out (two refs whose tags collapse to the same alphanumeric slug get distinct slugs).
- `discover-base-images.Tests.ps1` — the awk `FROM` parser: digest extraction, stage-alias / `ARG` / `scratch` exclusion, case-insensitive `FROM`, registry-with-port, dedup across multiple Dockerfiles, and the empty-repo case.

They invoke the bash scripts inside a throwaway `git` work tree on the ubuntu Pester runner (`scripts/tests` is already discovered by the suite) and `-Skip` cleanly where `bash`/`git` are unavailable, so local non-Linux runs don't spuriously fail. 16 new tests, all green under Pester 5.7.1.

I left a full PowerShell port (and the `CIHelpers.psm1`/SARIF-exporter reuse) out of scope — testing the shipped bash directly gives the coverage without a rewrite, and the `gh`/`git` apply phase stays bash. Happy to revisit the port separately if you'd prefer it.

Addressed in 6d3791eb.

.github/workflows/container-cve-remediation.yml
algattik: Confirmed — no cross-workflow collision. `${{ github.workflow }}` resolves to each workflow's `name:` field, which is `Container CVE Remediation` for this file and is unique to it. The three other workflows that key on `${{ github.workflow }}` (`codeql-analysis`, `check-binary-integrity`, `scorecard`) each additionally append `-${{ github.ref }}` *and* resolve `github.workflow` to their own distinct names, so none can share this group. Serialization via `group: ${{ github.workflow }}` with `cancel-in-progress: false` is correct for this daily job.

.github/workflows/container-scan.yml
algattik: Agree this belongs before the gate flips from report-only to blocking — leaving the thread open to track it as the pre-blocking follow-up rather than landing it here.

Wiring `trivy ... --vex` into both channels is mechanical, but the suppression channel is only meaningful once the four scannable bases (python, node, nginx, azureml) actually have triaged OpenVEX docs. `generate-vex.sh` emits `under_investigation` stubs that the security team must individually assess to `not_affected` / `affected` / `fixed` before any of them can suppress a finding. Landing auto-generated, un-triaged VEX (or `--vex` pointed at a file that doesn't yet exist) would either be a no-op or silently assert judgements no human made — the opposite of the human-in-the-loop model. So the honest unit of work here is: generate VEX for the four bases via `generate-vex.sh`, triage each statement, commit, then wire `--vex` into the `trivy-action` step and `scan_image`.

Because the PR gate is still `soft-fail: true`, no `not_affected`-but-un-suppressed CVE can block an unrelated PR in the meantime, so this doesn't gate the current report-only PR. Keeping open to track the VEX-generation + triage + wiring ahead of the blocking flip.

.github/workflows/container-cve-remediation.yml
algattik: Handled the decision-independent, documentable part and left the governance decision open.

Added to `docs/security/workflow-permissions.md` (alongside the App-token table) a note that the "Allow GitHub Actions to create and approve pull requests" repo/org toggle is intentionally *not* required here: it governs the default `GITHUB_TOKEN`, not App tokens, and the App token is chosen partly because App-token PRs trigger `pr-validation.yml` (including the `container-scan` gate) whereas `GITHUB_TOKEN` PRs would not.

The two items that remain are owner calls, so I'm deliberately not deciding them in-PR and leaving this thread open:

- App identity — reuse `RELEASE_APP_ID` vs. provision a dedicated security-remediation App. This is the identity-level blast-radius / audit-trail trade-off you laid out; a maintainer/security-owner decision.
- Prerequisites before the first non-quiet run — the `security` label must exist (creation is a repo action, and `gh` won't auto-create it), and the chosen App must have `Issues: write` installed (`create-github-app-token` can only narrow, never add). Whichever App is chosen, its rationale should be documented alongside the note above.

Doc note in 6d3791eb; leaving the decision + provisioning to the owners.

algattik: Agreed this needs a decision before the daily job can run, so it's been split out of this PR into #1117 to unblock the scanning channel. That issue owns the App-identity choice (reuse `RELEASE_APP_ID` vs a dedicated security-remediation App, weighing audit-trail/blast-radius) and the runtime prerequisites (the `contents`/`issues`/`pull-requests` write scopes and the `security` label). The remediation workflow and script are removed here and will land there.

.github/workflows/container-scan.yml
algattik: Agreed — split into #1118, which must land before the gate flips to blocking. It covers triaging each fixable CRITICAL/HIGH finding on the four bases, recording the non-exploitable verdicts as committed OpenVEX (`not_affected` + justification) consumed by the scan, bumping digests where a clean fix exists, then removing the report-only override. This PR keeps the gate report-only in the meantime, so untriaged findings surface without blocking.

scripts/security/discover-base-images.sh
```diff
@@ -0,0 +1,31 @@
...
+
+awk '
+  /^[[:space:]]*[Ff][Rr][Oo][Mm][[:space:]]/ {
+    line = $0
+    while (match(line, /([A-Za-z0-9.-]+(:[0-9]+)?\/)?[A-Za-z0-9._\/-]+(:[A-Za-z0-9._-]+)?@sha256:[0-9A-Fa-f]{64}/)) {
+      print substr(line, RSTART, RLENGTH)
+      line = substr(line, RSTART + RLENGTH)
+    }
+  }
+' "${dockerfiles[@]}" | sort -u
```
katriendg: **Bug (portability) — the awk `{64}` interval makes this parser emit nothing under mawk, silently no-op'ing the scan.**

The digest match relies on the `{64}` interval expression. Under **mawk 1.3.4** — the default `/usr/bin/awk` on Debian/Ubuntu, including this repo's own devcontainer — interval expressions are unsupported (and can't be enabled: `-W interval` is a "vacuous option" and `POSIXLY_CORRECT` has no effect on this build), so the regex never matches and the script prints **zero** images with exit 0.

Impact: `container-scan.yml`'s discover step then sets `count=0`, the scan job's `if: needs.discover.outputs.count != '0'` skips, and the gate goes **green having scanned nothing** — a silent failure of a security control. CI happens to pass only because the GitHub runner's `awk` supports intervals; the new `discover-base-images.Tests.ps1` (4 of its cases) **fail locally in the devcontainer** for the same reason.

Fix: drop awk and extract with `grep -oiE`, whose interval support is not mawk-dependent (verified: reproduces the exact four refs and passes the Pester suite under mawk). The `|| true` preserves the original "no digest-pinned bases → exit 0" behavior under `set -o pipefail`:

```suggestion
printf '%s\0' "${dockerfiles[@]}" \
  | xargs -0 grep -hiE '^[[:space:]]*FROM[[:space:]]' \
  | grep -oiE '([A-Za-z0-9.-]+(:[0-9]+)?/)?[A-Za-z0-9._/-]+(:[A-Za-z0-9._-]+)?@sha256:[0-9a-f]{64}' \
  | sort -u || true
```

(Non-blocking for the approval — the report-only gate means no false green blocks a PR today — but worth fixing before merge so the shipped tests pass in the devcontainer and the gate can't silently scan nothing on a mawk runner.)

katriendg (approved):
Approving — the follow-up round addressed the review cleanly: scripts relocated to `scripts/security/` (RI-1), Pester coverage added for the slug and FROM-parser helpers (RI-2/RI-5), Trivy scan scope aligned across channels with the secret-scanner divergence fixed too (RI-4), and the auto-remediation channel (with the App-identity/prerequisite and force-push concerns, RI-6/RI-5b) split cleanly into #1117. VEX suppression (RI-3) is tracked in #1118 to land before the gate flips from report-only to blocking. Good, low-risk scoping.

One new inline bug to close before merge: `discover-base-images.sh` silently emits nothing under mawk (the devcontainer's default awk), which no-ops the scan and fails 4 of the newly-added Pester cases locally. Left an inline suggestion with a verified mawk-safe `grep` fix. It doesn't block the approval (the gate is report-only), but please resolve that thread before merging.

scripts/security/discover-base-images.sh
algattik: Fixed in `0b79b16f` — the `FROM` parser now extracts with `grep -oiE` instead of awk, so it no longer depends on interval-expression support.

Confirmed the failure and the fix against the exact runtime you flagged: under Debian bookworm's `/usr/bin/mawk` (1.3.4, the devcontainer's `python:3.12-bookworm` awk) the old `{64}` parser emitted **0 refs at exit 0** — the silent-green scan you described — while the `grep` version emits the correct four digest-pinned bases. Output is byte-identical across mawk, gawk and BSD awk, and `discover-base-images.Tests.ps1` passes 8/8.

`|| true` preserves the original "no digest-pinned bases → exit 0" behaviour under `pipefail`.

.github/workflows/container-cve-remediation.yml
WilliamBerryiii: I'd prefer to provide a restricted/limited app id for this. The main challenge will be that we have routinely dealt with dependencies that would not patch their vulnerabilities for several releases (e.g. the crypto 4.8 library update that froze AML for about 2 months). One other concern to think about in this is that all the operational apps right now are registered as private apps that "I" own ... which for a "hit by a bus" factor, is pretty brittle. I'd rather default right now in this operational environment to "stupid secure" so that there are never any questions about crossing boundaries or isolation ... we have a ton of threat vectors that are equally as meaningful in both supply chain and developer environments. The scanning has historically treated both direct supply chain and developer environments as equally weighted, because of the complexity here and that repo users may not be operating from environments that isolate and pre-scan dependencies ... so I've wanted to constantly err towards caution, despite the pain this causes everyone.

WilliamBerryiii: To register the app, I'll need a brief of all the details on registering it. @algattik you can send these details to be in a private channel so I can evaluate and make sure I've documented for the other maintainers a backup in case I don't show up to work one day 🤣

PR #1107, opened 2026-07-03
security(scripts): add workflow-npm-commands pinning detector

### LINKED ISSUE #1103: security(scripts): adopt hve-core workflow-npm-commands detector and remediation suggestions
Discovered during review of #1092 (closes #1091).

## Background

The repo's `scripts/security/Test-DependencyPinning.ps1` already exceeds `microsoft/hve-core`'s equivalent in dependency-type coverage (repo adds `docker`, `gh-extension`, `powershell-modules`, `shell-inline-pip`), so no wholesale swap is warranted. However, hve-core has two capabilities worth cherry-picking:

- `workflow-npm-commands` / `Get-WorkflowNpmCommandViolations` — flags `npm install` / `npm update` in workflow `run:` blocks and requires `npm ci`. Repo CI already uses `npm ci`; this detector would prevent regressions. It is the one detector hve-core has that this repo lacks.
- `-Remediate` / `Get-RemediationSuggestion` — resolves an action tag to a commit SHA via the GitHub API and emits a ready-to-paste `uses: name@<sha>` pin.

Reference: [hve-core Test-DependencyPinning.ps1](https://github.com/microsoft/hve-core/blob/main/scripts/security/Test-DependencyPinning.ps1).

## Proposed work

- Add a `workflow-npm-commands` detector (indentation-aware `run:` block parsing) with fixtures under `scripts/tests/Fixtures/Workflows/`.
- Optionally add an opt-in `-Remediate` mode that resolves tags to SHAs via the GitHub API (respecting `GITHUB_TOKEN`), emitting suggestions without failing the scan by default.
- Preserve the current wrapper split: keep detection/reporting in `Test-DependencyPinning.ps1` and CI glue in `Invoke-DependencyPinningWorkflow.ps1`.

## Acceptance criteria

- New detector flags unpinned `npm install/update` in workflow `run:` blocks and passes on `npm ci`, with Pester fixtures.
- If implemented, `-Remediate` produces SHA-pinned suggestions and is off by default.

algattik:
Closes #1103.

Adopts the one dependency-pinning detector `microsoft/hve-core` has that this repo lacked: `workflow-npm-commands`. It flags unpinned `npm install` / `npm update` in GitHub Actions `run:` steps and requires `npm ci`, so a lockfile-deterministic install cannot silently regress to a mutating one. The repo's `Test-DependencyPinning.ps1` already exceeds hve-core on dependency-type coverage (`docker`, `gh-extension`, `powershell-modules`, `shell-inline-pip`), so this cherry-picks the missing detector rather than swapping the scanner.

### New `workflow-npm-commands` detector

- Adds `Test-NpmCommandLine` (single-line classifier) and `Get-WorkflowNpmCommandViolations` (workflow walker) to `scripts/security/Test-DependencyPinning.ps1`.
- Flags the mutating forms `npm install`, `npm i`, `npm update`, and `npm install-test`; passes on `npm ci`. A trailing non-word/non-hyphen boundary keeps compliant hyphenated subcommands such as `npm install-ci-test` from being mis-flagged as `npm install`.
- Registers the type in the detector table and in the default `-IncludeTypes`, so the scan runs it by default.

### Workflow-parsing robustness

- Indentation-aware `run:` handling covers all three shapes: inline `run: npm install`, `- run:` list items, and multi-line `run: |` / `run: >` block scalars.
- Block boundary is computed on the `run:` key column (dash indent plus prefix length), so sibling step keys (`name:`, `env:`, `with:`) are not scanned as shell content.
- Block-scalar evasion guard: a `run:`-prefixed shell line **inside** an already-open block scalar is treated as shell content, not a new YAML key. Without this, such a line closed the block early and hid later npm commands in the same step.

### Exemptions and Windows shim

- Honors an inline `# pinning-ignore` on the command line and a `# pinning-ignore` on the line directly above the command, for both inline and block-scalar commands. A blank line between the marker and the command breaks the exemption (it applies only to the immediately following line), and an in-block marker does not leak across a step boundary.
- Recognizes the `npm.cmd` Windows shim (`npm.cmd install`) as equivalent to `npm`.

### Remediation

- The new detector plugs into the existing `-Remediate` path, emitting deterministic "use `npm ci`" guidance (no SHA to resolve, so it is flag-independent). The issue also listed hve-core's SHA-pinning `-Remediate`/`Get-RemediationSuggestion` as a candidate to port, but that capability already exists in this repo (GitHub Actions tag→commit-SHA resolution via the API, honoring `GITHUB_TOKEN`, off by default); only the npm branch was added.

The wrapper split is preserved: detection/reporting stays in `Test-DependencyPinning.ps1`; the CI glue in `Invoke-DependencyPinningWorkflow.ps1` is untouched.

## Docs

- `docs/security/README.md` documents the detector and its inline / above-line `# pinning-ignore` usage.

## Tests and fixtures (test-only)

- Adds `scripts/tests/Fixtures/Workflows/npm-commands-unpinned.yml` (expects 3 violations) and `npm-commands-compliant.yml` (expects 0, includes a `- run: |` block with `npm ci` and an unrelated `env:` mentioning "npm install" to guard against false positives).
- Extends `scripts/tests/security/Test-DependencyPinning.Tests.ps1`: unit cases for the classifier (including the `npm install-ci-test` non-match), block-scalar evasion, list-item vs inline vs block forms, and the inline / above-line / non-leak `# pinning-ignore` behavior.

## Validation

- Full Pester suite green; `Test-DependencyPinning.Tests.ps1` passes in full.
- PSScriptAnalyzer clean on the changed files.
- Detector self-scan of the repo reports 0 `workflow-npm-commands` violations (CI already uses `npm ci`), so the gate is green on merge.
- `lint:md`, `lint:yaml`, and `spell-check` clean.

## Known limitations (out of scope)

- Global options placed between `npm` and the subcommand (`npm --prefix app install`, `npm -w pkg i`) are not detected. No such invocation exists in the repo today; deferred to avoid regex-complexity and false-positive risk.
- Alias forms (`npm add` / `up` / `upgrade`) are not covered. The detector targets the `install` / `i` / `update` / `install-test` set named in the issue.


bindsi (approved):
Great catch...thanks for adding the additional checks!!!

katriendg (approved):
Nice, thanks!

PR #1132, opened 2026-07-06
feat(build): add weekly hve-core upstream freshness check

### LINKED ISSUE #1121: Add an upstream-drift signal for vendored hve-core content and the RPI pin
Extend the repository's freshness-check pattern to cover the hve-core content it depends on, so that stale vendored files and an out-of-date persona pin surface on the existing weekly cadence and get refreshed deliberately.

## What this adds

- A weekly signal when the pinned hve-core RPI bootstrap ref falls behind its latest reviewed release, so the pin is advanced as a conscious, reviewed step.
- A weekly signal when a vendored, hve-core-authored file (the security, linting, and mock PowerShell modules) diverges from its upstream source, so upstream fixes — including security fixes — are noticed and pulled in.
- Delivery of both through the mechanism the repository already uses for binary, doc-date, and action-SHA freshness, surfacing as a tracked, actionable item on the same schedule.

## Context

The repository carries committed copies of [hve-core](https://github.com/microsoft/hve-core)-authored files and pins the persona bootstrap to a fixed ref. Its existing freshness checks cover binaries, documentation dates, and action-SHA staleness; adding upstream-content coverage completes that story for the hve-core dependency. The immediate first instance this surfaces is the current bootstrap pin, which is several months behind its latest [release tag](https://github.com/microsoft/hve-core/releases).

Follows #1094.

algattik:
## Summary

Adds a weekly CI check that flags when the repo's vendored `microsoft/hve-core` PowerShell modules have drifted from upstream, giving maintainers a single actionable signal instead of discovering drift by accident.

Closes #1121.

### New workflow — `.github/workflows/check-hve-core-freshness.yml` (behaviour)

- Runs weekly on a cron schedule and on `workflow_dispatch`.
- When drift is detected, files/updates a **single deduped tracking issue** (marker `automation:hve-core-freshness`; labels `dependencies,automated,needs-triage`), adds one comment per stale run, and **auto-closes** it once everything is current again.
- Deliberately does **not** open a PR: the vendored files are locally adapted, so there is no mechanical patch to auto-apply — a human must reconcile the change.

### New script — `scripts/security/Test-HveCoreFreshness.ps1` (behaviour)

- Emits a **content-based** drift signal for the vendored hve-core `.psm1` modules, comparing the **upstream blob SHA** at the pinned `UPSTREAM_REF` (the last reviewed hve-core ref, read from `copilot-setup-steps.yml`) against the same path in the newest non-draft `microsoft/hve-core` release. The local (adapted) copies are never consulted, so local edits do not cause false positives.
- Comparing content rather than release tags is the key design choice: the check goes stale only when a module the repo actually consumes changed upstream, not on every hve-core prerelease (~2/week).
- Writes results to `hve-core-freshness-results.json` for the workflow's issue steps.

### Scope — derived modules only

This check tracks the vendored `.psm1` modules and intentionally does **not** track the RPI persona/agent artifacts bootstrapped in `copilot-setup-steps.yml`. hve-core is migrating RPI from agents to skills (a breaking change landing in hve-core 4.0.0), so a persona-drift signal would fire for a known, planned change and its "re-pin" remedy would be wrong — the correct response is to adopt the skill. That adoption is tracked separately in #1151.

### Docs — `scripts/README.md`

- Documents the script and its content-based comparison.

## Tests (test-only)

- `scripts/tests/security/Test-HveCoreFreshness.Tests.ps1` — Pester coverage for pinned-ref parsing, blob-drift detection, and issue/job-summary formatting. 21 tests.

## Validation

- Pester **21/21**; PSScriptAnalyzer **0 errors / 0 warnings**; markdownlint and cspell clean.
- Live run against the current pin (v3.2.2) vs latest release (v3.3.101) confirms intent: 4/6 derived modules flagged as genuinely drifted.


katriendg:
General comment: I like the freshness check for the scripts we reuse here, though I'm very much on the fence for the RPI or other agents/AI artifacts, mainly because of the speed of change we are having in HVE core. Things are moving over to skills, and potentially the RIP agent becomes more of a Skill with subagents, which means the usage of this needs to change in this repo as well. Not sure we should already take the effort. What do you think, does it make sense even if then it means we are surfacing a bigger change?

algattik:
Fair concern — I've cut the RPI/agents half from this PR.

Looking upstream confirmed it: hve-core is mid-migration from the RPI agent/persona model to skills. The skill-forward RPI workflows are already merged on `main` (`.github/skills/rpi/*`) and ship as a **breaking change** in the upcoming 4.0.0 release. A persona-drift signal would therefore fire for a change we already know is coming, and its implied remedy — re-pin to a newer release — would be wrong; the right response is to adopt the skill, not re-pin the agents.

So the check now tracks **only the vendored `.psm1` modules**, where the asymmetry holds: those files survive the restructure and genuinely need a drift signal (live run: 4/6 drifted between the current pin and the latest release).

Adopting the RPI skill for the cloud agent is tracked separately in #1151.

scripts/security/Test-HveCoreFreshness.ps1
```diff
@@ -0,0 +1,460 @@
...
+    }
+}
+
+function Resolve-RepoRoot {
+    [CmdletBinding()]
+    param([string]$Hint)
+
+    if ($Hint) { return (Resolve-Path $Hint).Path }
+
+    try {
```
rezatnoMsirhC, line 420: In `Invoke-HveCoreFreshnessCheck`, `$pin` only stores `PinnedTag` and `File`, not `PinnedSha`. If `Get-PinnedHveCoreRef` cannot match the `hve-core release:` comment (returns `'unknown'`), `Format-HveCoreIssueBody` emits a broken compare link:

```
compare/unknown...hve-core-v3.3.101
```

The compare URL is the primary "How to Refresh" guidance in the tracking issue, so a broken link substantially reduces the alert's actionability.

Suggested fix: include `PinnedSha` in `$pin` and fall back to it in the formatter when the tag is `'unknown'`.

```powershell
# In Invoke-HveCoreFreshnessCheck:
$pin = [ordered]@{
    PinnedTag = $pinRef.Tag
    PinnedSha = $pinRef.Sha   # add this
    File      = $script:SetupWorkflow
}
```

```powershell
# In Format-HveCoreIssueBody:
$compareBase = if ($pin.PinnedTag -ne 'unknown') { $pin.PinnedTag } else { $pin.PinnedSha }
...
Compare link: https://github.com/microsoft/hve-core/compare/$compareBase...$($Result.LatestTag)
```

scripts/security/Test-HveCoreFreshness.ps1
rezatnoMsirhC, line 1: Nit: `Format-HveCoreJobSummary` only emits `Derived File | Status` columns, while the tracking issue table includes short blob SHAs. A developer reviewing a failing Actions run needs to open the tracking issue to see which blobs changed. Aligning the summary table with the issue body table would make the Actions view self-contained.

```powershell
$fileRows = foreach ($f in $files) {
    $fStatus = Get-HveCoreStateLabel -State $f.State
    $pSha = if ($f.PinnedUpstreamSha) { $f.PinnedUpstreamSha.Substring(0, [Math]::Min(7, $f.PinnedUpstreamSha.Length)) } else { '' }
    $lSha = if ($f.LatestUpstreamSha) { $f.LatestUpstreamSha.Substring(0, [Math]::Min(7, $f.LatestUpstreamSha.Length)) } else { '' }
    "| $($f.Path) | $pSha | $lSha | $fStatus |"
}
```

With header:

```markdown
| Derived File | Pinned blob | Latest blob | Status |
|--------------|-------------|-------------|--------|
```

rezatnoMsirhC (approved):

scripts/security/Test-HveCoreFreshness.ps1
algattik: Fixed in a1ea9180. Added `PinnedSha` to the pin state and the formatter now falls back to it for the compare base when `PinnedTag` is `'unknown'`, so the compare link stays valid.

scripts/security/Test-HveCoreFreshness.ps1
algattik: Fixed in a1ea9180. `Format-HveCoreJobSummary` now emits `Pinned blob` and `Latest blob` short-SHA columns alongside Status, matching the tracking issue table so the Actions summary is self-contained.

Issue #1133, opened 2026-07-06
Extend hve-core freshness coverage to the vendored PowerShell security linters


algattik:
Follow-up to #1121 / #1132.

## Problem

The weekly hve-core freshness check added in #1132 surfaces drift for the
hve-core-derived `.psm1` modules and the RPI bootstrap personas. The
hve-core-derived `.ps1` security linters under `scripts/security/` are outside
that coverage, so when their upstream sources move ahead, nothing surfaces it —
the exact blind spot #1121 set out to close, left open for this file class.

## Gap

- The freshness check's derived-file set does not include the hve-core-derived
  `.ps1` security linters, so drift in them is undetected.
- The check's baseline model does not fit these files. It compares the upstream
  blob at a single global pinned ref against the newest release. At least one of
  these linters exists only on hve-core `main` and in no release tag, and others
  were adapted from `main` commits ahead of the global pin. A single-ref,
  release-anchored comparison cannot express their true source revision.

## Desired outcome

- Drift in the vendored `.ps1` security linters is surfaced by the same weekly
  signal that already covers the derived modules.
- The comparison baseline is each file's own recorded source revision, so files
  adapted from different upstream commits — including `main`-only files — are
  compared against the correct point. The workflow-permission and
  dangerous-workflow linters already record such a per-file source revision in
  their headers and can seed this model.
- Local adaptations and extensions (for example the untrusted-checkout rule) do
  not produce false positives.

The mechanism is open: generalize the existing check to read a per-file pinned
revision, or a central manifest keyed by upstream path and revision.

PR #1134, opened 2026-07-06
feat(scripts): add workflow permission and dangerous-pattern linters

### LINKED ISSUE #1122: Add workflow permission and dangerous-pattern linting
Add automated linting that verifies every GitHub Actions workflow declares least-privilege permissions and is free of known dangerous patterns, run as part of the existing PowerShell test suite.

## What this adds

- A CI check that verifies each workflow declares a least-privilege `permissions:` block, keeping the workflows and the `docs/security/workflow-permissions.md` reference in agreement automatically.
- A CI check that flags known dangerous workflow patterns — for example untrusted `${{ github.event.* }}` interpolation inside `run:` steps, and risky `pull_request_target` checkouts of untrusted code.
- Both delivered as Pester-tested linters through the mirrored-module pattern the repository already consumes from hve-core, and wired into `npm run test:ps` alongside the existing security tests.

## Context

The repository already enforces action SHA-pinning and maintains a hand-written workflow-permissions reference. Adding these checks turns that reference into an enforced invariant and catches permission-scope and injection risks at PR time. hve-core's [`Test-WorkflowPermissions.ps1`](https://github.com/microsoft/hve-core/blob/e6f414dabf65d67d59763ce776fa2212bd70b028/scripts/security/Test-WorkflowPermissions.ps1) and [`Test-DangerousWorkflow.ps1`](https://github.com/microsoft/hve-core/blob/e6f414dabf65d67d59763ce776fa2212bd70b028/scripts/security/Test-DangerousWorkflow.ps1) are a ready basis to adopt as pinned, vendored snapshots.

Follows #1094.

algattik:
## Description

Closes #1122.

Adds two PowerShell linters that enforce GitHub Actions security invariants in
CI, adapted from `microsoft/hve-core` and extended for this repository:

- **`Test-WorkflowPermissions.ps1`** — fails any workflow that omits a top-level
  `permissions:` block, so the read-only-by-default token posture is enforced
  rather than assumed.
- **`Test-DangerousWorkflow.ps1`** — flags two classes of dangerous pattern:
  template injection (direct interpolation of attacker-controllable event
  values such as `${{ github.head_ref }}` or `${{ github.event.* }}` into `run:`
  / `github-script` bodies) and untrusted `pull_request_target` checkouts of
  pull-request head code.

Both run in a dedicated CI workflow and under the local Pester suite, giving the
repository a machine-checked gate for the workflow-hardening invariants
documented in `docs/security/workflow-permissions.md`.

### Behaviour changes

**New linters (`scripts/security/`).** The two scripts above support `console`,
`json`, and `sarif` output, an `-ExcludePaths` filter, and `-FailOnViolation`
(non-zero exit for CI gating).

**Local extension — untrusted-checkout detection.** Beyond the upstream
template-injection matcher, `Test-DangerousWorkflow.ps1` adds an
untrusted-`pull_request_target`-checkout rule. Expression matching is fed
through a normalization helper that rewrites bracket/index access
(`github['event']['pull_request']`, `commits[0]`) to dot form before pattern
tests, so equivalent expressions cannot slip past the matcher by changing
syntax. Normalization is monotonic (brackets → dots only), so it cannot turn a
non-match into a match or vice versa.

**CI workflow (`.github/workflows/workflow-permissions-scan.yml`).** Runs both
linters as standalone `pwsh` steps, installing `powershell-yaml` pinned to
`0.4.12`. Each linter propagates its exit code, so a workflow introducing a
missing-permissions or dangerous pattern fails the check.

**Shared module touch-ups (behaviour-enabling, used by the new linters).**
- `SecurityClasses.psm1`: adds `MissingPermissions` to the `ViolationType`
  `ValidateSet`, so the permissions linter can report through the shared
  `DependencyViolation` class.
- `SecurityHelpers.psm1`: imports `CIHelpers.psm1` (without `-Force`, to avoid
  shadowing its standalone export) and adds a `-CIAnnotation` switch to
  `Write-SecurityLog` that forwards `Warning`/`Error` messages as GitHub Actions
  annotations.

**Upstream provenance pinning.** Each vendored linter records the exact
`microsoft/hve-core` source path and commit it was adapted from (in the header
`.DESCRIPTION` and a `.LINK` blob permalink). `Test-DangerousWorkflow.ps1` diverges
from that revision (the untrusted-checkout rule, the `Get-NodeMember` node-access
refactor, and expression normalization); these deltas are marked inline with
`# LOCAL` comments and enumerated in the file header, so a re-sync restructures
upstream function bodies and is a 3-way merge rather than a purely mechanical
diff. Coverage of the remaining vendored files by the
upstream freshness check is tracked separately in #1133 (follow-up to #1121 /
#1132).

### Test-only changes

- `Test-DangerousWorkflow.Tests.ps1` / `Test-WorkflowPermissions.Tests.ps1`:
  Pester suites covering every injection and untrusted-ref pattern (including
  bracket/index forms and `head_commit` / `commits.*` author sub-fields),
  precise violation line numbers, the three output formats, malformed-YAML
  handling, and the repository invariant that every workflow declares a
  permissions block.
- `WorkflowSecurityTestHelpers.psm1`: repository-authored fixture helpers
  (`New-WorkflowFixture`, `Get-JsonReport`, `New-RunInjectionWorkflowContent`)
  shared across the suites.
- `scripts/tests/Fixtures/Workflows/*.yml`: permissions fixtures.
- `SecurityClasses.Tests.ps1`: asserts the new `MissingPermissions` enum value.

### Validation

- Pester (`scripts/tests/security`): 541 passed, 0 failed, 1 skipped.
- PSScriptAnalyzer: 0 errors / 0 warnings (repo config; warnings fail CI).
- `actionlint`, `markdownlint-cli2`, `cspell`, and frontmatter validation: clean.

## Type of Change

## Component(s) Affected

- [x] `docs/` - Documentation

## Documentation Impact

## Checklist


scripts/security/Test-WorkflowPermissions.ps1
```diff
@@ -0,0 +1,339 @@
...
+    [string]$Format = 'json',
+
+    [Parameter(Mandatory = $false)]
+    [string]$OutputPath = 'logs/workflow-permissions-results.json',
+
+    [Parameter(Mandatory = $false)]
+    [switch]$FailOnViolation,
+
+    [Parameter(Mandatory = $false)]
+    [string]$ExcludePaths = ''
```
katriendg, line 73: Maintainability (nit): `ExcludePaths` defaults to `''` here vs upstream's `'copilot-setup-steps.yml'`. This is an intentional repo-specific delta (every workflow here declares permissions), which is correct — a short header note recording *why* it diverges keeps it deliberate-by-record when this file is next re-synced from hve-core.

scripts/security/Test-DangerousWorkflow.ps1
```diff
@@ -0,0 +1,643 @@
+#!/usr/bin/env pwsh
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: MIT
+#Requires -Version 7.0
```
katriendg, line 4: Maintainability (nit): both linters lower `#Requires -Version 7.4` (upstream) to `7.0`. This matches the repo standard (e.g. `Test-BinaryFreshness.ps1`) and is fine, but it's an undocumented delta that will re-conflict on every upstream sync. A one-line header note ("version requirement lowered from upstream 7.4 to repo standard 7.0") preserves the intent — and worth confirming no upstream 7.4-only syntax is relied upon. Same applies to `Test-WorkflowPermissions.ps1` line 4.

scripts/security/Test-DangerousWorkflow.ps1
```diff
@@ -0,0 +1,643 @@
...
+)
+
+$ErrorActionPreference = 'Stop'
+
+Import-Module (Join-Path $PSScriptRoot '../lib/Modules/CIHelpers.psm1') -Force
+Import-Module (Join-Path $PSScriptRoot 'Modules/SecurityHelpers.psm1') -Force
+
+#region Functions
+
+function Get-NodeMember {
```
katriendg, line 86: Maintainability (non-blocking): the `Get-NodeMember` refactor here, plus the `ConvertTo-NormalizedExpression` call inserted into `Test-IsUntrustedInjectionExpression`, restructure the upstream function bodies rather than only appending to them. Combined with the checkout rule below, a future re-sync against a newer hve-core revision becomes a 3-way merge rather than the "mechanical diff" the description implies. To keep this genuinely re-syncable, consider keeping the upstream bodies byte-identical and layering local behavior as append-only wrappers/post-passes, or marking each local delta with explicit `# LOCAL:` begin/end markers. At minimum, worth softening the "mechanical diff" wording in the PR description.

scripts/security/Test-DangerousWorkflow.ps1
```diff
@@ -0,0 +1,643 @@
...
+                            }
+
+                            $violation = New-DangerousWorkflowViolation -File $relativePath -Line $lineNumber -RuleId 'dangerous-workflow/template-injection' -Description "Untrusted expression '$expression' is interpolated into a code execution context in job '$jobName' step '$stepName'." -Remediation 'Avoid directly interpolating untrusted GitHub event or workflow-output values into shell or script blocks.' -JobName $jobName -StepName $stepName
+                            $violations += $violation
+                            break
+                        }
+                    }
+                }
+
+                # Untrusted checkout: a pull_request_target workflow that checks out the PR head ref
```
katriendg, line 559: Design (confirmed, keep as-is): upstream deliberately omits checkout detection because it delegates that to the Poutine scanner in CI. I confirmed this repo runs **no Poutine scanner** — the only Poutine references are inert `# poutine:ignore untrusted_checkout_exec` annotations in the generated `aw-dependabot-pr-review.lock.yml`. So this rule is a justified gap-fill, not duplication — good to keep. Suggest a one-line header note that it intentionally diverges from upstream's Poutine delegation because Poutine isn't run here, so the divergence is deliberate-by-record on the next sync.

scripts/security/Modules/SecurityHelpers.psm1
```diff
@@ -9,6 +9,9 @@
 
 #Requires -Version 7.0
 
+# Omit -Force so the standalone CIHelpers export is not shadowed by a nested re-import.
+Import-Module (Join-Path $PSScriptRoot '../../lib/Modules/CIHelpers.psm1')
```
katriendg, line 17: Reliability (low): importing `CIHelpers.psm1` without `-Force` here, while both linters import it *with* `-Force` and then import this module, creates a subtle load-order coupling. The comment explains the intent, but it's worth confirming `Write-CIAnnotation` resolves to a single definition across all three entry paths (standalone linter, dot-sourced test, module-only import). A one-line note on the load-order contract would help future readers.

scripts/tests/security/WorkflowSecurityTestHelpers.psm1
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "`${{ $Expr }}"
+"@
+}
+
+Export-ModuleMember -Function @(
+    'New-WorkflowFixture',
+    'Get-JsonReport',
+    'New-RunInjectionWorkflowContent'
```
katriendg: Nit: file is missing a trailing newline (`\ No newline at end of file`). Add one to match repo convention and keep linters happy.

katriendg (commented):
Thanks for this, @algattik — this is a well-executed hardening PR. Turning the hand-maintained `docs/security/workflow-permissions.md` reference into an enforced, machine-checked invariant is exactly the right direction, and the execution is thorough: 541 Pester assertions across three output formats, malformed-YAML handling, bracket/index expression forms, precise line-number assertions, and a repo-invariant test that every workflow declares a `permissions:` block. Reuse of the shared `DependencyViolation` / `ComplianceReport` / `Write-SecurityLog` plumbing (rather than re-implementing it) keeps this consistent with the existing mirrored-module pattern, and the per-file upstream provenance pins are a genuine improvement over the older vendored files.

A few non-blocking notes inline. The themes:

- **Upgradeability of `Test-DangerousWorkflow.ps1`.** The local changes (normalization helper, `Get-NodeMember` refactor, checkout rule) are interwoven with the upstream function bodies, so a future re-sync is a 3-way merge rather than the "mechanical diff" the description implies. A couple of options inline to keep it genuinely re-syncable.
- **The untrusted-checkout rule is a *justified gap-fill*, not duplication.** I confirmed this repo runs no Poutine scanner (upstream delegates checkout detection to Poutine; here only inert `poutine:ignore` annotations exist in the generated `aw-dependabot-pr-review.lock.yml`). Keeping the rule is the right call — just worth a header note that it intentionally diverges from upstream because Poutine isn't run here.
- **Patch coverage lands at ~89%** (42 lines, concentrated in the two new scripts — `Test-DangerousWorkflow.ps1` 88.44% / 29 missing, `Test-WorkflowPermissions.ps1` 90.51% / 13 missing). The misses look like the non-happy-path branches: SARIF/console formatting, output-directory creation, file-write paths, the malformed-YAML `catch`, the missing-`ConvertFrom-Yaml` early return, and the `-FailOnViolation` soft-fail vs. fail exit-code branches. A few targeted cases would lift this into the project's 88–100% band and lock in the CI-gating behavior these linters exist to provide.
- Minor: the `#Requires` downgrade and the CIHelpers import-order coupling would each benefit from a one-line note so intent survives the next sync; and one file is missing a trailing newline.

On broader dangerous-workflow coverage: since this repo doesn't run Poutine, we've got a backlog issue drafted and ready to evaluate adopting it — happy to file it if you'd like it on the backlog, otherwise no action needed here.

Recommendation: approve-with-nits — none of these are correctness blockers.

scripts/security/Test-WorkflowPermissions.ps1
algattik: Added a note on the `-ExcludePaths` parameter in the header recording that the empty default diverges from upstream's `copilot-setup-steps.yml`, and why: every workflow in this repository declares a permissions block. That keeps the divergence deliberate-by-record on the next hve-core sync. (commit b145e88b)

scripts/security/Test-DangerousWorkflow.ps1
algattik: Added a one-line note beneath `#Requires -Version 7.0` in both linters recording the lowering from upstream 7.4 to the repo standard (cf. `Test-BinaryFreshness.ps1`). Confirmed no 7.4-only syntax is relied upon: scanned both scripts and `SecurityHelpers.psm1` for version-sensitive constructs (`clean {}` blocks, `-DateKind`, `ConvertTo/FromCliXml`, `-ProgressAction`, `??=`, …) — none present; only 7.0-compatible constructs are used, and the Pester suite passes. (commit b145e88b)

scripts/security/Test-DangerousWorkflow.ps1
algattik: Marked each local delta with explicit `# LOCAL` comments — `Get-NodeMember`, `ConvertTo-NormalizedExpression` and its insertion into `Test-IsUntrustedInjectionExpression`, and `# LOCAL (begin/end)` around the untrusted-checkout block — and added a header note enumerating these divergences and stating they restructure upstream bodies, so a re-sync is a 3-way merge rather than a mechanical diff. Also softened the "mechanical diff" wording in the PR description. (commit b145e88b)

scripts/security/Test-DangerousWorkflow.ps1
algattik: Added a header note recording that the untrusted-checkout rule intentionally diverges from upstream's Poutine delegation: upstream omits checkout detection because it relies on the Poutine scanner in CI, and this repository runs no Poutine scanner, so the rule fills that gap rather than duplicating it. The rule block is bracketed with `# LOCAL (begin/end)` markers. (commit b145e88b)

scripts/security/Modules/SecurityHelpers.psm1
algattik: Expanded the comment in `SecurityHelpers.psm1` to document the load-order contract explicitly. Confirmed `Write-CIAnnotation` resolves to a single definition: it is defined and exported only by `CIHelpers.psm1` and is merely *called* by `SecurityHelpers.psm1` (never redefined or re-exported — its `Export-ModuleMember` lists five other functions). So all three entry paths (standalone linter, dot-sourced test, module-only import) bind the same function regardless of load order. (commit b145e88b)

scripts/tests/security/WorkflowSecurityTestHelpers.psm1
algattik: Added the missing trailing newline to `WorkflowSecurityTestHelpers.psm1`. (commit b145e88b)

nguyena2 (approved):

algattik:
Follow-up on the patch-coverage note: coverage on the two scripts is now 96% (was ~89%), inside the project band, in 0a362d20. Against the specific misses called out:

- Console and default output-path formatting for both linters, including `logs/` directory creation when `-OutputPath` is unsupplied.
- The missing-`ConvertFrom-Yaml` early return (exit 1).
- Direct unit tests for the local helpers: `Get-NodeMember` (PSObject/null paths), `Get-ExpressionMatches`, `Test-IsUntrustedInjectionExpression` / `Test-IsUntrustedCheckoutRef` (whitespace), `Test-HasPullRequestTargetTrigger` (boolean-folded / scalar / list `on` forms), and `Find-NextMatchingLine`.
- Both CLI entry points exercised as subprocesses (exit 0 on a clean scan, exit 1 on a non-existent path) via a shared `Invoke-SecurityLinterScript` test helper.

Two categories remain uncounted by design. The entry-point wrappers are asserted through the subprocess tests but do not register in the in-process coverage instrument since they run in a separate process. The injection/ref line-resolution fallbacks and the file-read `catch` are left uncovered because reaching them requires mocking internal functions or forcing nondeterministic hashtable ordering, which the repo's "mocks reserved for external dependencies" convention discourages. Malformed-YAML parsing and the `-FailOnViolation` soft-fail vs. fail exit branches were already covered.

On the Poutine offer: yes, please file the drafted backlog issue. Evaluating adoption belongs on its own track rather than in this PR, and having it recorded keeps the intentional divergence documented here traceable to a follow-up.

scripts/security/Test-WorkflowPermissions.ps1
```diff
@@ -0,0 +1,342 @@
...
+    [Parameter(Mandatory = $false)]
+    [string]$Path = '.github/workflows',
+
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('json', 'sarif', 'console')]
+    [string]$Format = 'json',
+
+    [Parameter(Mandatory = $false)]
+    [string]$OutputPath = 'logs/workflow-permissions-results.json',
+
```
rezatnoMsirhC, line 68: `$OutputPath` defaults to `'logs/workflow-permissions-results.json'` and is never adjusted when `-Format sarif` is passed. Calling this script with `-Format sarif` writes SARIF content into a `.json` file.

`Test-DangerousWorkflow.ps1` handles this correctly: `$OutputPath` defaults to `''` and the format branch runs early in `Invoke-DangerousWorkflowCheck`. Suggest matching that pattern here:

1. Change the default in both the top-level `param` block and the `Invoke-WorkflowPermissionsCheck` `param` block to `''`.
2. Add a format branch at the top of `Invoke-WorkflowPermissionsCheck`:

```powershell
if ([string]::IsNullOrWhiteSpace($OutputPath)) {
    $OutputPath = if ($Format -eq 'sarif') {
        'logs/workflow-permissions-results.sarif'
    } else {
        'logs/workflow-permissions-results.json'
    }
}
```

.github/workflows/workflow-permissions-scan.yml
```diff
@@ -28,43 +28,23 @@ jobs:
...
-          if [ "$VIOLATIONS" -gt 0 ]; then
-            echo "::error::${VIOLATIONS} workflow(s) missing top-level permissions blocks"
-            if [ "${{ inputs.soft-fail }}" != "true" ]; then
-              exit 1
-            fi
-          else
-            echo "All workflows have top-level permissions blocks."
-          fi
+      - name: Validate dangerous workflow patterns
+        shell: pwsh
```
rezatnoMsirhC, line 45: Nit: this step uses `PERMISSIONS_SOFT_FAIL` as its env var name, but the name is semantically tied to the permissions check above it. Renaming to `DANGEROUS_WORKFLOW_SOFT_FAIL` (and updating the `$env:` reference in the `run:` block) would make the relationship between the var and the step it controls explicit at a glance.

scripts/security/Test-WorkflowPermissions.ps1
```diff
@@ -0,0 +1,342 @@
...
+    [string]$ExcludePaths = ''
+)
+
+$ErrorActionPreference = 'Stop'
+
+Import-Module (Join-Path $PSScriptRoot '../lib/Modules/CIHelpers.psm1') -Force
+Import-Module (Join-Path $PSScriptRoot 'Modules/SecurityHelpers.psm1') -Force
+
+# region Helper Functions
+
```
rezatnoMsirhC, line 82: Nit: `# region` with a space after `#` is treated as a plain comment by PowerShell and VS Code; it will not fold. `Test-DangerousWorkflow.ps1` uses the correct form: `#region Functions` / `#endregion Functions`. Change to `#region Helper Functions` / `#endregion` to match.

scripts/security/Test-WorkflowPermissions.ps1
algattik: Done in 442c12ef. Matched the `Test-DangerousWorkflow.ps1` pattern: `$OutputPath` now defaults to `\'` in both the top-level `param` block and the `Invoke-WorkflowPermissionsCheck` `param` block, and a format branch at the top of `Invoke-WorkflowPermissionsCheck` selects `logs/workflow-permissions-results.sarif` for `-Format sarif` and `logs/workflow-permissions-results.json` otherwise. Updated the `.PARAMETER OutputPath` doc accordingly and verified `-Format sarif` writes a `.sarif` file.

.github/workflows/workflow-permissions-scan.yml
algattik: Done in 442c12ef. Renamed the env var on the dangerous-workflow step to `DANGEROUS_WORKFLOW_SOFT_FAIL` and updated the `$env:` reference in its `run:` block, so the var name matches the step it controls.

scripts/security/Test-WorkflowPermissions.ps1
algattik: Done in 442c12ef. Changed the markers to `#region Helper Functions` / `#endregion` (no space after `#`) so they fold in PowerShell/VS Code, matching `Test-DangerousWorkflow.ps1`.

PR #1135, opened 2026-07-06
security(scripts): auto-apply GitHub Actions SHA pins

### LINKED ISSUE #1104: security(build): auto-remediate GitHub Actions SHA pins
Discovered during review of #1092 (closes #1091). Future / deferred — explicitly out of scope for #1092.

## Background

SHA-pinning validation already exists in this repo (`Test-DependencyPinning.ps1` validates `github-actions` SHA pins at `High` severity, and workflows SHA-pin their own actions). The natural next step, borrowed from hve-core's `-Remediate`, is to auto-suggest or auto-apply the exact SHA pin for any flagged action rather than only reporting it. Kept separate from #1103 because it targets the SHA-pinning enforcement story specifically and may overlap with existing SHA-pinning scan workflows.

## Acceptance criteria

- Given an unpinned or tag-pinned action, the tooling resolves and reports the immutable SHA pin.
- No behavior change to the failing / soft-fail gate unless explicitly enabled.

algattik:
Closes #1104.

## Summary

`Test-DependencyPinning.ps1` already resolves a tag-pinned GitHub Action to its immutable commit SHA and *reports* it as a suggestion (`-Remediate`). This adds the step the issue asked for — *applying* that pin — as an opt-in `-Apply` switch, fixes a resolution bug that silently affected monorepo (subpath) actions, and makes the remediation discoverable from the CI gate's failure summary.

The compliance gate is deliberately untouched: `-Apply` is opt-in, is never passed by CI, and does not change the report, the threshold, or the exit code.

### New: `-Apply` (opt-in auto-remediation)

- Adds `-Apply` to `Test-DependencyPinning.ps1`. When set, it rewrites tag/branch-pinned `uses:` references in place to the immutable form `uses: <name>@<sha> # <ref>`.
- New `Invoke-ActionShaPinApply` performs the rewrite: it operates on the specific violation line, uses word-boundary matching so `@v4` never clobbers `@v40` (and a short action name never matches inside a longer one), preserves the file's existing newline style (LF/CRLF) and trailing newline, and returns the number of pins applied.
- Wiring: `-Apply` implies resolution (`-Remediate:($Remediate -or $Apply)`), then an `if ($Apply)` block applies the pins after the scan. The apply step reads the already-resolved SHA from violation metadata, so it issues no extra network calls and is deterministic.

### Bug fix: subpath (monorepo) action resolution

- SHA resolution now targets the `owner/repo` slug (the first two path segments) instead of the full action path. Previously an action such as `github/codeql-action/init@v3` was queried as-is and returned 404, so no SHA was resolved or reported. This fix also corrects the existing `-Remediate` reporting path for these actions.
- The resolved SHA is stashed in the violation's `Metadata['ResolvedSha']`, so the report and the apply step share a single resolution.

### CI gate summary: discoverability

- The pinning gate's **Action Required** summary now names the remediation command (`... -IncludeTypes github-actions -Apply`) so a contributor blocked by the gate can self-remediate. This is text only — no network call is added to the gate, and the pass/fail logic is unchanged.
- That summary was converted to a single-quoted here-string with `-f` so the inline-code backticks render (a double-quoted here-string consumes them).

## Intentionally not changed

- The scheduled/PR scan (`Invoke-DependencyPinningWorkflow.ps1`) never passes `-Apply` or `-Remediate`; the gate stays hermetic (offline, deterministic) and its threshold/exit behavior is identical.
- No auto-PR bot: ongoing SHA-pin maintenance is already owned by Dependabot's `github-actions` updater. `-Apply` covers the gap Dependabot does not — the *initial* pin of a newly added tag-pinned action — as a manual fixer, consistent with the repo's existing `update-*.sh` scripts.

## Tests (test-only)

- 8 new Pester cases: 2 cover the resolver changes (subpath slug resolution; SHA stashed in metadata), and 6 cover `Invoke-ActionShaPinApply` (applies the pin; word-boundary safety including `v4`/`v40`; newline preservation; out-of-range and missing-file skips; return count). All are network-free.

## Validation

- Full security Pester suite: 487 passed / 0 failed / 1 skipped.
- PSScriptAnalyzer: 0 findings on both changed `.ps1` files.
- markdownlint + cspell: clean on the changed doc.
- Live end-to-end: `actions/checkout@v4` and subpath `github/codeql-action/init@v3` both resolved and rewrote correctly.

## Acceptance criteria

- Given an unpinned or tag-pinned action, the tooling resolves and reports the immutable SHA pin — reporting via `-Remediate` (now correct for subpath actions), applying via `-Apply`.
- No behavior change to the failing / soft-fail gate unless explicitly enabled — `-Apply` is opt-in and CI does not pass it.


scripts/security/Test-DependencyPinning.ps1
```diff
@@ -1615,6 +1632,83 @@ function Get-RemediationSuggestion {
     return "Manually research and pin to immutable reference"
 }
 
+function Invoke-ActionShaPinApply {
```
katriendg, line 1637: 💡 Optional (info): `-Apply` mutates files, so idiomatic PowerShell would gate it behind `[CmdletBinding(SupportsShouldProcess)]` + `$PSCmdlet.ShouldProcess(...)` to enable `-WhatIf`/`-Confirm`. Flagging as optional only — no other script in this repo uses `ShouldProcess` and PSScriptAnalyzer doesn't flag the `Invoke-*` verb, so adopting it here would introduce a new pattern rather than follow one. Fine to defer.

scripts/security/Test-DependencyPinning.ps1
```diff
@@ -1599,6 +1615,7 @@ function Get-RemediationSuggestion {
                 $sha = $response.sha
 
                 if ($sha) {
+                    $Violation.Metadata['ResolvedSha'] = $sha
```
katriendg, line 1620: 💡 Minor (non-blocking): consider validating the resolved SHA before stashing/writing it. `-Apply` writes `$response.sha` verbatim into workflow files, so a malformed or unexpected API response would be persisted as-is. A cheap 40-hex guard adds defense-in-depth and guarantees only a valid commit SHA is ever written:

```powershell
if ($sha -match '^[0-9a-f]{40}$') {
    $Violation.Metadata['ResolvedSha'] = $sha
    return "Pin to SHA: uses: $name@$sha # $version"
}
```

scripts/security/Test-DependencyPinning.ps1
```diff
@@ -1885,16 +1979,23 @@ try {
...
             $allViolations += $violations
         }
 
         Write-PinningLog "Found $(@($allViolations).Count) dependency pinning violations" -Level Info
 
+        # Auto-apply resolved SHA pins to GitHub Actions references when requested.
+        # This does not alter the compliance report or the exit-code gate below.
+        if ($Apply) {
+            $appliedCount = Invoke-ActionShaPinApply -Violations $allViolations -BasePath $Path
+            Write-PinningLog "Applied $appliedCount GitHub Actions SHA pin(s)" -Level Info
```
katriendg: 💡 Minor (UX, non-blocking): in a local `-Apply` run the compliance report is built from the pre-apply violations, so it still lists the just-fixed actions as non-compliant. That's correct for gate hermeticity, but can confuse a user who just fixed the pins. Consider making the log line explicit that a re-run verifies:

```powershell
Write-PinningLog "Applied $appliedCount GitHub Actions SHA pin(s); re-run without -Apply to verify compliance" -Level Info
```

scripts/tests/security/Test-DependencyPinning.Tests.ps1
```diff
@@ -1804,6 +1804,182 @@ Describe 'Get-RemediationSuggestion' -Tag 'Unit' {
...
+        Mock Invoke-RestMethod {
+            return @{ sha = 'abc123def456abc123def456abc123def456abcd' }
+        }
+
+        $null = Get-RemediationSuggestion -Violation $violation -Remediate
+        $violation.Metadata['ResolvedSha'] | Should -Be 'abc123def456abc123def456abc123def456abcd'
+    }
+}
+
+Describe 'Invoke-ActionShaPinApply' -Tag 'Unit' {
```
katriendg, line 1853: 💡 Minor (coverage, non-blocking): the out-of-range line guard in `Invoke-ActionShaPinApply` (`$partIndex -ge $parts.Count` → `Write-Warning` skip) isn't exercised, though the PR description lists "out-of-range... skips" as covered. A quick case closes the gap:

```powershell
It 'Skips violations whose line is past the end of the file' {
    $workflow = Join-Path $TestDrive 'short.yml'
    [System.IO.File]::WriteAllText($workflow, "steps:`n  - uses: actions/checkout@v4`n")

    $applied = Invoke-ActionShaPinApply -Violations @(
        New-ActionViolation -Name 'actions/checkout' -Version 'v4' -Line 999 -Sha $script:Sha -File 'short.yml'
    ) -BasePath $TestDrive

    $applied | Should -Be 0
}
```

katriendg (approved):
Thanks for this one, really valuable. Nothing blocking just a few suggestions, but approving as-is if you want to skip them.

scripts/security/Test-DependencyPinning.ps1
algattik: Deferred. No script in this repo uses `SupportsShouldProcess`, and PSScriptAnalyzer doesn't flag the `Invoke-*` verb here, so adopting `-WhatIf`/`-Confirm` for this single call site would introduce a new pattern rather than follow an existing one — declining under YAGNI. Worth revisiting if `ShouldProcess` is adopted repo-wide.

scripts/security/Test-DependencyPinning.ps1
algattik: Done in 3c7f9016. The resolved SHA is now validated with the shared `Test-SHAPinning -Type 'github-actions'` (the same `^[a-fA-F0-9]{40}$` guard used elsewhere) before it is stashed in `Metadata['ResolvedSha']` or written by `-Apply`; a malformed or unexpected API response falls through to the manual-research suggestion. Reused the existing validator instead of a fresh inline regex so the commit-SHA rule stays in one place.

scripts/security/Test-DependencyPinning.ps1
algattik: Done in 3c7f9016. The `-Apply` summary now reads `Applied N GitHub Actions SHA pin(s); re-run without -Apply to verify compliance`, so the pre-apply compliance figures aren't mistaken for the post-apply state.

scripts/tests/security/Test-DependencyPinning.Tests.ps1
algattik: Done in 3c7f9016. Added a Pester case ("Skips violations whose line is past the end of the file") that applies a `Line 999` violation against a two-line file, asserting the apply count is 0 and the file is left byte-for-byte unchanged — exercising the `$partIndex -ge $parts.Count` skip path.

rezatnoMsirhC (approved):

PR #1139, opened 2026-07-07
perf(training): eliminate PyTorch reinstall storm and inline scripts in OSMO GR00T fine-tuning

### LINKED ISSUE #1136: perf(training): eliminate PyTorch reinstall storm and inline scripts in OSMO GR00T fine-tuning
The OSMO GR00T VLA fine-tuning workflow provisions its Python/CUDA environment at container start on the stock base image (no prebuilt image). Two problems make that bootstrap slow and hard to maintain.

### 1. PyTorch and CUDA are reinstalled several times

`Isaac-GR00T`'s packaging extras pin an older PyTorch (`torch==2.5.1` for N1.5) than the flash-attn build and runtime need (`2.7.1`). Installing the editable gr00t package therefore downgrades the CUDA stack that flash-attn was just built against, and a later step reinstalls the newer stack to undo it. In the last successful run before this work, PyTorch churned through three versions (base image build → `2.5.1` → `2.7.1`), the multi-hundred-MB CUDA wheels were re-downloaded roughly three times, and flash-attn and gr00t were built/installed twice. Environment setup consumed roughly 11 of ~14 minutes, almost all of it redundant.

### 2. The workflow's entry logic is embedded inline in the YAML

The container bootstrap, dataset download, and Azure ML model-registration logic live as large inline script blocks inside `groot-train.yaml`. Inline scripts are not covered by the repo's shellcheck/ruff gates, are awkward to review, and drift from repo conventions.

## Impact

- Every fine-tuning submission pays a multi-minute penalty and repeated large wheel downloads for no functional benefit.
- The most complex, failure-prone part of the workflow — the environment bootstrap — is invisible to linting and hard to maintain.

## Desired outcome

- The PyTorch/CUDA/flash-attn stack installs exactly once; installing gr00t resolves against that stack instead of downgrading it. The fix stays in the runtime bootstrap — no prebuilt image is introduced.
- The workflow's entry logic lives in standalone, lint-clean repo files; the workflow YAML carries configuration only, with no embedded scripts.

## Acceptance criteria

- [ ] A fine-tuning run installs the torch/torchvision/torchaudio/flash-attn stack once, with no torch downgrade/reinstall cycle and no repeated CUDA wheel downloads.
- [ ] Container environment setup time is dominated by the single necessary install rather than reinstall churn.
- [ ] The workflow YAML contains no inline shell or Python scripts.
- [ ] The extracted scripts pass the repo's shellcheck and ruff gates.
- [ ] An end-to-end OSMO fine-tuning run completes with a trained checkpoint.

## Constraints

- No prebuilt or custom container image — the environment is provisioned at runtime on the stock base image.

algattik:
## Description

Eliminates the PyTorch reinstall storm in the OSMO GR00T fine-tuning container, extracts the formerly-inline workflow scripts into first-class reviewable files, and hardens the Isaac-GR00T checkout against mutable-ref drift.

Closes #1136

### Perf: eliminate the PyTorch reinstall storm

Previously the container installed the CUDA stack, then `pip install -e` of the cloned Isaac-GR00T package pulled gr00t's older `torch==2.5.1` pin — downgrading the stack `flash-attn` had just been built against, forcing a multi-hundred-MB wheel re-download and a second reinstall to undo it.

The install flow now:

- pins the torch stack once under a constraints file and installs it a single time;
- builds `flash-attn` against that stack once (`--no-build-isolation --no-deps`);
- rewrites gr00t's `torch`/`torchvision`/`torchaudio` pins to the installed versions before `pip install -e ".[base]"`, so the editable install is constrained and nothing churns.

Runtime GPU detection selects the CUDA 12.8 (`cu128`) stack on Blackwell (50xx / B100 / B200) and CUDA 12.6 otherwise. Training behavior is unchanged — this is an install-time optimization.

### Refactor: inline scripts → first-class files

The OSMO workflow embedded ~550 lines of shell/Python inline in `groot-train.yaml`. Those move to reviewable, lint-able files — `training/vla/scripts/groot/osmo-train-entry.sh` (container entry: env setup, dependency install, fine-tune, optional MLflow mirror) and `download_blob.py` (Azure Blob dataset downloader) — injected into the container via base64, the same mechanism already used for config injection. `groot-train.yaml` shrinks by ~550 lines.

The extracted entry script uses the shared `training/utils/aml_mirror.py` for MLflow mirroring. `aml_mirror.py` is extended so it also serves the GR00T path: logging selected training parameters, a configurable `source` tag, and an `AZURE_RESOURCE_GROUP` ← `AZUREML_RESOURCE_GROUP` fallback.

### Security: enforce an immutable GR00T ref

`ISAAC_GROOT_REF` is validated as a 40-hex commit SHA before checkout, so a mutable tag or branch cannot silently pull new upstream GR00T code between runs (mirrors the existing base-model-revision guard). The shipped `--vla-version` defaults already resolve to pinned SHAs, so no supported path regresses.

A comment records that the hand-pinned CUDA stack is intentionally outside automated dependency management (Dependabot / uv-lock / OSV): versions are chosen at runtime per GPU, `flash-attn` is source-built against the chosen torch, and gr00t's pins are rewritten to match — so they must be bumped together, not piecemeal, or the ABI breaks.

## Type of Change

- [x] ♻️ Refactoring (no change to training behavior; install-time perf optimization plus input validation)

## Component(s) Affected

## Testing Performed

- `shellcheck` clean on the changed shell — no new errors or warnings (only pre-existing `info`-level notices remain).
- Dependency-pinning `shell-inline-pip` gate: every `pip install` is exact-pinned (`==`).
- e2e path traced: `submit_osmo_vla_finetune` passes `--vla-version` (default `1.5`), which resolves `isaac_groot_ref` to an immutable SHA, so the new guard is transparent to `test_e2e_osmo_vla_finetune.py`.
- The GPU fine-tune e2e was not run locally (requires GPU / OSMO).

## Documentation Impact

- [x] Documentation updated in this PR (`training/vla/README.md` directory map + in-script comments)

## Checklist


training/vla/scripts/groot/download_blob.py
```diff
@@ -0,0 +1,90 @@
...
+    sas_client = ContainerClient(endpoint, container, credential=sas_token)
+    count = download_with_client(sas_client)
+    print(f"Downloaded {count} files to {DATASET_ROOT}/ using SAS token")
+    raise SystemExit(0)
+
+cred_client = None
+try:
+    cred = DefaultAzureCredential()
+    cred_client = ContainerClient(endpoint, container, credential=cred)
+except Exception:
```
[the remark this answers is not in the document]
algattik: Addressed in 3e1d60f1: the `except` is no longer empty. Constructing `DefaultAzureCredential` can fail when no managed identity is present in the container; that is an expected condition, so we intentionally fall through to the anonymous-access path below rather than abort. The handler now logs a diagnostic (`DefaultAzureCredential unavailable (...); will try anonymous access`) and carries an explanatory comment, so the fallback is visible in job logs instead of silently swallowed.

nguyena2 (approved):

training/vla/workflows/osmo/groot-train.yaml
```diff
@@ -25,8 +28,14 @@ workflow:
...
       image: "{{image}}"
-      command: ["bash"]
-      args: ["/tmp/train.sh"]
+      command: ["bash", "-euc"]
+      args:
+        - |
+          base64 -d /tmp/download_blob.py.b64 > /tmp/download_blob.py
+          base64 -d /tmp/aml_mirror.py.b64 > /tmp/aml_mirror.py
+          base64 -d /tmp/train.sh.b64 > /tmp/train.sh
+          [ -s /tmp/train.sh ] || { echo "ERROR: /tmp/train.sh is empty — submit via training/vla/scripts/submit-osmo-lerobot-vla-fine-tuning.sh" >&2; exit 1; }
```
katriendg: nit: checks for `/tmp/train.sh` but not the other two `.py` files also downloaded.

training/vla/scripts/groot/download_blob.py
```diff
@@ -0,0 +1,92 @@
...
+from urllib.parse import urlparse
+
+from azure.core.exceptions import HttpResponseError
+from azure.identity import DefaultAzureCredential
+from azure.storage.blob import ContainerClient
+
+url = os.environ["BLOB_URL"]
+parsed = urlparse(url)
+if parsed.scheme != "https" or not parsed.netloc:
+    raise SystemExit(f"Invalid BLOB_URL: {url}")
```
katriendg: ### File: training/vla/scripts/groot/download_blob.py

#### Comment 1 (Lines 14-19)

* Category: Security
* Severity: Low
* Status: Pre-existing (identical logic on `main`; carried over during the inline→file extraction)

**Potential SAS-token leak in error messages**

`BLOB_URL` can carry a SAS token in its query string (`?...&sig=...`). Two failure
paths print the full, unredacted URL to stdout, which lands in OSMO/AzureML job
logs:

```python
url = os.environ["BLOB_URL"]
parsed = urlparse(url)
if parsed.scheme != "https" or not parsed.netloc:
    raise SystemExit(f"Invalid BLOB_URL: {url}")            # L15 — leaks ?sig=...

path_parts = parsed.path.lstrip("/").split("/", 1)
if len(path_parts) < 2 or not path_parts[0] or not path_parts[1]:
    raise SystemExit(f"BLOB_URL must include account/container/prefix, got: {url}")  # L19 — leaks ?sig=...
```

Any operator or downstream log sink with read access to the job logs can then
recover a live SAS credential for the storage container. The check fires on
malformed input, but a wrong scheme (e.g. `http://…?sig=…`) or a missing
container segment on an otherwise-valid signed URL is exactly the case that
would expose the token.

**Suggested Resolution**

Strip the query (and thus the SAS signature) before echoing the URL in any error
message:

```python
def _redact(u: str) -> str:
    # Drop the query string so a SAS signature never reaches the logs.
    return urlparse(u)._replace(query="").geturl()

url = os.environ["BLOB_URL"]
parsed = urlparse(url)
if parsed.scheme != "https" or not parsed.netloc:
    raise SystemExit(f"Invalid BLOB_URL: {_redact(url)}")

path_parts = parsed.path.lstrip("/").split("/", 1)
if len(path_parts) < 2 or not path_parts[0] or not path_parts[1]:
    raise SystemExit(f"BLOB_URL must include account/container/prefix, got: {_redact(url)}")
```

**Follow-up Notes**

Not a blocker and not introduced by this PR, but the file is already being moved
and hardened here, so redacting the two messages is a cheap, in-scope close of a
credential-in-logs vector. No other call site prints the raw `BLOB_URL`.

katriendg (approved):
Really good one, thanks!
Two minor comments.

training/vla/workflows/osmo/groot-train.yaml
algattik: Good catch. Replaced the single `train.sh` check with a loop that asserts all three decoded files (`download_blob.py`, `aml_mirror.py`, `train.sh`) are non-empty before exec. The submit wrapper always injects all three, so the guard now covers the whole set. Fixed in 502c8911.

training/vla/scripts/groot/download_blob.py
algattik: Addressed in 502c8911. Added a `_redact()` helper that drops the query string via `urlparse(u)._replace(query="").geturl()`, and both `SystemExit` messages now emit `_redact(url)` so a SAS signature never reaches the logs. Verified the `http://…?sig=` case is redacted too.

PR #1157, opened 2026-07-08
test(evaluation): add Azure ML RL train->eval lifecycle e2e

### LINKED ISSUE #1156: End-to-end lifecycle test for the Azure ML RL (Isaac/SKRL) train → eval path
# End-to-end lifecycle test for the Azure ML RL (Isaac/SKRL) train → eval path

## Problem

The Azure ML RL (Isaac Lab / SKRL) training and software-in-the-loop (SIL) evaluation paths are validated only in isolation. Nothing exercises the full lifecycle — train a policy, register its checkpoint, then evaluate that registered model — so regressions that surface only when a freshly trained model is handed to the evaluator go undetected.

The SIL Isaac evaluation path additionally prepares its in-container Isaac runtime independently of training, so the two can drift, and its SKRL inference call passes arguments the agent API does not accept. Together these prevent the evaluation path from reliably running a trained policy end to end.

## Desired outcome

- An `e2e`-marked test that submits a real Azure ML RL training job, registers the checkpoint under a unique model name, resolves the concrete registered version (never `latest`), and evaluates that model through the SIL Isaac evaluation submission.
- The test asserts the training job's expected outputs (uploaded code snapshot, MLflow tracking, checkpoint) and that the evaluation job completes successfully.
- An opt-in shortcut to evaluate a pre-existing registered model without re-training, for a fast inner loop while iterating on the eval path.
- The Isaac SIL evaluation shares the training in-container runtime setup so the two stay consistent, and invokes the SKRL policy correctly.

## Out of scope

- The Isaac evaluation entrypoint emits JSON metrics only (no MLflow), so evaluation-leg MLflow assertions are not in scope; training-leg MLflow tracking is asserted.
- Live Azure ML e2e tests run under the `e2e` marker and are excluded from default PR CI.

algattik:
## Description

Adds end-to-end coverage of the Azure ML RL (Isaac Lab / SKRL) train → eval lifecycle: a single `e2e`-marked test submits a real training job, registers the checkpoint under a unique model name, resolves the concrete registered version (never `latest`), then evaluates that model through the SIL Isaac evaluation submission. An opt-in `E2E_AML_ISAAC_EVAL_MODEL` override evaluates a pre-existing model without re-training, for a fast inner loop.

To make that path runnable, the Isaac in-container runtime setup is extracted into shared scripts sourced by both training and evaluation (removing drift), the SKRL inference call in the evaluation loop is corrected, and `.env` files are excluded from Azure ML code snapshots.

Closes #1156

## Type of Change

- [ ] 📚 Documentation update

## Component(s) Affected

- [ ] `docs/` - Documentation

Also affected (no template box): `evaluation/sil/` (shared runtime + eval path) and `tests/e2e/`.

## Testing Performed

`ruff check` and `ruff format --check` clean; the `evaluation/` unit suite passes; `tests/e2e` collects without import errors; changed shell scripts pass `shellcheck`. The live lifecycle test runs under `-m e2e` against a real Azure ML workspace and is not part of default PR CI.

## Documentation Impact

## Bug Fix Checklist

- [x] Regression test included, OR

## Checklist


tests/e2e/test_e2e_aml_rl_lifecycle.py
```diff
@@ -0,0 +1,124 @@
...
+        log_e2e("Validating AzureML training MLflow tracking")
+        assert_aml_job_has_mlflow_tracking(job, aml_workspace)
+        log_e2e("Validating AzureML checkpoint output")
+        assert_job_has_checkpoint(job)
+        model = resolve_registered_model(repo_root, aml_workspace, model_name=register_model_name)
+        request.addfinalizer(lambda: archive_all_model_versions(repo_root, aml_workspace, register_model_name))
+    else:
+        log_e2e(f"Using pre-registered eval model {model.name}:{model.version} (training skipped)")
+
+    eval_job = submit_aml_isaaclab_eval(
```
bindsi, line 111: nit: the eval submission doesn't pass `--task`, so it relies on `submit-azureml-isaaclab-evaluation.sh`'s default task happening to equal `_TASK`. They align today, but if `_TASK` is ever changed the training and eval tasks would silently desync. Consider passing `task=_TASK` through to the eval submission (or driving it from the registered model's metadata) so the two stay coupled.

evaluation/sil/evaluation.sh
```diff
@@ -2,52 +2,22 @@
...
-  python_cmd=(python)
-fi
-
-export PYTHONPATH="${SRC_DIR}:${PYTHONPATH:-}"
-
-runtime_project="${SRC_DIR}/training/il/lerobot"
-
-if [[ ! -f "${runtime_project}/uv.lock" ]]; then
-  echo "Error: LeRobot lockfile not found at ${runtime_project}/uv.lock" >&2
+ROOT_DIR="$(cd "${SCRIPT_DIR}/../.." && pwd)"
```
bindsi, line 5: nit: this refactor drops the local `.env` sourcing that `evaluation.sh` previously did, while the sibling `infer.sh` still auto-loads `.env`. Harmless for the AML path (config comes from job env-vars/inputs, and `.env` is now excluded from snapshots), but it's a small inconsistency for local runs. Either intentional-and-fine, or worth aligning the two scripts.

training/rl/scripts/setup_isaac_runtime.sh
```diff
@@ -0,0 +1,68 @@
...
+if command -v uv &>/dev/null; then
+  echo "uv detected, exporting locked manifest dependencies from ${ISAAC_PROJECT_DIR}..."
+  # The Isaac Lab container ships a torch + CUDA runtime stack matched to its GPU
+  # driver and to the omni.*/pip_prebundle trees the simulator imports. torch is only
+  # a transitive dependency of skrl/rsl-rl, so installing the lock's torch would
+  # replace the container's and split-load the package across the prebundle and
+  # site-packages, breaking skrl. Therefore, torch/CUDA/nvidia-cu* are excluded via
+  # regex and every other locked package is installed with --no-deps. pynvml/nvidia-ml-py
+  # are pure-Python NVML bindings and are intentionally retained.
+  isaac_provided_re='^(torch|torchvision|triton|cuda-bindings|cuda-pathfinder|cuda-toolkit|nvidia-(cu|nccl|nvjitlink|nvshmem|nvtx)[a-z0-9.-]*)=='
```
bindsi, line 55: Non-blocking question (not a nit): this shared script changes the **training** install behavior, not just eval. The previous `train.sh` piped the full locked export into `uv pip install --no-deps`, so it installed the lock's `torch`/CUDA/`nvidia-cu*` wheels too; this version `grep -Ev`-excludes them. The rationale in the comment above is sound (preserve the container's matched torch stack, avoid split-loading that breaks skrl), but since it alters a previously-working path, can you confirm the lifecycle e2e was run with training enabled (i.e. `E2E_AML_ISAAC_EVAL_MODEL` unset) so a real train → register → eval was exercised end-to-end with this exclusion in place?

evaluation/tests/test_policy_evaluation.py
```diff
@@ -342,6 +342,25 @@ def test_skrl_path_collects_metrics(self) -> None:
         assert metrics.lengths == [3, 3]
         agent.act.assert_called()
 
+    def test_skrl_act_arguments(self) -> None:
```
bindsi, line 345: nit: since `agent` is a `MagicMock`, this test asserts the *call shape* rather than real skrl compatibility — valuable as a regression guard against the old `inference=` kwarg creeping back, but it wouldn't catch a future skrl signature change. Optional: a one-line comment noting the asserted shape mirrors skrl 2.1.0's `act(observations, states, *, timestep, timesteps)` would make the intent clear to future readers.

evaluation/sil/scripts/submit-azureml-isaaclab-evaluation.sh
```diff
@@ -217,6 +221,22 @@ model_json=$(az ml model show \
...
   "$resource_group" "$workspace_name" "$subscription_id"
 
+#------------------------------------------------------------------------------
+# Stage Code Snapshot
+#------------------------------------------------------------------------------
+# policy_evaluation.py loads Isaac Lab policies (skrl/rsl-rl) and imports both
+# training.rl.simulation_shutdown and training.utils.integrity, so the job needs the
+# training/rl runtime and training/utils packages next to evaluation/sil. Azure ML's
+# single code root cannot span multiple top-level trees, so stage them into one directory.
+code_path="$(mktemp -d)"
```
bindsi, line 232: nit: with this staging change the file header at the top (lines 1-3) is now stale — it still says *"using evaluation/ as the code directory"* and *"The .amlignore file controls which files are excluded from the code snapshot"*, but the code root is now this staged `mktemp` dir. Worth updating the header to match.

bindsi (commented):
Reviewed the AML RL train→eval lifecycle work. Overall this is a solid, well-scoped change: the SKRL `act()` fix is correct against skrl 2.1.0's `act(observations, states, *, timestep, timesteps)` signature (the old `inference=None` kwarg was invalid and omitted the required `states` positional), the shared-runtime refactor sensibly removes training/eval drift, and excluding `.env` from AML snapshots is good hardening. No blocking issues.

Left one non-blocking question on the training-install behavior change (torch/CUDA now excluded on the training path too) plus a few nits. Approving in spirit — just the one confirmation on F2 (fresh-training run exercised) would be reassuring before merge.

evaluation/.amlignore
```diff
@@ -14,6 +14,11 @@ venv/
 env/
 .env/
 
+# Secrets
```
katriendg, line 17: nice one

tests/e2e/test_e2e_aml_rl_lifecycle.py
katriendg: +1, and it's actually a bit stronger than a silent desync risk. `submit-azureml-isaaclab-evaluation.sh` hardcodes `task="${TASK:-Isaac-Velocity-Rough-Anymal-C-v0}"` at line 84, so `task` is never empty by the time the model-metadata fallback runs — `[[ -z "$task" ]] && task=$(... jq .tags.task ...)` at line 209 is effectively **dead code whenever `--task` is omitted**. So today the eval doesn't just "happen to" match `_TASK`; it will *always* use the hardcoded default and never the registered model's `tags.task`, even if the model was trained on a different task. Passing `task=_TASK` through from the lifecycle test (and/or preferring the model-metadata resolution over the hardcoded default) would fix both the test coupling and make the standalone script honor the model it's actually evaluating.

Still not a blocker for the PR, just a nice to address.

katriendg (approved):

tests/e2e/test_e2e_aml_rl_lifecycle.py
algattik: Addressed in c7bb00e7 with two changes:

- The lifecycle test now passes `task=_TASK` to `submit_aml_isaaclab_eval`, so training and eval use the same task and can no longer silently desync.
- `submit-azureml-isaaclab-evaluation.sh` drops its hardcoded `task="${TASK:-Isaac-Velocity-Rough-Anymal-C-v0}"` default (now `task="${TASK:-}"`). With `--task` omitted, the `[[ -z "$task" ]] && task=$(... jq .tags.task ...)` model-metadata resolution is now live instead of dead, so the standalone script honours the task the model was actually trained on (falling back to the `auto` sentinel only when the model carries no `tags.task`).

evaluation/sil/evaluation.sh
algattik: Fixed in c7bb00e7 — restored the local `.env` sourcing, reading `evaluation/.env` to match `infer.sh` (and the pre-refactor `evaluation.sh`). Still a no-op on the AML path, but local runs are consistent again.

training/rl/scripts/setup_isaac_runtime.sh
algattik: The exclusion is deliberate and the rationale in the comment holds: the regex keeps the container's matched stack (`torch`, `torchvision`, `triton`, `cuda-*`, and `nvidia-(cu|nccl|nvjitlink|nvshmem|nvtx)*`) and installs every other locked package with `--no-deps`, avoiding the split-load across the pip_prebundle and site-packages that breaks skrl.

A live train-enabled lifecycle run (`E2E_AML_ISAAC_EVAL_MODEL` unset, exercising a real train → register → eval with this exclusion in place) was run manually, of course, to verify the changes.

evaluation/tests/test_policy_evaluation.py
algattik: Done in c7bb00e7 — added a one-line comment noting the asserted call shape mirrors skrl 2.1.0's `act(observations, states, *, timestep, timesteps)` signature, so the intent is clear to future readers.

evaluation/sil/scripts/submit-azureml-isaaclab-evaluation.sh
algattik: Fixed in c7bb00e7 — the header now describes the staged snapshot (the job stages `training/rl` and `training/utils` alongside `evaluation/sil` into a temp code root, and the staged `.amlignore` controls exclusions), replacing the stale "using evaluation/ as the code directory" wording.

bindsi (approved):

PR #392, opened 2026-04-05
feat(build): enforce strict warnings across all linters

WilliamBerryiii:
This PR established a **warnings-as-errors policy** across all CI linters and fixed a critical bug where Pester test failures were silently masked in CI. ShellCheck was added as a new linter with full CI integration, YamlLint was promoted to strict mode, and all remaining soft-fail linter jobs were switched to hard-fail.

Closes #6

### Critical Bug Fix

The Pester test workflow contained a `$global:LASTEXITCODE = 0` reset that silently swallowed all test failures, allowing broken builds to pass CI. This was removed and replaced with explicit `exit 1` calls when Pester reports failures or throws exceptions. Two test files also had stale dot-source paths referencing the `scripts/lib/` symlink target instead of the canonical `shared/lib/` location — these were corrected.

### ShellCheck Integration

A complete **ShellCheck linting pipeline** was added to the repository:

- Added *.shellcheckrc* at the repository root with `shell=bash`, `severity=warning`, and `external-sources=true`
- Added a **reusable GitHub Actions workflow** (*shellcheck.yml*) that installs ShellCheck, runs the wrapper, and uploads results as a CI artifact
- Added *Invoke-ShellCheck.ps1* (218 lines), a PowerShell wrapper that discovers `.sh` files, runs ShellCheck with JSON output, classifies findings, writes structured results to `logs/`, and emits CI annotations — treating **warnings as failures**
- Added *Invoke-ShellCheck.Tests.ps1* (334 lines) with 10 Pester test contexts covering tool availability, changed-files-only mode, error/warning classification, JSON export structure, and directory exclusion
- Integrated ShellCheck into both *main.yml* and *pr-validation.yml* workflows with `soft-fail: false`

### YamlLint Strict Mode

*Invoke-YamlLint.ps1* was updated to treat warnings as failures by broadening the failure condition to include `$warningCount -gt 0`. The corresponding Pester tests in *Invoke-YamlLint.Tests.ps1* were updated to reflect strict mode expectations.

### Soft-Fail Policy

**Linting jobs** enforce strict failures (`soft-fail: false`):

| Job | `main.yml` | `pr-validation.yml` |
|-----|-----------|---------------------|
| Terraform Validation | `soft-fail: false` | `soft-fail: false` |
| Go Lint | `soft-fail: false` | `soft-fail: false` |
| ShellCheck | `soft-fail: false` | `soft-fail: false` |

[...]

PR #1234, opened 2026-07-21
# feat(data): add episode analysis workflows to dataviewer

akzaidi:
# feat(data): add episode analysis workflows to dataviewer

**PR Soundrack**: [Juicy J & Project Pat - Dem Goats](https://www.youtube.com/watch?v=ZA4HVAjzy9o&list=PLZvY2YfsLBV4&index=2)

## Description

Added an Episode Analyzer workflow that combines trajectory metrics, VLM outcomes, persisted analysis, filterable labels, and language instructions in the dataviewer. The backend now stores structured per-episode analysis and supports importing selected analysis fields as labels, while the standalone labeling CLI processes LeRobot datasets with Qwen3-VL.

The VLM judge request flow now returns completed judgments synchronously from a worker thread instead of returning `202 Accepted` and requiring client polling. Dataset instruction lookup also supports LeRobot v3 task metadata.

Related issues: None

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [ ] `docs/` - Documentation
- [x] `data-management/viewer/` - Dataset analysis and annotation
- [x] `evaluation/vlm_judge/` - VLM judge service contract

## Testing Performed

- [ ] Dataviewer validation passed (`npm run validate`)
- [x] Frontend validation passed (`npm run validate` from `data-management/viewer/frontend`)

## Documentation Impact

## Bug Fix Checklist

## Checklist

### Analysis and labeling

* Persisted structured episode analysis alongside dataset labels and exposed read, write, and label-import endpoints.
* Added motion analysis with selectable log-scaled and radian-based smoothness normalization.
* Added a Qwen3-VL labeling CLI with multi-view sampling, structured output parsing, and resumable JSONL results.

### Dataviewer interface

* Added the **Episode Analyzer** tab with motion metrics, VLM judging, analysis results, labels, and language instructions.
* Added collapsible analysis sections and retained their open state across episode navigation.
* Added label search and dataset-wide import controls for supported analysis fields.

### Compatibility and tests

* Added LeRobot v3 task metadata support for task strings stored in episode metadata and Parquet columns.
* Changed VLM judge requests from asynchronous polling to synchronous completed responses.
* Added backend and frontend behavior coverage, including a regression test for duplicate trajectory tooltip labels.


data-management/viewer/backend/scripts/vlm_label_dataset.py
```diff
@@ -0,0 +1,349 @@
...
+    model_id: str,
+    device_map: str,
+    dtype: str,
+    limit: int | None,
+    scene_context: str | None = None,
+) -> dict[str, Any]:
+    """Label every (or ``limit``) episode and write JSONL + CSV to ``output_dir``."""
+    selected_views = resolve_views(dataset_root, views)
+    output_dir.mkdir(parents=True, exist_ok=True)
+    jsonl_path = output_dir / "labels.jsonl"
```
katriendg, line 224: **Clarity / completeness (High).** This script writes `labels.jsonl` / `labels.csv`, but the Episode Analyzer consumes `meta/episode_labels.json` (the `analysis` map read by `useDatasetLabels`). Nothing in this PR bridges the two, so a user who runs the script cannot get results into the viewer without hand-writing JSON or scripting `PUT /episodes/{idx}/analysis`. This flow is unclear for users — please **document this script** (what it emits and how those results reach the viewer), and ideally add a converter (or a `--write-analysis` mode that emits/merges `episode_labels.json`). `label_dataset()` itself is also untested — a fake-backend test would cover row-building, error rows, and JSONL/CSV writing without a GPU.

data-management/viewer/backend/scripts/vlm_label_dataset.py
```diff
@@ -0,0 +1,349 @@
...
+
+    _LOGGER.info("Loading %s ...", model_id)
+    backend = Qwen3VLBackend(model_id=model_id, device_map=device_map, dtype=dtype)
+    gen_cfg = GenerationConfig(max_new_tokens=512, temperature=0.0)
+
+    episodes = list(iter_episodes(dataset_root, views=selected_views, limit=limit))
+    _LOGGER.info("Labeling %d episodes from %s (views: %s)", len(episodes), dataset_root.name, list(selected_views))
+
+    rows: list[dict[str, Any]] = []
+    with jsonl_path.open("w") as jf:
```
katriendg, line 235: **Correctness / intent mismatch (Medium).** The PR description advertises "resumable JSONL results", but `jsonl_path.open("w")` truncates the output and `label_dataset` always iterates every episode from scratch — there is no skip-existing/resume logic. On an expensive GPU labeling pass, an interrupted run restarts from zero. Either implement resume (append + skip already-labeled `episode_index`) or drop the "resumable" claim.

data-management/viewer/backend/src/api/services/trajectory_analysis.py
```diff
@@ -146,6 +164,31 @@ def _compute_smoothness(self, jerk: NDArray[np.float64]) -> float:
         smoothness = 1.0 / (1.0 + rms_jerk)
         return float(np.clip(smoothness, 0.0, 1.0))
 
+    def _compute_normalized_smoothness(self, jerk: NDArray[np.float64]) -> float:
```
katriendg, line 167: **Test coverage (Medium) — known gap, noting inline.** `_compute_normalized_smoothness` and the `smoothness_mode` validation (line 80) are new and untested — `test_ai_analysis.py` and `test_property_based.py` reference none of `normalized_smoothness` / `smoothness_mode` / `radian-based`. Add unit tests for: `log-scaled` vs `radian-based` output, the `ValueError` on an invalid mode, and that it discriminates on degree-scale trajectories where raw `smoothness` saturates near zero.

data-management/viewer/backend/src/api/routes/ai_analysis.py
```diff
@@ -172,7 +179,7 @@ async def analyze_trajectory(data: TrajectoryData) -> TrajectoryMetricsResponse:
     timestamps = np.array(data.timestamps)
     gripper_states = np.array(data.gripper_states) if data.gripper_states else None
 
-    analyzer = TrajectoryAnalyzer()
+    analyzer = TrajectoryAnalyzer(smoothness_mode=data.smoothness_mode)
```
katriendg, line 182: **Test coverage (Low) — known gap, noting inline.** The new `smoothness_mode` request field and `normalized_smoothness` response field have no route-level test. Add a `POST /api/ai/trajectory-analysis` test asserting the mode passes through to the analyzer and `normalized_smoothness` appears in the response for both modes.

data-management/viewer/frontend/src/hooks/use-labels.ts
```diff
@@ -153,6 +198,32 @@ export function useRemoveLabelOption() {
...
+  const queryClient = useQueryClient()
+
+  const mutation = useMutation({
+    mutationFn: ({ field, overwrite }: { field: ImportableAnalysisField; overwrite?: boolean }) => {
+      if (!currentDataset) throw new Error('No dataset selected')
+      return importAnalysisLabels(currentDataset.id, field, { overwrite })
+    },
+    onSuccess: (data) => {
+      setAvailableLabels(data.available_labels)
+      setAllEpisodeLabels(data.episodes)
```
katriendg, line 217: **Edge-case / data loss (Medium).** `setAllEpisodeLabels(data.episodes)` also resets `savedEpisodeLabels` for **all** episodes to the server response. Any unsaved in-progress label edits on other episodes are silently discarded when a user imports. Consider merging only the affected episodes, or invalidating the query and letting the normal load path reconcile. `importAnalysisLabels` / `useImportAnalysisLabels` also lack tests (success updates the store, non-ok throws).

data-management/viewer/frontend/src/components/annotation-panel/LabelPanel.tsx
```diff
@@ -91,6 +120,24 @@ export function LabelPanel({ episodeIndex }: LabelPanelProps) {
     }
   }, [pendingDeleteLabel, removeOption])
 
+  const handleImport = useCallback(
+    async (field: ImportableAnalysisField) => {
+      try {
+        const result = await importLabels.mutateAsync({ field })
```
katriendg, line 126: **Incomplete capability (Low).** `handleImport` only sends `{ field }`, so the backend's `overwrite` and `prefix` options (implemented in the endpoint and in `useImportAnalysisLabels`) are unreachable from the UI. Users can't refresh/overwrite stale imported labels. Either expose an "overwrite" affordance or drop the unused parameters until the UI needs them.

data-management/viewer/frontend/src/components/annotation-panel/LabelPanel.tsx
```diff
@@ -172,6 +219,29 @@ export function LabelPanel({ episodeIndex }: LabelPanelProps) {
         <div className="text-muted-foreground text-xs">Applied: {currentLabels.join(', ')}</div>
       )}
 
+      {/* Import analysis fields as labels */}
+      {importableFields.length > 0 && (
```
katriendg, line 223: **Test coverage (Medium) — known gap, noting inline.** This file is the lowest-covered patch file (41%). The import controls, `importableFields` derivation, `handleImport` success/error messaging, and the delete-confirm dialog are untested. Add `LabelPanel` tests for: importable-field detection from `episodeAnalysis`, a successful import message, and the import error branch.

data-management/viewer/backend/src/api/routers/vlm_judge.py
```diff
@@ -256,108 +217,35 @@ async def run_episode_judgment(
...
-            _run_judgment_job,
-            job=job,
-            judge_service=judge_service,
-            dataset_id=dataset_id,
-            episode_idx=episode_idx,
+    try:
+        # Model inference is blocking and GPU-bound; run it in a worker thread so
+        # the single event loop stays free to serve episode/video requests while
+        # a judgment is in flight (otherwise the whole backend stalls per run).
+        result = await run_in_threadpool(
```
katriendg, line 226: **Cleanup / stale contract (Medium).** Moving the judge to a synchronous `run_in_threadpool` response is a good change, and the frontend happy path still works (the new `JudgeResponse.episode_id` is handled by the first branch of `runVlmJudge`). But this removes the `202` / `job_status` / `error` polling contract while the frontend keeps the now-dead scaffolding: the poll loop and `VLM_JUDGE_POLL_*` in `api-client.ts`, the `refetchInterval` on `jobStatus` in `use-vlm-judge.ts`, `jobStatus` in `types/vlm-judge.ts`, and `lib/__tests__/api-client.test.ts` still asserts the 202 polling path (green but exercising a branch the backend can no longer produce). Recommend removing the dead client path in this PR so the test suite reflects the real contract.

katriendg (commented):
Nice feature set — the Episode Analyzer flow, persisted per-episode analysis, and the switch to a synchronous worker-thread judge all read well, and the `import-from-analysis` endpoint is genuinely well-tested. Leaving this as a **comment** (non-blocking); the inline notes are suggestions plus a few known gaps.

**Please add user-facing docs.** The PR is marked "no documentation changes needed," but several of these flows are unclear for end users and there's nothing in `docs/`, the dataviewer README, or a skill covering them. A short user guide would go a long way:
- The Qwen3-VL labeling **script** (`vlm_label_dataset.py`): document that it's a script, how to run it, what it emits (`labels.jsonl` / `labels.csv`), and — most importantly — how those results actually reach the viewer. Right now nothing bridges the script output to the `meta/episode_labels.json` the Episode Analyzer reads, so it's not obvious how a user closes that loop.
- The Episode Analyzer tab end-to-end: importing analysis fields as labels, and what each motion metric (including the new `log-scaled` vs `radian-based` smoothness modes) means.

**Missing tests (known — captured inline, not blocking).** The lower-coverage areas are called out on the relevant lines: `_compute_normalized_smoothness` + `smoothness_mode` validation, the `smoothness_mode` route field, `LabelPanel` (lowest patch coverage at 41%), and `useImportAnalysisLabels`. Keeping these as inline notes so they're easy to pick up.

**A couple of small non-blocking items** are inline too: the "resumable JSONL" claim doesn't match `open("w")` truncation; `setAllEpisodeLabels` resets saved labels for all episodes on import; the import UI can't reach the backend's `overwrite`/`prefix`; and the frontend still carries dead 202-polling scaffolding after the judge went synchronous.

One pre-existing observation (not introduced here, worth a follow-up issue): `BlobLabelStorage.load` uses `except (json.JSONDecodeError, Exception)` and reaches into provider privates — it can silently return empty labels that a later save persists over real data.

data-management/viewer/frontend/src/components/annotation-workspace/AnnotationWorkspaceAnalyzerTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+              Results are written to this episode's labels.
+            </p>
+          </div>
+          {motionMetricsPanel}
+          {judgePanel}
+        </div>
+        <Card
+          data-testid="analyzer-results-panel"
+          className="order-2 min-h-[280px] overflow-hidden lg:min-h-0"
+        >
```
rezatnoMsirhC, line 46: **Correctness / UX (High).** The description here says "Results are written to this episode's labels", but neither `MotionMetricsPanel` nor the VLM judge actually calls `PUT /api/datasets/{id}/episodes/{idx}/analysis` from the UI. `MotionMetricsPanel` posts to `POST /api/ai/trajectory-analysis` and renders the result in memory only; the VLM judge writes to the disk cache only. `EpisodeAnalysisCard` reads exclusively from `useLabelStore.episodeAnalysis`, which is loaded from `GET /labels` (i.e., `meta/episode_labels.json`). The only path that actually writes there today is the external `vlm_label_dataset.py` CLI.

Two ways to resolve this:

**Option A: Add a "Save analysis" button.** After `MotionMetricsPanel` receives metrics, call `PUT /episodes/{idx}/analysis` with the computed fields. Do the same in the VLM judge's `onSuccess` path. This makes the description accurate.

**Option B: Correct the copy now, add persistence later.** Change the description to something accurate for the current state, for example: "Score motion quality and judge task success. Saved analysis results appear in the Episode Analysis card once written to the dataset." This is low-risk and unblocks the PR without adding scope.

Either option prevents users from being confused when they run the motion analysis, see results in the panel, navigate to a different episode, come back, and find the card empty.

data-management/viewer/backend/src/api/services/lerobot_loader.py
```diff
@@ -875,11 +891,14 @@ def get_tasks(self) -> dict[int, str]:
...
+                if "task_index" in cols:
+                    # LeRobot v3 serializes the task string as the parquet index
+                    # (``__index_level_0__``); older exports use a ``task`` column.
+                    task_col = "task" if "task" in cols else next((c for c in cols if c != "task_index"), None)
+                    if task_col is not None:
+                        indices = table.column("task_index").to_pylist()
+                        strings = table.column(task_col).to_pylist()
+                        return {int(idx): str(s) for idx, s in zip(indices, strings)}
             except (OSError, pa.ArrowException) as e:
                 logger.warning("Failed to read %s: %s", tasks_parquet, e)
```
rezatnoMsirhC, line 904: **Correctness (Medium).** The original code required both `task_index` and `task` to be present. The fallback for v3 uses:

```python
task_col = "task" if "task" in cols else next((c for c in cols if c != "task_index"), None)
```

`next(c for c in cols if c != "task_index")` picks the first column that is not `task_index`, which in a real v3 Parquet file could be `chunk_index`, `__null_dask_index__`, or another data column, silently producing wrong task strings. The comment says v3 uses `__index_level_0__` as the task string column, so that name should be checked explicitly.

Suggested fix, inlined for one-time use:

```python
if "task_index" in cols:
    # Check known task-string column names in priority order; v3 uses the Parquet index level.
    task_col = next((c for c in ("task", "__index_level_0__", "task_description") if c in cols), None)
    if task_col is not None:
        indices = table.column("task_index").to_pylist()
        strings = table.column(task_col).to_pylist()
        return {int(idx): str(s) for idx, s in zip(indices, strings)}
```

This preserves priority order (v2 `task` column wins over v3 index level) and avoids accidentally using unrelated columns.

data-management/viewer/backend/tests/scripts/test_vlm_label_dataset.py
```diff
@@ -0,0 +1,138 @@
...
+    spec = importlib.util.spec_from_file_location("vlm_label_dataset", _SCRIPT_PATH)
+    assert spec and spec.loader
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+mod = _load_module()
+
+
```
rezatnoMsirhC, line 28: **Nit / Test reliability.** `mod = _load_module()` at module level runs `vlm_label_dataset.py`'s top-level imports (including `from evaluation.vlm_judge.backend import ...`) during pytest collection. If `evaluation` or its GPU dependencies are absent, the entire module fails to collect and every test here is reported as an error with a non-obvious traceback into the module loader. A session-scoped fixture confines any failure to the tests that actually need it:

```python
@pytest.fixture(scope="session")
def mod():
    spec = importlib.util.spec_from_file_location("vlm_label_dataset", _SCRIPT_PATH)
    assert spec and spec.loader
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    return module
```

Each test then takes `mod` as a parameter instead of referencing the module-level variable.

rezatnoMsirhC (commented):
PR Soundtrack Approved 🔥

data-management/viewer/frontend/src/components/annotation-workspace/AnnotationWorkspaceAnalyzerTab.tsx
akzaidi: Addressed via B in f0e5a371.

data-management/viewer/frontend/src/components/annotation-panel/LabelPanel.tsx
akzaidi: Addressed: tests cover importable-field detection, custom prefix/overwrite, success messaging, error messaging, and delete confirmation; patch coverage is now >85%

data-management/viewer/backend/src/api/routers/vlm_judge.py
akzaidi: Addressed via f0e5a3716e135e658df34f99928f99962bef03a3: removed polling constants, sleep/loop logic, jobStatus/error scaffolding, query refetch polling, and the obsolete polling test.

data-management/viewer/frontend/src/hooks/use-labels.ts
akzaidi: Addressed: imports now use dataset-scoped three-way reconciliation, preserve local edits, retain incoming imported labels, and ignore late responses after dataset switches

data-management/viewer/backend/src/api/routes/ai_analysis.py
akzaidi: Addressed: tests now cover log-scaled and radian-based calculations, invalid modes, and discrimination where raw degree-scale smoothness approaches zero

data-management/viewer/backend/scripts/vlm_label_dataset.py
akzaidi: Added --write-analysis to merge successful CLI results into meta/episode_labels.json while preserving existing labels and unrelated analysis fields. Documented CLI usage, JSONL/CSV outputs, viewer persistence, resume behavior, nested dataset IDs, and concurrent-writer constraints. 

Also added fake-backend tests covering successful and failed rows, output generation, resumability, interrupted-output recovery, and analysis-file merging

akzaidi (commented):
## Review result: changes required before merge

The branch is current with `main` at `6582c578`, and the post-sync validation workflow passes. The following functional issues remain:

### Blocking

1. **Partial analysis updates erase existing fields.** [`labels.py:505-516`](https://github.com/microsoft/physical-ai-toolchain/blob/6582c57803ed8b759542ccff9c112390c22a474b/data-management/viewer/backend/src/api/routers/labels.py#L505-L516) replaces the complete record even though every field is optional. Merge explicitly supplied fields or require a complete PUT representation.
2. **Concurrent label and analysis writes can lose data.** The endpoints use an unlocked load-modify-save cycle, while Blob storage writes with unconditional overwrite. Add an atomic storage mutation with per-dataset locking and Blob ETag retries.
3. **Saved per-episode judge instructions are ignored.** [`vlm_judge.py:147-195`](https://github.com/microsoft/physical-ai-toolchain/blob/6582c57803ed8b759542ccff9c112390c22a474b/data-management/viewer/backend/src/api/routers/vlm_judge.py#L147-L195) resolves only request or dataset metadata, breaking annotation-only batch judging and cache retrieval. Restore request → saved annotation → dataset precedence.
4. **Synchronous VLM inference can exceed the deployed request limit.** [`vlm_judge.py:176-248`](https://github.com/microsoft/physical-ai-toolchain/blob/6582c57803ed8b759542ccff9c112390c22a474b/data-management/viewer/backend/src/api/routers/vlm_judge.py#L176-L248) keeps the request open while inference may take minutes; Azure Container Apps ingress limits requests to 240 seconds. Return a durable asynchronous job and poll its status.
5. **Inference lacks single-flight deduplication and a GPU concurrency bound.** Concurrent misses can duplicate model execution, exhaust GPU memory, and race on the same cache temporary path. Add keyed coordination, a bounded executor or semaphore, and unique atomic cache temporaries.
6. **Malformed model output is persisted as complete.** [`vlm_label_dataset.py:128-205`](https://github.com/microsoft/physical-ai-toolchain/blob/6582c57803ed8b759542ccff9c112390c22a474b/data-management/viewer/backend/scripts/vlm_label_dataset.py#L128-L205) accepts any JSON object, fills defaults, and lets `--resume` skip it forever. Validate a strict response schema and retain schema failures as retryable errors.

### Follow-ups

- Make optional `.env` lookups in `start.sh` safe under `errexit` and `pipefail`.
- Add immutable model revision input and provenance to the labeling CLI.
- Return a nonzero CLI exit code when episode errors remain.
- Memoize trajectory-derived analysis inputs during playback.

Fresh CI has no failed or pending checks. CodeQL alert #392 is pre-existing on `main` from July 9 and is not attributable to this PR. GitHub does not permit the PR author account to submit a formal request-changes review, so these blockers are recorded as a comment review.

agreaves-ms (approved):

PR #1383, opened 2026-08-18
fix(training): pin AzureML environment templates to explicit versions

### LINKED ISSUE #1079: Pin AzureML environment assets to explicit versions
## Problem

AzureML workflows reference their environment assets as `azureml:<name>:latest`. The floating `:latest` tag resolves to whatever environment version was most recently registered, so a job submitted today and the same job submitted next week can run against different environments with no change in the repository. This is neither reproducible nor auditable.

This surface is distinct from OCI container images: `azureml:<name>:<version>` refs are registered AzureML assets, not registry images, so they carry an environment **version**, not an `@sha256:` digest. Digest-pinning of OCI `image:` references and its CI enforcement are tracked separately in #1082; this issue covers only the AzureML environment-asset half, which that effort does not address.

## Goal

Make every AzureML environment reference resolve to an explicit, intended version instead of the floating `:latest`, wherever the submission flow allows — while preserving the dynamic submit-time environment registration the pipeline flow depends on.

## Coverage and the two reference kinds

AzureML environment refs split into two kinds with different pinning behaviour:

- **Standalone command jobs** — the submit scripts register the environment and override the job's `environment` with the concrete version via `--set environment=azureml:<name>:<version>` at submit time, so `:latest` here is only a direct-submit fallback that the normal path never uses.
- **Pipeline components** — the `--set` override reaches only the top-level command job; it cannot reach the `environment:` refs nested inside pipeline **component** files (referenced by the LeRobot pipeline definitions). Those components rely on `:latest` to bind to the environment version registered in the same submit run. Pinning them statically would decouple them from the environment-version flag and break dynamic registration.

The covered references live in the AzureML training and evaluation workflow templates and their pipeline component definitions.

## Acceptance criteria

- Standalone-job environment references are pinned to explicit versions rather than `:latest`.
- Any remaining `:latest` is limited to pipeline components and is documented in place as intentional submit-time binding, so it is not mistaken for a gap.
- The direct-submit fallback stays consistent with the version the submit scripts register.
- CI is green.

algattik:
## Summary

AzureML workflow templates referenced their environment assets with the floating `azureml:<name>:latest` tag, which resolves to whichever version was most recently registered — not reproducible or auditable.

## Changes

- Pin the direct-submit fallback in `training/il/workflows/azureml/lerobot-train.yaml`, `training/rl/workflows/azureml/train.yaml`, `evaluation/sil/workflows/azureml/isaaclab-evaluation.yaml`, and `evaluation/sil/workflows/azureml/lerobot-eval.yaml` to the explicit version derived from each script's default image (`scripts/lib/common.sh`), matching what `derive_azureml_environment_version_from_image` produces at submit time.
- Document each pin as a direct-submit fallback in-place, and cross-reference the sync obligation from `scripts/lib/common.sh` back to the four templates, since `scripts/update-image-digests.sh` does not rewrite these derived version strings.
- Pipeline components (`components/train.yaml`, `components/preprocess.yaml`, `components/evaluate.yaml`, `components/register.yaml`) already bind via inline digest-pinned images rather than `azureml:<name>:latest`, so no further pinning is needed there.

## Testing

- `npm run lint:yaml`
- `npm run spell-check` (0 issues)
- `Invoke-Pester scripts/tests/security/Test-DependencyPinning.Tests.ps1` (248/248 passed)
- `shellcheck scripts/lib/common.sh`

Fixes #1079

### DIFF
--- a/evaluation/sil/workflows/azureml/isaaclab-evaluation.yaml
+++ b/evaluation/sil/workflows/azureml/isaaclab-evaluation.yaml
@@ -33,8 +33,10 @@ identity:
 
 # All values below are placeholders and will be overridden by submit-azureml-isaaclab-evaluation.sh
 # This file serves as a structural template for the job definition.
+# The environment version below is a direct-submit fallback; it must match the
+# version derived from DEFAULT_ISAAC_LAB_IMAGE in scripts/lib/common.sh.
 code: .
-environment: azureml:isaaclab-training-env:latest
+environment: azureml:isaaclab-training-env:2.3.2-sha256-388dbc806f48359a964cb9f807feb226da95d0a107f470fdcad9780ea10fe6f2
 compute: azureml:cpu-cluster
 
 # Inputs follow the same pattern as isaaclab-train.yaml
--- a/evaluation/sil/workflows/azureml/lerobot-eval.yaml
+++ b/evaluation/sil/workflows/azureml/lerobot-eval.yaml
@@ -27,8 +27,10 @@ identity:
 
 # Code root is the repository root so the entrypoint and helper scripts resolve
 # consistently when the template is submitted directly or via the wrapper script.
+# The environment version below is a direct-submit fallback; it must match the
+# version derived from DEFAULT_LEROBOT_EVAL_IMAGE in scripts/lib/common.sh.
 code: ../../../..
-environment: azureml:lerobot-inference-env:latest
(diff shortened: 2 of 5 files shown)


algattik:
Superseded by a PR from a renamed branch (fix/1079-pin-azureml-environment-versions).

PR #1384, opened 2026-08-18
fix(training): pin AzureML environment templates to explicit versions

### LINKED ISSUE #1079: Pin AzureML environment assets to explicit versions

algattik:
## Problem and context

Azure ML command-job templates referenced environment assets as `azureml:<name>:latest`. Azure ML environment references identify registered assets, not OCI images, so the final segment is an Azure ML asset version. The `latest` name is explicit syntax but remains operationally ambiguous because it does not tie the repository template to the image selected by the submit script.

Issue #1079 requires standalone command-job templates to use explicit versions while preserving submit-time environment registration.

The submit scripts already derive an environment version from the selected container image, register that exact asset, verify the registered image, and override the template with `--set environment=azureml:<name>:<version>`. They use `derive_azureml_environment_version_from_image`, `register_azureml_environment`, and the final `--set` argument. The change aligns each template's direct-submit fallback with that established path.

## Proposed solution

The change derives Azure ML environment versions from the checked-in image tag and immutable digest. For example, an image ending in `:2.11.0-cuda12.8-cudnn9-runtime@sha256:<digest>` maps to the environment version `2.11.0-cuda12.8-cudnn9-runtime-sha256-<digest>`. `scripts/lib/common.sh` owns the derivation function because all Azure ML submit scripts already source that library.

Standalone training and evaluation YAML templates use the derived versions directly. `scripts/update-image-digests.sh` owns synchronization because it already resolves current OCI digests after image-tag changes. Its write and dry-run modes update both OCI digest pins and the dependent Azure ML asset versions; its check mode reports both kinds of drift without modifying files.

`scripts/security/Test-DependencyPinning.ps1` adds a separate `azureml-environments` validator. This separation is required because Azure ML assets and OCI images have different reference grammars and immutability mechanisms. The validator accepts explicit Azure ML versions, rejects labels, unversioned references, and the repository-prohibited version name `latest`, and supports the existing `# pinning-ignore` convention for intentional exceptions.

### Environment-version derivation and template pins

`scripts/lib/common.sh` documents the relationship between checked-in default images and derived Azure ML environment versions. The derivation function validates tags and digests, lowercases digest text, and returns a version accepted by Azure ML.

The command-job templates under `training/rl/workflows/azureml/`, `training/il/workflows/azureml/`, `training/vla/workflows/azureml/`, and `evaluation/sil/workflows/azureml/` replace `latest` with the version derived from the default image used by their submit script.

Inline comments explain that direct submission requires prior asset registration and that the update script maintains alignment.

### Digest updater integration

`scripts/update-image-digests.sh` adds a mapping from each checked-in image default to its Azure ML environment name and template path. It discovers mapped templates, rejects unmapped Azure ML workflow references when the repository contains the mapped workflow set, resolves image digests before writing, derives replacement versions, and applies digest and environment changes in one candidate-file pass.

The updater retains main's `--check`, `--dry-run`, `--sarif-output`, and `--config-preview` modes. Check mode includes Azure ML environment drift in its exit status and summary. SARIF uses separate rules for OCI digest drift and Azure ML environment-version drift.

### Dependency pinning policy

`scripts/security/Test-DependencyPinning.ps1` registers `azureml-environments` as a default scan type for YAML under workflow directories. `Get-AzureMLEnvironmentViolations` parses scalar environment references in common YAML forms, distinguishes explicit versions from labels and missing versions, applies adjacent ignore comments, and emits warning-level dependency violations.

The workflow fixtures add one compliant explicit environment version and one deliberately mutable reference. The Pester suite covers supported Azure ML reference forms, ignore behavior, scanner discovery, default inclusion, and isolation from the OCI image validator.

### Synchronization tests

`scripts/tests/security/AzureMLEnvironmentPins.Tests.ps1` treats the updater's pin map as the enumerated contract. It verifies that every standalone Azure ML workflow reference is mapped, each checked-in template matches the version derived from its default image, dry-run leaves files untouched, write mode updates all mapped templates and is idempotent, indented YAML is handled, and newly introduced unmapped workflow references fail.

### Documentation and repository policy

`.github/copilot-instructions.md`, `docs/reference/workflow-templates-azureml.md`, and `docs/security/README.md` record the explicit-version rule, direct-submit registration requirement, updater command, and dependency-scanner behavior. This rationale is required by repository policy because the change modifies a public workflow configuration interface and its operational maintenance procedure.

## Summary

- Pin five standalone Azure ML command-job templates to explicit environment versions derived from their container image tag and digest.
- Extend `scripts/update-image-digests.sh` to synchronize OCI digest pins and dependent Azure ML environment versions in write, dry-run, and check modes.
- Add Azure ML environment validation to `Test-DependencyPinning.ps1` and first-class `azureml-environment-version-drift` SARIF findings to the scheduled freshness check.

Closes #1079.

## Design

`azureml:<name>:latest` names an explicit Azure ML asset version rather than a label, but it does not correlate the checked-in template with the image selected by the submission path. The templates now use:

```text
<image-tag>-sha256-<image-digest>
```

The training and evaluation submit scripts already derive this version, register the environment asset, verify that an existing asset uses the expected image, and override the template at submission time. Direct YAML submission therefore remains deterministic but requires the referenced environment version to be registered first.

The synchronization map covers Isaac Lab training and evaluation, LeRobot training and evaluation, and VLA pi0 training. Its tests reject newly introduced standalone Azure ML workflow references that are not mapped. The dependency scanner separately validates Azure ML asset syntax because Azure ML versions and OCI digests use different reference formats. Intentional mutable references can use the existing `# pinning-ignore` exemption.

Check mode reports both image-digest and Azure ML environment-version drift through SARIF. Exit code 2 remains a non-gating drift signal; resolution and report-generation failures return exit code 1.

## Validation

- 301 Pester pinning and synchronization tests
- PSScriptAnalyzer across 92 PowerShell files
- ShellCheck and Bash syntax validation
- Actionlint across 55 GitHub Actions workflows
- Repository spell check
- Targeted Markdown lint
- Non-mutating registry resolution and SARIF smoke test
- Repository-wide Azure ML environment scan with zero violations
- **e2e tests run: test_e2e_aml_rl_lifecycle, test_e2e_aml_il_lifecycle, test_e2e_aml_vla_pi0_training**


katriendg (approved):
Great one, thanks. Looks good to merge as far as I see fully complete and thanks also for the extensive integration with new tests, and the additions within the existing Bash script `update-image-digest.sh`.

PR #1387, opened 2026-08-18
feat(skills): adopt hve-core RPI skill workflow

### LINKED ISSUE #1151: Adopt hve-core's skill-based RPI workflow for the cloud agent
## Problem

The cloud agent's RPI (Research → Plan → Implement) capability is sourced from hve-core's **agent/persona** model. The `Bootstrap hve-core RPI persona` step in `.github/workflows/copilot-setup-steps.yml` downloads an umbrella agent and its subagents from a pinned hve-core release into `.copilot-tracking/upstream/hve-core-rpi/`, and the `Physical-AI RPI` umbrella plus its hidden worker read those files at session start.

hve-core is retiring that model. The RPI workflows now live as **skills** (`.github/skills/rpi/*` on hve-core `main`: plan, research, implement, review, quick) and ship as a breaking change in the upcoming hve-core 4.0.0 release. The legacy `.github/agents/hve-core/rpi-agent.agent.md` and its subagents that this repo fetches are on the path to removal.

## Gap

While the pin sits at hve-core v3.2.2 the bootstrap still succeeds, but the persona files it depends on are being superseded upstream. The step is `continue-on-error`, and the umbrella validates the downloaded persona files at session start — so once the pin advances past the upstream removal, RPI cloud sessions fail fast with a missing-files error rather than degrading quietly.

The persona-drift freshness signal that might have warned us here was intentionally dropped in #1132: for an agent→skill migration, "re-pin to a newer release" is the wrong remedy, and adopting the skill is the correct response. That leaves this adoption as the outstanding work.

## Desired outcome

- The cloud agent obtains its RPI capability from hve-core's RPI **skill** instead of the persona download.
- The cloud-agent environment provisions that skill so cloud sessions match what local contributors get.
- The persona bootstrap and the `Physical-AI RPI` umbrella/worker indirection are retired once the skill path is in place.
- `docs/reference/copilot-artifacts.md` reflects the skill-based RPI.

## Context

- Follow-up from #1132.
- Upstream: hve-core RPI skills under `.github/skills/rpi/`; hve-core 4.0.0 release in flight (release-please), where the skill-forward RPI workflows land under BREAKING CHANGES.

algattik:
## Problem and Context

The repository previously gave GitHub Copilot coding agent an RPI workflow through two custom agents that downloaded and interpreted upstream personas from `microsoft/hve-core`.

Issue #1151 records that hve-core is replacing this persona model with native skills. The cloud agent cannot fetch arbitrary upstream content during its session, so `.github/workflows/copilot-setup-steps.yml` must provision the skills before the coding session starts.

## Proposed Solution

The branch removes the two checked-in RPI agents and installs the upstream RPI skill suite directly into `.github/skills/rpi-*`, where Copilot discovers skills as immediate children of `.github/skills/`.

The setup workflow uses `gh skill install` with eight explicit upstream skill paths and an immutable hve-core commit. GitHub CLI preserves each skill's references and templates and injects source commit and tree metadata into `SKILL.md`. The generated directories remain gitignored and excluded from repository Markdown linting.

The runtime skill pin remains separate from `HVE_CORE_DERIVED_FILES_REF`, which tracks the reviewed release baseline for unrelated hve-core-derived PowerShell files.

### Runtime provisioning

`.github/workflows/copilot-setup-steps.yml` installs these pinned skills after all core environment setup:

- `rpi-quick`
- `rpi-research`
- `rpi-plan`
- `rpi-implement`
- `rpi-review`
- `rpi-challenger`
- `rpi-plan-critique`
- `rpi-walkthrough`

Each installation selects the exact upstream `SKILL.md` path, targets `.github/skills`, uses the reviewed commit through `--pin`, and fails the setup workflow on error.

### Retirement of the persona adapter

`.github/agents/physical-ai-rpi.agent.md` and `.github/agents/physical-ai-rpi-worker.agent.md` are deleted. Repository-specific risk overlays and PR artifact-persistence rules move into `.github/copilot-instructions.md`, where they apply to the cloud agent using the native skills.

### Discovery and documentation

The setup workflow also downloads hve-core's shared `.copilot-tracking` instruction from the same pinned commit as the skills. `.gitignore`, lint configuration, `docs/reference/copilot-artifacts.md`, `docs/contributing/architecture.md`, and path-specific instructions describe the runtime-only artifacts, direct discovery model, RPI lifecycle, and repository-specific safeguards.

### Freshness-check separation

`scripts/security/Test-HveCoreFreshness.ps1` parses `HVE_CORE_DERIVED_FILES_REF` instead of treating the runtime RPI pin as the baseline for release-derived PowerShell files. Its issue and summary output identify the dedicated baseline.

## Validation

- installed all eight skills and 23 files from the pinned commit with `gh skill install`
- passed the focused hve-core freshness suite: 87 tests
- passed the full Pester suite: 1,563 passed and 1 skipped
- passed workflow YAML and actionlint validation
- passed PowerShell, Markdown, and spelling checks

Closes #1151


scripts/ci/bootstrap-hve-core-rpi-skills.sh
```diff
@@ -0,0 +1,244 @@
...
+    local endpoint="$1"
+    local attempt
+    local max_attempts=3
+    local output_file
+    local error_file
+
+    output_file="$(mktemp)"
+    error_file="$(mktemp)"
+    for ((attempt = 1; attempt <= max_attempts; attempt++)); do
+        if gh api "$endpoint" >"$output_file" 2>"$error_file"; then
```
agreaves-ms: Can we try using the `gh skill` command to install the microsoft/hve-core skill library? I haven't tried it yet but it might clean up some of the manual downloading and jq checking you have here: https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills#managing-skills-with-github-cli

algattik: Great idea! Addressed in f7c52ba. The setup workflow now uses `gh skill install` for eight explicit hve-core skill paths, pins every install to the reviewed commit, and targets `.github/skills`. I removed the custom downloader, its transport-specific tests, and the obsolete audit/staging documentation while retaining the allowlist and fatal workflow contract.

katriendg (approved):
I love this approach, thanks for changing the broken imports to the agents.
`gh skill install` approach is pretty cool, I did not know it either.

One only thing to call out, the RPI skills are mostly self contained, but they do reference a shared instruction file [.github/instructions/hve-core/copilot-tracking.instructions.md](https://github.com/microsoft/hve-core/blob/main/.github/instructions/hve-core/copilot-tracking.instructions.md), which the plugin does package but the Skill install approach does not. 

Can we do without the extra instructions? Or could we import that file into local instructions as well. Either that or adding some context around additional `.copilot-tracking` into the local `copilot-instructions.md`

`gh plugin install` unfortunately is not something that seems to exist.

algattik:
Addressed in `d360cd3`. The setup now downloads hve-core’s shared `copilot-tracking.instructions.md` from the same pinned revision as the RPI skills, keeps repository-specific overlays local, and includes contract coverage for the download and generated-artifact handling. Thanks for identifying the missing shared dependency.

agreaves-ms (approved):

PR #1388, opened 2026-08-18
security(build): detect container image digest drift

### LINKED ISSUE #1093: feat(security): detect drift/staleness of pinned container image digests
## Component

Other

## Problem Statement

Container image references across the repository are digest-pinned (`repo:tag@sha256:…`) in workflow YAML, Kubernetes manifests, Helm values, and shell libraries, and CI enforces that every readable image reference carries a digest. Enforcement proves a digest is *present*; it cannot tell whether that digest is still the one the human-readable tag currently resolves to upstream.

Every other pinned dependency class in the repo has a loop that re-checks the pin against upstream over time — binaries are re-hashed against their source and checked against the releases API for newer versions, and GitHub Actions pins are checked for SHA staleness. Container images are the one surface pinned *without* such a loop: the digests are refreshed only by a manually run script, so between runs a pinned digest can silently diverge from what its tag now serves — or grow stale relative to a newer image — with no signal. There is no scheduled, tamper-evidence check that re-resolves each image tag to its current registry digest and reports drift.

## Proposed Solution

Give container images the same freshness and tamper-evidence coverage the binary and GitHub Actions pins already have: a scheduled check that, for each digest-pinned image reference in the repository, re-resolves the tag to its current registry digest and reports drift (optionally flagging staleness against a newer tag), surfaced non-gating to the Security tab and/or as annotations.

Coverage should span **every** surface where image digests live — workflow YAML, Kubernetes manifests, Helm values, and shell libraries — discovered by digest-reference shape rather than tied to a single YAML key or path, so the freshness signal stays symmetric with how the pins are refreshed. The registry tag→digest resolution needed here already exists in the repository and should be shared rather than reimplemented.

## Alternatives Considered

- **Rely on the manual digest-refresh script run ad hoc.** Leaves drift undetected between runs and depends on a maintainer remembering to run it; provides no scheduled signal.
- **Treat base-image CVE scanning as sufficient.** CVE scanning surfaces vulnerabilities in the image layers, not tag→digest drift, and targets a different (Dockerfile `FROM`) surface. It does not answer "has this pinned digest diverged from its tag?"
- **Do nothing and accept enforcement-only.** Static enforcement guarantees a digest exists but never revalidates it, so a re-pushed tag or a superseded image goes unnoticed.

## Additional Context

Related in-flight work:

- **#1082** — establishes digest-pinning of container images and CI enforcement that pins exist. This request is the "one step further": revalidating those pins over time rather than only asserting their presence.
- **#1088** (PR, in flight) — extends the binary freshness/staleness loop to ORAS artifacts and adds digest-pin enforcement across the image surfaces (workflow YAML, Kubernetes manifests, Helm values). It sets the freshness pattern this request generalizes to container images.
- **#1085** — container base-image CVE scanning. Sibling effort on the *Dockerfile `FROM`* surface: it scans for and remediates CVEs (including CVE-triggered digest bumps) and already contains registry tag→digest resolution. This request covers the complementary surface (non-Dockerfile image references) and CVE-independent digest drift; the two should share the resolution logic rather than duplicate it.
- **#1091** — hardening of CI release and dependency validation, including a shared dependency-pinning workflow entrypoint with SARIF and annotations. That scheduled/SARIF surface is a natural place for an image-drift check to plug in.
- **#522** — the binary-dependency staleness signal this request mirrors for images.

algattik:
This PR adds a scheduled check for digest-pinned container image tags whose checked-in digest no longer matches the registry.

## Problem and Context

Container references in this repository commonly combine a readable tag with an immutable digest:

```text
registry.example/image:tag@sha256:<digest>
```

The digest fixes the exact manifest used by a deployment. Existing checks establish that a digest is present, but they do not periodically re-resolve the tag and compare its current registry digest with the checked-in value.

Issue #1093 documents the requirement for a scheduled, non-gating freshness signal covering digest references in workflow YAML, Kubernetes manifests, Helm values, and shell libraries. It also asks the implementation to discover references by shape and reuse existing registry-resolution behaviour.

## Proposed Solution

The branch expands `scripts/update-image-digests.sh` from a digest updater into a repository-wide discovery, update, preview, and check command.

| Mode | Behaviour |
| --- | --- |
| Default | Resolve discovered tags and rewrite stale digest pins |
| `--dry-run` | Show the rewrites without changing tracked files |
| `--check` | Report stale occurrences without changing tracked files and exit 2 when drift exists |
| `--config-preview` | List discovered image tags and files without registry access |

Check mode can serialize findings as SARIF. A weekly GitHub Actions workflow accepts exit code 2 as a non-gating drift result, rejects operational failures, and uploads SARIF to the Security tab.

The implementation discovers references with `git grep`, resolves OCI manifest digests through registry HTTP APIs, and excludes files maintained by other mechanisms: Dockerfiles and Compose files, gh-aw compiled workflows, test fixtures, and Pester test files.

### Discovery and registry resolution

`scripts/update-image-digests.sh` defines the accepted tag-plus-digest shape, searches tracked files, normalizes Docker Hub references, obtains anonymous pull tokens for Docker Hub and NGC, and reads `Docker-Content-Digest` from registry manifest responses.

The issue establishes repository-wide shape-based discovery and reuse of registry-resolution behaviour. The specific registry routing and authentication design is inferred from the prior updater and the new tests; no ADR documents this protocol boundary.

### Drift evaluation and source rewriting

Check mode records each stale occurrence with its file, line, columns, reference, pinned digest, and resolved digest. Default and dry-run modes use a boundary-aware replacement expression so a short image name is not rewritten inside a longer image name.

Exit code 2 distinguishes detected drift from operational errors. This contract is documented in the script help, security documentation, tests, and workflow.

### SARIF generation

Check mode can write warning-level SARIF results, one per stale occurrence. The writer rejects directories and tracked repository files, creates parent directories, renders through a temporary file, and atomically moves a complete report into place.

The scheduled workflow uses a stable SARIF category so recurring runs contribute to one code-scanning analysis category.

### Scheduled workflow

`.github/workflows/check-image-digest-freshness.yml` runs each Monday and supports manual dispatch. It grants read-only repository access and job-scoped `security-events: write`, runs the script in check mode, treats drift as non-gating, propagates other failures, and uploads SARIF when a report exists.

### Test harness and behavioural tests

`scripts/tests/Mocks/BashScriptHarness.psm1` accepts an argument array for Bash entry points. `scripts/tests/security/update-image-digests.Tests.ps1` creates temporary Git repositories and stubs registry calls to exercise discovery, update, dry-run, check, SARIF, authentication routing, exclusion rules, boundary matching, counters, and failure paths.

### Documentation and repository metadata

`docs/security/README.md` documents the operational check, ownership exclusions, exit codes, and remediation command. `docs/security/workflow-permissions.md` records the additional job-scoped SARIF permission. `.github/copilot-instructions.md` adds the weekly check to the security workflow inventory. `.gitignore` excludes interrupted Pester test artefacts.

Closes #1093


scripts/update-image-digests.sh
```diff
@@ -72,63 +107,196 @@ cd "$REPO_ROOT"
...
   local host="$1" repo="$2" tag="$3" token="" token_url="" manifest_url
   local accept='application/vnd.oci.image.index.v1+json,application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json'
+  local curl_args=(-fsSL --proto '=https' --proto-redir '=https' --connect-timeout 10 --max-time 30 --retry 2)
   # Registry API metadata calls (anonymous pull token + manifest HEAD), not
   # artifact downloads: the sha256 digest is itself the integrity value being
   # resolved here, so there is nothing to checksum.
   case "$host" in
-    registry-1.docker.io) token_url="https://auth.docker.io/token?service=registry.docker.io&scope=repository:${repo}:pull" ;;
-    nvcr.io)              token_url="https://nvcr.io/proxy_auth?scope=repository:${repo}:pull" ;;
+  registry-1.docker.io) token_url="https://auth.docker.io/token?service=registry.docker.io&scope=repository:${repo}:pull" ;;
```
katriendg, line 119: Generic registries requiring a bearer-token challenge are not supported here. For example, public GHCR returns `401` with a `WWW-Authenticate` challenge, but token acquisition is hard-coded for Docker Hub and NGC. This conflicts with the stated support for GHCR, Quay, and ACR and will make the scheduled check fail when such an image is added. 
Let's explicitly restrict the supported registry set here?

scripts/update-image-digests.sh
```diff
@@ -1,60 +1,95 @@
...
+if [[ -n "$sarif_output" && "$check" != "true" ]]; then
+  fatal "--sarif-output requires --check"
+fi
+
 require_tools git curl jq
 
 # Reference shape: registry/repo:tag@sha256:<64 hex>. Pathspecs exclude the files
 # whose digests are maintained by other tooling (Dependabot, gh-aw).
-digest_ref_re='[A-Za-z0-9][A-Za-z0-9._/-]*:[A-Za-z0-9._-]+@sha256:[0-9a-f]{64}'
+digest_value_re='sha256:[0-9a-f]{64}'
```
katriendg, line 90: This reference regex does not support registry hosts with ports. A valid reference such as `registry.example.com:5000/team/robot:1.0@sha256:...` is matched as `5000/team/robot:1.0`, which is then treated as a Docker Hub reference. That could resolve or rewrite against the wrong image instead of failing safely. Please consider support for parsing OCI references with host-port support.

katriendg (approved):
Thanks Alexandre! Left two minor comments, approving as they get closed.

scripts/update-image-digests.sh
algattik: Addressed in 4f1aae4. The script help and security documentation now define the supported boundary: anonymous OCI registries are supported, while anonymous pull-token acquisition is limited to Docker Hub and NGC. The misleading claim that all dotted-host registries such as GHCR, Quay, and ACR are supported was removed.

scripts/update-image-digests.sh
algattik: Addressed in 4f1aae4. Discovery now parses an optional registry host and numeric port as part of the OCI reference, and resolution recognizes colon-bearing first path components as registry hosts. A reference such as registry.example.com:5000/team/robot:1.0 is therefore sent to registry.example.com:5000 rather than being truncated and routed to Docker Hub.

PR #1390, opened 2026-08-18
security(training): restrict checkpoint deserialization

### LINKED ISSUE #1097: Harden SKRL/RSL-RL checkpoint loading against pickle deserialization (training resume, play, monitor)
## Problem

Policy checkpoints resumed or replayed through the SKRL and RSL-RL runners are loaded with the frameworks' native loaders (`runner.load()`, `runner.agent.load()`), which deserialize the checkpoint with an unrestricted Python unpickler. A checkpoint sourced from local training logs, blob storage, MLflow, or the model registry is therefore a code-execution vector: loading a tampered file runs arbitrary code inside the training or evaluation container.

The evaluation/inference path loads checkpoints under a restricted `weights_only=True` unpickler, so a hostile checkpoint fails safely there. The following paths still pass the raw file straight to the framework loader with no such guard:

- Training resume — `training/rl/scripts/skrl_training.py` (`runner.agent.load`) and `training/rl/scripts/rsl_rl/train.py` (`runner.load`)
- Software-in-the-loop replay — `evaluation/sil/play.py` (`runner.load`)
- Checkpoint monitoring — `evaluation/sil/monitor_checkpoints.py` (`runner.load`)

The SKRL evaluation path is only partially covered: it validates the checkpoint with the restricted unpickler and then re-loads the same file through the framework loader, leaving a load-after-validate window between the check and the real read.

## Desired outcome

Every checkpoint load — including those routed through the SKRL and RSL-RL runners in the training-resume, replay, and monitoring paths — restricts or rejects pickle deserialization of untrusted checkpoints, matching the guarantee the evaluation/inference path already provides. Loading a tampered checkpoint fails safely rather than executing code, with no window in which an unrestricted read of the checkpoint can occur.

Follow-up to #1086.

algattik:
This PR prevents SKRL and RSL-RL checkpoint paths from executing arbitrary Python objects embedded in checkpoint files.

## Problem and Context

PyTorch checkpoints can contain pickled Python objects. Loading a hostile checkpoint with unrestricted pickle deserialization can execute code inside a training or evaluation container.

Issue #1097 identifies four vulnerable framework-owned load paths: SKRL training resume, RSL-RL training resume, software-in-the-loop replay, and checkpoint monitoring. It also identifies a load-after-validate window in SKRL policy evaluation.

## Solution

Checkpoint deserialization is centralized in `training/utils/integrity.py`:

| Helper | Purpose |
|--------|---------|
| `safe_load_checkpoint` | Deserialize checkpoint mappings with `weights_only=True` and translate restricted-unpickler failures into actionable errors |
| `safe_load_skrl_checkpoint` | Deserialize once, then apply state to the SKRL agent's registered checkpoint modules |
| `safe_load_rsl_rl_checkpoint` | Deserialize once, then apply state through the RSL-RL algorithm loader and restore iteration metadata |

The framework adapters preserve each pinned framework's state-loading semantics without replacing `torch.load` or changing process environment variables. Unrelated threads and child processes cannot observe a process-global deserialization override.

Restricted-load error detection recognizes PyTorch's stable outer `pickle.UnpicklingError` message as well as explicit safe-unpickler markers. Protocol 4 and 5 checkpoints therefore receive the same actionable `ValueError` contract.

## Covered Paths

- SKRL training resume
- RSL-RL training resume
- software-in-the-loop replay
- checkpoint monitoring
- SKRL and RSL-RL policy evaluation

The AzureML LeRobot submission scripts also suppress ShellCheck reachability diagnostics for functions invoked indirectly through signal traps.

## Validation

- focused Ruff checks for changed Python files
- 1,001 training and evaluation tests passed; 3 optional-dependency tests skipped

Fixes #1097


evaluation/tests/test_rsl_rl_checkpoint_routing.py
```diff
@@ -0,0 +1,178 @@
...
+    module = ModuleType(name)
+    for attribute, value in attributes.items():
+        setattr(module, attribute, value)
+    return module
+
+
+class _AppLauncher:
+    @staticmethod
+    def add_app_launcher_args(parser: object) -> None:
+        del parser
```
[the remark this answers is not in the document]
algattik: Addressed in e0d741c by removing the unnecessary `del parser` statement from the test stub.

training/utils/integrity.py
```diff
@@ -27,11 +82,72 @@ def safe_load_checkpoint(path: str, *, map_location: str = "cpu") -> dict:
...
+
+        def safe_torch_load(*args: Any, **kwargs: Any) -> Any:
+            nonlocal torch_load_invoked
+            torch_load_invoked = True
+            kwargs["weights_only"] = True
+            return original_torch_load(*args, **kwargs)
+
+        _set_environment_value(_FORCE_WEIGHTS_ONLY_ENV, "1")
+        _set_environment_value(_DISABLE_WEIGHTS_ONLY_ENV, None)
+        torch.load = safe_torch_load
```
katriendg: This process-global replacement is observable by unrelated threads and inherited by child processes; the lock serializes only callers of this helper. I reproduced an unrelated concurrent `torch.load(..., weights_only=False)` being silently forced into restricted mode and failing.

Recommended fix: avoid process-global interception by using framework-specific adapters that deserialize once through `safe_load_checkpoint()` and apply the resulting state, or isolate the framework-owned load in a dedicated process so unrelated work cannot observe the override.

training/utils/integrity.py
```diff
@@ -1,11 +1,66 @@
...
+        f"weights_only=False. Underlying error: {error}"
+    )
+
+
+def _is_weights_only_error(error: Exception) -> bool:
+    current: BaseException | None = error
+    seen: set[int] = set()
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        if any(marker in str(current) for marker in _WEIGHTS_ONLY_ERROR_MARKERS):
```
katriendg, line 39: Could we broaden this detection to PyTorch's stable outer restricted-load error? I reproduced benign protocol 4 and 5 checkpoints raising raw `pickle.UnpicklingError` (`Weights only load failed` / `Unsupported operand`) because none of these markers match. That bypasses the actionable `ValueError` contract, and the raw PyTorch message recommends retrying with `weights_only=False`.

Recommended fix: recognize `pickle.UnpicklingError` messages beginning with `Weights only load failed.` while walking the exception chain, retain the explicit `pickle_module` handling, and add protocol 4/5 regression cases.

katriendg (approved):
Thank you for hardening the checkpoint-loading paths and for the thorough malicious-payload, routing, restoration, and concurrency coverage. The change looks good overall and closes the vulnerable framework-owned load paths identified in #1097.

Approving with two inline comments to address before merge: broaden restricted-unpickler error classification so operators consistently receive the safe actionable error, and remove or isolate the process-global `torch.load` interception so unrelated concurrent work is unaffected.

training/utils/integrity.py
algattik: Addressed in f65567d. Restricted-load detection now recognizes pickle.UnpicklingError messages beginning with the stable "Weights only load failed." prefix while retaining explicit pickle_module markers. Protocol 4 and 5 regression cases verify the actionable ValueError contract.

training/utils/integrity.py
algattik: Addressed in f65567d. The process-global environment and torch.load overrides were removed. SKRL and RSL-RL now use framework-specific adapters that deserialize once with safe_load_checkpoint() and apply the loaded state through each framework’s state-loading API, so unrelated threads and child processes cannot observe an interception.

PR #1502, opened 2026-09-09
fix(pipeline): close cloud E2E runtime gaps

### LINKED ISSUE #1490: test(training): close cloud E2E runtime validation gaps
## Component

Cloud E2E coverage for Azure ML, OSMO, runtime dependency installation, and MLflow integration

## Problem Statement

The cloud E2E suite leaves three production integration boundaries insufficiently validated:

1. The Azure ML to OSMO proxy has a CPU smoke workflow and production submission script, but no automated E2E invokes the path. Dependency or Azure ML MLflow plugin incompatibility can leave the Azure ML job apparently successful because proxy MLflow failures are non-fatal.
2. The OSMO RL tests execute `training/rl/scripts/setup_isaac_runtime.sh`, but do not prove that the runtime loaded the versions exported from `training/rl/uv.lock` rather than compatible packages already present in the container image.
3. The OSMO VLA fine-tuning test executes `training/vla/scripts/groot/osmo-train-entry.sh`, but does not verify MLflow results, model registration, or pinned runtime and base-model provenance.

These gaps can allow dependency, plugin, or runtime-bootstrap regressions to pass task-level success checks.

## Proposed Solution

Add one CPU-only AML-to-OSMO proxy E2E lane and strengthen the existing OSMO RL and VLA tests with runtime and integration assertions. Reuse existing production entrypoints and cloud jobs rather than adding duplicate GPU lanes.

### AML-to-OSMO proxy

- [ ] Add a tracked default E2E test for `workflows/azureml/submit-osmo-proxy-job.sh`, `workflows/azureml/osmo-proxy-job.yaml`, and `workflows/osmo/smoke-test-proxy-e2e.yaml`.
- [ ] Submit the proxy through Azure ML and wait for both the Azure ML job and OSMO workflow to complete.
- [ ] Verify the OSMO task ran successfully and produced the declared durable blob output.
- [ ] Verify the proxy registered the expected Azure ML data asset.
- [ ] Query the Azure ML MLflow tracking URI and verify `osmo.workflow_id`, terminal status, task counts, success rate, and duration.
- [ ] Fail when the Azure ML MLflow plugin is unavailable or metric logging is absent, even if the proxy process otherwise exits successfully.

### OSMO RL runtime

- [ ] Extend an existing OSMO RL E2E to capture machine-readable runtime provenance after `setup_isaac_runtime.sh`.
- [ ] Verify the installed runtime versions match the committed `training/rl/pyproject.toml` and frozen lock resolution.
- [ ] Verify the assertion cannot pass solely because compatible packages are preinstalled in the base image.

### OSMO VLA runtime and MLflow

- [ ] Extend `test_e2e_osmo_vla_finetune` rather than adding another GPU workflow.
- [ ] Verify the runtime and base-model provenance produced by `training/vla/scripts/groot/osmo-train-entry.sh`.
- [ ] Verify the expected Azure ML MLflow run, correlation identifiers, metrics, and model-registration result.

### E2E lifecycle

- [ ] Preserve exact submission commands, Azure ML job and OSMO workflow identifiers, attempts, logs, retry classification, and terminal state in the E2E handle.
- [ ] Clean up proxy outputs, registered data assets, MLflow/model artifacts created for the test, and other test-owned cloud resources.
- [ ] Keep the proxy lane CPU-only and avoid adding duplicate RL or VLA GPU jobs.

## Out of Scope

- HiL edge coverage tracked by #1489
- Physical robot hardware and laboratory networking
- New GPU SKUs, node pools, or infrastructure changes

algattik:
## Summary

- harden Azure ML and OSMO cloud E2E runtime validation and cleanup
- validate MLflow provenance, model registration, and runtime dependency state
- align OSMO LeRobot submission arguments with the current CLI

## Validation

- all 10 cloud E2E test modules passed against the deployed environment
- `npm run lint:py`
- `uvx --with ruff==0.15.21 ruff format --check .`
- `npm run spell-check`
- `env -u UV_DEFAULT_INDEX uv lock --check --offline`
- PR validation summary passed

Closes #1490


data-management/viewer/frontend/src/api/export.ts
```diff
@@ -159,61 +106,29 @@ export function createExportStream(
         const lines = buffer.split('\n')
         buffer = lines.pop() ?? ''
 
+        let currentEventType = 'message'
```
katriendg, line 109: **High - Preserve SSE state across reader chunks**

`currentEventType` is recreated for every network chunk. If an `event:` line and its `data:` line arrive in separate reads, the data is misclassified and completion or error callbacks may never run. Keep the event state outside the read loop, reset it only at an SSE event boundary, flush the decoder tail, and report EOF without a terminal event.

data-management/viewer/frontend/src/api/export.ts
```diff
@@ -83,15 +27,15 @@ export async function exportEpisodes(
...
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      ...(await mutationHeaders()),
     },
-    transformExportResult,
-  )
+    body: JSON.stringify(request),
+  })
+  return handleResponse<ExportResult>(response)
```
katriendg, line 38: **High - Normalize export statistics at the wire boundary**

The backend emits snake_case statistics such as `total_episodes`, while `ExportResult` requires camelCase fields such as `totalEpisodes`. This direct cast bypasses normalization, so successful exports can expose undefined statistics. Decode and validate both synchronous and streamed result payloads at this boundary.

data-management/viewer/frontend/src/lib/api-client.ts
```diff
@@ -172,73 +163,25 @@ export class ApiClientError extends Error {
...
  * Handle API response, throwing on error.
  */
-export async function handleResponse<T>(
-  response: Response,
-  transform: (data: unknown) => T = transformKeys<T>,
-): Promise<T> {
+export async function handleResponse<T>(response: Response): Promise<T> {
   if (!response.ok) {
-    let code = `HTTP_${response.status}`
+    let error: ApiError
```
katriendg, line 171: **High - Handle FastAPI error envelopes**

FastAPI commonly returns `{"detail": ...}`, but this path assumes every JSON error is an `ApiError`. That produces undefined message and code values. Parse the body as `unknown`, support the FastAPI envelope explicitly, and retain stable status-derived fallbacks for malformed or non-JSON errors.

data-management/viewer/frontend/src/lib/api-client.ts
```diff
@@ -66,22 +67,12 @@ export async function mutationFetch(
   const method = (init.method ?? 'GET').toUpperCase()
   const needsCsrf = method !== 'GET' && method !== 'HEAD'
   const baseHeaders = needsCsrf ? await mutationHeaders() : await requestHeaders()
-  const headers = new Headers(baseHeaders)
-  new Headers(init.headers).forEach((value, name) => headers.set(name, value))
   return fetch(input, {
     ...init,
-    headers: Object.fromEntries(headers.entries()),
+    headers: { ...baseHeaders, ...(init.headers ?? {}) },
```
katriendg, line 72: **High - Centralize native `fetch` in the Viewer API client**

Native `fetch` is correct here; the problem is that feature APIs, hooks, synchronization code, and components now call it directly instead of routing through shared helpers in this module. Keep `fetch` centralized here so authentication, CSRF, URL construction, error handling, header merging, and response normalization remain consistent.

data-management/viewer/frontend/src/lib/api-client.ts
```diff
@@ -398,15 +373,20 @@ export interface CacheStats {
...
+  const data = await handleResponse<unknown>(response)
+  return transformKeys<CacheStats>(data)
 }
 
 /**
  * Warm the episode cache for a dataset by preloading the first N episodes.
  */
 export async function warmCache(datasetId: string, count = 5): Promise<void> {
-  await apiRequest(`/datasets/${datasetId}/cache/warm?count=${count}`, {
+  await fetch(`${API_BASE}/datasets/${datasetId}/cache/warm?count=${count}`, {
```
katriendg, line 387: **Medium - Fail cache warming on HTTP errors**

The current implementation resolves successfully for 4xx and 5xx responses. Capture this response and route failures through `handleResponse` so cache warm-up cannot report success when the backend rejected it.

tests/e2e/test_e2e_aml_osmo_proxy.py
```diff
@@ -0,0 +1,149 @@
...
+            "-o",
+            "tsv",
+        ],
+        cwd=repo_root,
+    )
+    if result.returncode != 0 or result.stdout.strip().lower() != "true":
+        raise AssertionError(f"Proxy durable output was not found\n\n{format_command_failure(result)}")
+
+    asset_name = f"osmo-{workflow_id}-output-0"
+    request.addfinalizer(lambda: archive_aml_data_asset(repo_root, aml_workspace, asset_name))
```
katriendg, line 141: **Medium - Register data-asset cleanup before assertions**

The asset is created before this point, but its finalizer is registered only after MLflow, workflow, and durable-output assertions. Any earlier failure leaves it active. Capture the deterministic asset identity and register archival immediately after submission, before the first fallible post-submission assertion.

data-management/viewer/frontend/src/api/__tests__/export.test.ts
```diff
@@ -278,105 +210,18 @@ describe('createExportStream', () => {
...
-      totalFrames: 300,
-      percentage: 50,
-      status: 'processing',
-    }
-    mockFetch.mockResolvedValueOnce(
-      streamResponse([
-        `event: error\ndata: not-json\n\nevent: progress\ndata: ${JSON.stringify(progress)}\n\n`,
-      ]),
-    )
+  it('ignores malformed JSON data lines', async () => {
```
katriendg, line 213: **Medium - Restore export behavior tests**

The revised suite accepts malformed payloads and incomplete streams while the dedicated contract suite is removed. Add behavior tests using real `Response` and `ReadableStream` instances for arbitrary chunk boundaries, schema validation, snake_case normalization, redaction, and EOF without a terminal event.

tests/e2e/test_e2e_osmo_vla_finetune.py
```diff
@@ -32,6 +36,16 @@
 _VLA_TASK_NAME = "train"
 
 
+def _vla_runtime_provenance(logs: str) -> dict[str, object]:
```
katriendg, line 39: **Medium - Centralize provenance marker parsing**

RL and VLA tests duplicate marker scanning, JSON decoding, mapping validation, and error behavior. Move a typed parser into `tests/e2e/_common.py` so malformed provenance is handled consistently.

tests/e2e/_common.py
```diff
@@ -14,6 +15,22 @@
 _STATUS_HEARTBEAT_INTERVAL_SECONDS = 300
 
 
+@dataclass
+class E2EHandle:
```
katriendg, line 19: **Medium - Make lifecycle evidence completeness executable**

The evidence model has the expected fields, but successful runs never assert that commands, resource IDs, attempts, logs, terminal state, and cleanup registrations are complete. Add resource-specific completion assertions and represent cleanup registrations as verifiable evidence.

training/vla/scripts/groot/osmo-train-entry.sh
```diff
@@ -236,6 +240,34 @@ elif [ -z "${BASE_MODEL_REVISION:-}" ] && [ ! -d "${BASE_MODEL}" ]; then
...
+import os
+import pathlib
+import sys
+
+packages = ("accelerate", "flash-attn", "numpy", "opencv-python", "torch", "torchaudio", "torchvision")
+payload = {
+    "base_model": os.environ["BASE_MODEL_SOURCE"],
+    "base_model_revision": os.environ.get("BASE_MODEL_REVISION", ""),
+    "expected_runtime_versions": {
+        "accelerate": "1.14.0",
```
katriendg, line 255: **Medium - Derive VLA expected versions from one source**

These expected versions are declared separately from the installation inputs, allowing producer and test to agree on stale values. Derive them from the same manifest, lock, or exported variables used to install the runtime, then compare the observed versions against that independent source.

data-management/viewer/backend/src/api/routers/export.py
```diff
@@ -388,17 +377,19 @@ def progress_callback(progress: ExportProgress):
...
-            yield f"event: error\ndata: {json.dumps(payload)}\n\n"
-        except Exception:
-            logger.exception("Export stream failed")
-            payload = {"code": "EXPORT_FAILED", "message": "Export failed"}
-            yield f"event: error\ndata: {json.dumps(payload)}\n\n"
+        except ImportError as e:
+            error_msg = f"Export not available: {e}"
+            yield f"event: error\ndata: {json.dumps({'error': error_msg})}\n\n"
+        except Exception as e:
+            yield f"event: error\ndata: {json.dumps({'error': str(e)})}\n\n"
```
katriendg, line 392: **Medium - Keep backend exception text private**

Returning `str(e)` can expose filesystem paths, storage identifiers, and dependency diagnostics. Log the exception server-side and return a stable public message.

```suggestion
        except Exception:
            logger.exception("Export stream failed")
            yield f"event: error\ndata: {json.dumps({'error': 'Export failed'})}\n\n"
```

workflows/azureml/osmo-proxy/osmo_proxy.py
```diff
@@ -735,46 +710,37 @@ def _register_aml_data_assets(
...
+        workspace_name=workspace,
+    )
+    for idx, url in enumerate(output_urls):
+        asset_name = f"osmo-{workflow_id}-output-{idx}"
+        aml_url = _to_aml_url(url)
+        data_asset = Data(
+            name=asset_name,
+            path=aml_url,
+            type=AssetTypes.URI_FOLDER,
+            description=f"OSMO workflow {workflow_id} output {idx} (source: {url})",
```
katriendg, line 735: **Medium - Validate and redact proxy output URIs**

Raw output URLs are persisted in AML metadata and logs without scheme, account, container, query, or fragment validation. Require normalized `azure://` values for the expected account and container and reject query or fragment data. This suggestion removes the raw source URL from the persisted description; validation and log redaction still require the broader boundary change.

```suggestion
            description=f"OSMO workflow {workflow_id} output {idx}",
```

docs/osmo-proxy.md
```diff
@@ -32,11 +32,19 @@ flowchart TD
 # Using the submission wrapper (resolves workspace from Terraform outputs)
 workflows/azureml/submit-osmo-proxy-job.sh
```
katriendg, line 34: **Medium - Align the documented AML instance type**

This new invocation relies on the configured AML compute profile, but the prerequisite above names `cpu-standard` while the executable job sets `resources.instance_type=defaultinstancetype`. Update the prerequisite to the canonical executable name or change the job configuration to match.

katriendg (changes requested):
Some changes are requested for this substantial PR. The cloud E2E additions substantially cover issue #1490, but this commit also carries an undisclosed Viewer rewrite with confirmed regressions in export streaming, statistics normalization, shared error handling, and transport architecture. Please fully justify the Viewer scope, add to issue or link, restore the affected contracts and tests, complete cleanup and URI safeguards, then update from main and rerun validation. PRs #1640 and #1641 conflict with six Viewer files in this branch; if these Viewer changes remain, merge #1502 first after its blockers are fixed, then refresh #1640 and stacked #1641. This one should probably be merged before 1640 and 1641.

PR #1556, opened 2026-09-15
feat(deployment): add Xavier GPU offload for robot inference

### LINKED ISSUE #1555: feat(deploy): add Xavier GPU offload for robot inference
### Component

Other (GPU offload / fleet deployment inference)

### Problem Statement

Robots running policy inference on-board (for example Jetson-class or CPU-only edge devices) are limited by compute, power, and battery. The toolchain covers capture, training, evaluation, and edge deployment, but has no supported path to transparently offload inference from the robot to a nearby on-premise GPU without rewriting the policy or robot control loop.

### Proposed Solution

Add a `gpu-offload/` domain based on the Xavier remote-execution runtime:

- **Runtime (`gpu-offload/runtime`)**: remoter library that executes opted-in functions and classes on a remote GPU server over TCP/UDP/Unix sockets, using a restricted MessagePack codec (no pickle) with size and type limits.
- **Admission controller (`gpu-offload/controller`)**: Kubernetes mutating webhook that injects server/client configuration into workloads that explicitly opt in.
- **Helm chart (`gpu-offload/helm`)**: deploys the controller with TLS, RBAC, and workload opt-in; node agent template retained but disabled.
- **Specifications**: offload opt-in contract and `remote.yaml` schema.
- **Examples**: first-run GPU validation, SO-101 LeRobot real-hardware rollout, and UR10e single-arm.
- **Local setup scripts**: k3s, NVIDIA container toolkit, and preflight checks for single-laptop validation (T0-compatible, Kubernetes optional).
- **Docs**: README and tier docs updated to describe efficient inference offload.

Work is ready on `dwhiteddsoft:xavier-integration`, rebased on current `main`, with lint, formatting, frontmatter, and spell checks passing locally.

### Alternatives Considered

- Running inference fully on-robot: simplest, but constrained by device compute and battery life.
- Cloud-hosted inference endpoints: adds latency and connectivity dependence unsuitable for real-time control loops.
- Per-policy custom client/server code: duplicates effort for each model and does not generalize across robots.

### Additional Context

Benefits include reduced energy and battery consumption on the robot and lower hardware cost, while keeping the robot-side code largely unchanged through opt-in configuration.

dwhiteddsoft:
# Pull Request

## Description

Adds a `gpu-offload/` domain that transparently offloads robot policy inference to a nearby on-premise GPU, based on the Xavier remote-execution runtime.

- **Runtime (`gpu-offload/runtime`)**: remoter library for opted-in functions and classes over TCP/UDP/Unix sockets, using a restricted MessagePack codec (no pickle) with size and type limits
- **Admission controller (`gpu-offload/controller`)**: mutating webhook that injects server/client configuration into workloads that explicitly opt in
- **Helm chart (`gpu-offload/helm`)**: controller deployment with TLS, RBAC, and workload opt-in; node agent template retained but disabled
- **Specifications**: offload opt-in contract and `remote.yaml` schema
- **Examples**: first-run GPU validation, SO-101 LeRobot real-hardware rollout, UR10e single-arm
- **Local setup scripts**: k3s, NVIDIA container toolkit, and preflight checks (T0-compatible, Kubernetes optional)
- **Docs**: README and tier docs updated for efficient inference offload; frontmatter added to touched docs that lacked it

Draft while #1555 is in triage.

Closes #1555

## Type of Change

- [x] 📚 Documentation update

## Component(s) Affected

- [x] `docs/` - Documentation
- [x] `gpu-offload/` - New GPU offload domain (runtime, controller, Helm chart, examples, scripts)

## Testing Performed

No Terraform, Azure, or OSMO changes; the checks above do not apply.

**Validation performed on the final branch (local, macOS):**

- `ruff check` and `ruff format --check` (ruff 0.15.21, CI-pinned): pass
- `taplo fmt --check`, `uv lock --check`, `npm run lint:hfpins`: pass
- `cspell`, `markdownlint-cli2`, `markdown-table-formatter --check`, `npm run lint:links`: pass on changed files
- Frontmatter validation and ms.date freshness (`-ChangedFilesOnly`): 0 errors, 0 stale files
- `py_compile` on all Python files modified during lint cleanup: pass
- Ruff safe fixes reviewed: only unused imports, f-strings without placeholders, import sorting, and `__all__` ordering; no removed import is referenced elsewhere

**Environment used:** local workstation only; no Azure resources deployed for this sync.

**Cost incurred:** none.

**Known limitations:**

- Frontmatter validation reports 8 non-blocking warnings (mostly missing Copilot footer)

**Update (2026-09-18):** `gpu-offload/runtime` and `gpu-offload/controller` pytest suites now run
locally (all passing) and in CI via the new path-gated `Pytest GPU Offload` job
(`.github/workflows/pytest-gpu-offload.yml`), which also caught and fixed a Hatch
packaging bug in the controller's `pyproject.toml`. `shellcheck` and `actionlint`
(the tool behind the "YAML Lint" check) were also run locally and pass; both already
ran in CI prior to this update.

## Documentation Impact

## Bug Fix Checklist

*Not applicable: feature PR.*

## Checklist

- [x] No new linting warnings introduced (verified locally and in CI: shellcheck, actionlint, pytest)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


algattik:
The current CI does not exercise the new `gpu-offload` domain. In particular, it does not run `gpu-offload/runtime/tests` or `gpu-offload/controller/tests`, install or build either Python package, lint/render the Helm charts, build the new container images, or perform a CPU-only client/server and admission-mutation smoke test.

Most of this coverage does not require Xavier hardware or a GPU. Please add a path-gated `gpu-offload` CI job that runs both pytest suites, builds both packages and relevant images, validates the Helm charts, and performs a hardware-independent smoke test. GPU allocation, CUDA execution, performance, and real-robot validation can remain separate hardware-backed checks.

The PR description currently says these test suites are “relying on CI”, but no workflow references `gpu-offload`; please correct that statement until the missing coverage is added.

algattik:
Could you please add dedicated CI coverage for the new `gpu-offload` domain? Most of the required building blocks already exist, so it may be possible to extend the repository’s current controls and compose the validation assets introduced by this PR.

#### 1. Existing unit tests

The PR adds 44 tests under:

- `gpu-offload/controller/tests`
- `gpu-offload/runtime/tests`

Please consider adding a path-gated reusable workflow based on the existing `pytest-inference.yml` or `pytest-dm-tools.yml` patterns. Since each subproject has its own manifest and lock, it would be preferable to run them in separate frozen environments and publish separate coverage artifacts or Codecov flags.

A frozen controller test invocation currently fails because Hatch cannot infer the wheel contents. Could you clarify the intended packaging contract and either configure the build target or mark the project as script-only and test it with `--no-install-project`?

The new job should also be included in the PR and main validation summaries.

#### 2. CPU smoke coverage

Please consider extending `shared/ci/smoke-import.sh` and the matrix in `smoke-cpu.yml` with GPU-offload probes. These could install from the committed locks and verify the runtime and controller import surfaces without requiring GPU hardware.

For container validation, could the existing path-gated runtime-image smoke pattern be reused to build and probe the controller, runtime payload and first-run CPU images?

#### 3. CPU-only Kubernetes E2E

The PR already contains most of a suitable E2E scenario:

- kind cluster creation;
- controller and runtime image-build scripts;
- Helm installation;
- the first-run chart and deployment script;
- `check-execution.sh`, which verifies an actual remote call.

Please consider composing these assets into one path-gated workflow on a standard GitHub runner. Keeping the scenario logic in the existing scripts would avoid duplicating it in workflow YAML.

This CPU path could verify chart installation, webhook TLS/RBAC, admission mutation, generated server reconciliation, packaging, networking and remote execution. GPU allocation, CUDA execution, latency measurements and robot motion could remain separate hardware-backed qualification checks.

#### 4. Containerfile discovery

`container-scan.yml` uses `scripts/security/discover-base-images.sh`, which currently discovers only `*Dockerfile*`. Several new images use `Containerfile`, so their bases appear to be outside this control.

Could you please extend the existing discovery and change-trigger logic to include `Containerfile*`? If final-image scanning is desired, please consider reusing the current Trivy tooling against images produced by the CPU image-smoke job.

#### 5. Helm and `remote.yaml` validation

There does not appear to be a shared Helm-validation harness today. Please consider adding a small reusable script or job that runs `helm lint --strict` and renders the main, first-run, SO-101 and UR10e charts with representative CPU and GPU values.

For `remote.yaml`, could the production `validate_xavier_config()` implementation be reused in a parameterized test over every committed example? Cross-reference checks such as unique stage names and valid `remoteloc` targets could be added there without introducing a second schema implementation.

The existing YAML job runs `actionlint` against GitHub workflows only, so it does not cover Helm values or `remote.yaml`.

#### 6. SO-101 submodule integrity

The standard checkout does not initialize `gpu-offload/examples/so101-real-hardware/upstream`.

Please consider adding a targeted check that confirms `.lerobot-version` agrees with the committed gitlink, while initializing the submodule only for the SO-101 build job. `update_upstream.sh` could remain the single update mechanism.

A practical delivery order might be:

1. Unit tests and controller packaging.
2. Path gating and validation-summary wiring.
3. Containerfile discovery.
4. CPU import and image smoke.
5. Helm and `remote.yaml` validation.
6. CPU kind E2E.
7. SO-101 submodule validation.

Lock consistency, Ruff, ShellCheck, documentation checks, CodeQL and dependency controls already apply and can remain unchanged.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "gpu-offload/examples/so101-real-hardware/upstream"]
```
algattik: Could we avoid introducing the repository’s first submodule here? Initializing this checkout consumes about 403 MiB (roughly 318 MiB of Git history/metadata plus an 85 MiB worktree), while the container does not patch LeRobot’s source. The published `lerobot==0.6.1` package corresponds to this pinned tag and exposes the required `lerobot-record`, `lerobot-train`, `lerobot-eval`, and `lerobot-rollout` entry points and extras. Please use a small local `pyproject.toml` with a committed `uv.lock` that pins `lerobot[core_scripts,feetech,training,evaluation,smolvla,pusht]==0.6.1`; that preserves reproducible transitive dependencies and hashes without adding the submodule checkout.

dwhiteddsoft: Good catch — done in dce4e6b7. Removed the `upstream` submodule and `.gitmodules`; the SO-101 Dockerfile now installs `lerobot[core-scripts,feetech,training,evaluation,smolvla,pusht]==0.6.1` from PyPI via a local `pyproject.toml`/`uv.lock` instead. Confirmed the 0.6.1 wheel matches the pinned tag and exposes `lerobot-record`, `lerobot-train`, `lerobot-eval`, and `lerobot-rollout`. `scripts/update_upstream.sh` was replaced with `scripts/update_lerobot_version.sh`, which bumps the pin and re-locks instead of updating a submodule checkout.

dwhiteddsoft:
Thanks for the detailed breakdown — this is a lot of good infra to add. To keep this PR reviewable I've scoped it to the highest-value, lowest-risk pieces from your list and will treat the rest as follow-up work:

**Done in dce4e6b7 / 2b8b0b0b:**
- **#6 SO-101 submodule integrity**: moot now — the submodule is gone entirely (see the `.gitmodules` thread), replaced with a pinned `lerobot==0.6.1` PyPI dependency + `uv.lock`.
- **#1 Unit tests / controller packaging**: confirmed and fixed the Hatch packaging bug you flagged — `gpu-offload/controller` had no package directory matching its project name, so `uv sync` failed trying to build it as an editable wheel. Dropped the unused `[build-system]`/Hatchling section and marked it `[tool.uv] package = false` (it's a flat script, never installed as a wheel anywhere). Added `.github/workflows/pytest-gpu-offload.yml`, a reusable workflow matrixed over `gpu-offload/controller` and `gpu-offload/runtime`, each synced from its own lock in an isolated environment, path-gated in `pr-validation.yml` and unconditional in `main.yml`, wired into both validation summaries.

**Left for follow-up** (would meaningfully grow this PR and each involves its own design decisions worth reviewing separately):
- #2 CPU smoke coverage (`smoke-import.sh` + `smoke-cpu.yml` matrix, runtime-image smoke for controller/runtime/first-run images)
- #3 CPU-only kind E2E (webhook TLS/RBAC, admission mutation, reconciliation, remote call)
- #4 Containerfile discovery in `discover-base-images.sh`
- #5 Helm-lint/`remote.yaml`-validation harness

Coverage/Codecov integration for the new pytest job is also intentionally out of scope for now — neither subproject has coverage tooling wired up yet.

Happy to open a tracking issue for the remaining items if that's useful.

dwhiteddsoft:
Replying here since this predates my more detailed reply on your follow-up comment (#issuecomment-5726472922):

- Added a path-gated `Pytest GPU Offload` CI job (`.github/workflows/pytest-gpu-offload.yml`) that runs both `gpu-offload/controller/tests` and `gpu-offload/runtime/tests` in their own locked environments, wired into both the PR and main validation summaries. It's green on this PR now.
- While wiring it up, found and fixed a real bug you'd have hit: `gpu-offload/controller`'s `pyproject.toml` declared a Hatchling `[build-system]` but had no package directory matching its project name, so `uv sync` failed trying to build it as an editable wheel. Fixed by marking it `[tool.uv] package = false` (it's a flat script, never installed as a wheel anywhere).
- Corrected the PR description — it no longer claims the pytest suites are "relying on CI" without being run anywhere; both now run and pass locally and in CI.

Building/scanning the images, Helm lint/render, and a CPU-only smoke test are still open (tracked in my reply on the other comment) — those are more involved and I'd like to scope them as follow-up work rather than grow this PR further.

.github/workflows/pr-validation.yml
```diff
@@ -69,6 +70,7 @@ jobs:
             echo "dm_tools=$(match '^data-management/tools/')"
```
katriendg, line 71: **High:** This gate ignores behavior-affecting changes under the Helm charts, scripts, registry, examples, image definitions, and committed remote manifests. Those paths can currently merge without any GPU-offload validation. Please broaden the path filter and add required CPU-capable chart/config/image discovery plus kind admission/remote-call smoke coverage.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+                        {
+                            "name": "remote-server",
+                            "image": serverimage,
+                            "command": ["python3", "-m", "remoter.autoremote"],
+                            "readinessProbe": copy.deepcopy(READINESS_PROBE),
+                        }
+                    ],
+                },
+            },
+        },
```
katriendg, line 684: **High:** Namespace ConfigMap data controls the generated server image and environment while the generated Deployment inherits the source workload's service account, pull secrets, runtime class, and allowed mounts. A ConfigMap editor can therefore cause attacker-selected code to run with a trusted workload identity. Please move executable image/identity/mount policy to administrator-owned immutable configuration; namespace input should select only an opaque allowlisted profile.

gpu-offload/runtime/remoter/remoter.py
```diff
@@ -0,0 +1,2217 @@
...
+    else:
+        clientloc = None
+    argsn, kwargs = rehydrate_args(args, kwargs, remotedClasses, clientloc, callbackOnCacheAdd)
+    if func_name == "__init__":
+        if len(argsn) > 0 and type(argsn[0]) in remotedClasses:
+            # set the remoted class owner to True
+            argsn[0].rmtowner_rmt0bf = True
+            logger.debug(f"Setting remoted class with ID {argsn[0].uuid_rmt0bf} owner to True")
+    # Get the function object
+    func_obj = getfuncobjfromname(key, module_name, func_name, class_name)
```
katriendg, line 776: **Critical:** The server resolves and imports the caller-supplied `module_name`, `class_name`, and `func_name` without authenticating the peer or checking an immutable server-side allowlist. Any pod that can reach this listener can construct a valid call for an undeclared importable callable (for example `os.system`) and execute it inside the inference server. Please authenticate the connection, validate the complete decoded callable identity against controller-owned policy before importing it, and add negative tests for forged peers and undeclared callables.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+        core_api, apps_api, batch_api = load_kubernetes_clients(args.kubeconfig)
+    controller = XavierAdmissionController(core_api=core_api, apps_api=apps_api, batch_api=batch_api)
+    if not args.disable_reconcile:
+        runtime = ReconcileRuntime(controller)
+        # Run the initial cluster-wide sync and watch loops in the background so the
+        # webhook server below can start accepting requests immediately instead of
+        # only after the sync completes -- otherwise Kubernetes can route admission
+        # requests (or pass a readiness probe) to this Pod before anything is
+        # actually listening for them.
+        threading.Thread(target=runtime.start, name="reconcile-runtime", daemon=True).start()
```
katriendg, line 1271: **High:** A transient failure in the initial cluster-wide list terminates this detached worker permanently, but `/readyz` continues returning 200. The webhook can then keep mutating clients without ever creating their server Deployments. Track initial-sync and worker liveness, retry startup failures with backoff, and make readiness fail while required reconciliation workers are unavailable.

gpu-offload/helm/gpu-offload/templates/mutating-webhook.yaml
```diff
@@ -0,0 +1,323 @@
...
+          operator: In
+          values:
+            - "true"
+            - "True"
+            - "TRUE"
+            - "1"
+    sideEffects: None
+    admissionReviewVersions:
+      - v1
+    failurePolicy: Fail
```
katriendg, line 76: **High:** `failurePolicy: Fail` can block opted-in workload creation when the controller image, Service, or TLS setup is unhealthy. Before merge, please add and validate a production rollback path covering canary admission, Helm rollback/uninstall, emergency webhook removal, generated Deployment cleanup, and post-rollback admission verification.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+    uid = owner_metadata.get("uid")
+    if not uid:
+        return
+    deployment.setdefault("metadata", {})["ownerReferences"] = [
+        {
+            "apiVersion": _api_version_for_owner(owner_obj),
+            "kind": owner_obj["kind"],
+            "name": owner_metadata["name"],
+            "uid": uid,
+            "blockOwnerDeletion": True,
```
katriendg, line 543: **High:** `blockOwnerDeletion=True` requires the controller identity to have delete authorization for the owner. The chart does not grant delete for Pods or Jobs, so direct-Pod and per-client Job paths can fail to create their server Deployment. Normal owner-reference garbage collection still works without this flag.

```suggestion
            "blockOwnerDeletion": False,
```

Alternatively, grant and test delete authorization for every supported owner kind.

gpu-offload/registry/registry-common.sh
```diff
@@ -0,0 +1,110 @@
...
+registry_render_k3s_config() {
+  printf 'mirrors:\n'
+  printf '  "%s":\n    endpoint:\n      - "http://%s"\n' "$REGISTRY_HOST" "$REGISTRY_HOST"
+  local spec host port
+  for spec in "${REGISTRY_UPSTREAMS[@]}"; do
+    IFS='|' read -r host port _ _ _ <<< "$spec"
+    printf '  "%s":\n    endpoint:\n      - "http://localhost:%s"\n' "$host" "$port"
+  done
+  printf 'configs:\n'
+  printf '  "%s":\n    tls:\n      insecure_skip_verify: true\n' "$REGISTRY_HOST"
```
katriendg, line 73: **High:** This configures plain-HTTP, unauthenticated image distribution and disables TLS verification. Combined with mutable `:local` references and privileged robot workloads, any local process able to push a replacement manifest can gain device-level execution on restart. Please use authenticated TLS with a private CA, remove `insecure_skip_verify`, and deploy digest-pinned images after push.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+        if labels.get(XAVIER_PARENT_NAME_LABEL) != parent_name:
+            labels[XAVIER_PARENT_NAME_LABEL] = parent_name
+            changed = True
+
+    changed |= add_configmap_volume(spec, xaviercfg["remoteablecm"])
+
+    remoteable_containers = set(xaviercfg.get("remoteableconts", []) or [])
+    workload_name = metadata.get("name", "unknown")
+    perclient_label = getserverlabel(workload_name) if kind == "Pod" else None
+    server_label = getserverlabel(workload_name)
```
katriendg, line 480: **High:** This injects a discovery label from the unbounded workload name, while generated server labels use `_deployment_name()`, which truncates and hashes values over 63 characters. For long names the client selector can never match the ready server pod. Please derive both injected and generated labels through one canonical function and test exact selector equality for long shared and per-client names.

gpu-offload/controller/Containerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.12-slim
+
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+# Install Python deps only (no apt-get, no runtime installs). PIP_INDEX_URL is
+# passed as a BuildKit secret, not a build arg, so a credential embedded in it
+# never lands in image history.
+RUN --mount=type=secret,id=PIP_INDEX_URL,env=PIP_INDEX_URL \
```
katriendg: **Medium:** The production image installs a separate hand-maintained dependency set with `pip`, so it can drift from the committed `pyproject.toml`/`uv.lock` environment exercised by CI. Please build from the frozen lock using the repository's approved `uv` flow and remove duplicate dependency versions from the Containerfile.

gpu-offload/runtime/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+
+ADD . /
```
katriendg, line 3: **Medium:** This file and `runtime/Containerfile` are identical independent build definitions. That creates ambiguous discovery and requires every future change to remain duplicated. Please keep one canonical build file, or generate/validate the alias if both names are required by external tooling.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+    # meaningful signal at all.
+    metadata, spec, xaviercfg = get_metadata_spec(obj, strict=False)
+    is_opted_root = (
+        metadata is not None
+        and spec is not None
+        and xaviercfg is not None
+        and not _is_parent_labeled_pod(metadata)
+        and _is_fully_opted_root_workload(metadata, spec)
+    )
+    if is_opted_root:
```
katriendg, line 963: **Medium:** Cleanup only runs while the parent remains fully opted in. Removing the annotation, label, or `REMOTERPORT` makes `is_opted_root` false, leaving the previously generated Deployment and its GPU allocation alive until the parent is deleted. Please reconcile resources by owner UID even when current desired state is no longer opted in, with tests for each opt-out signal.

gpu-offload/mise.toml
```diff
@@ -0,0 +1,221 @@
...
+#
+# PIP_INDEX_URL = "https://packagefeedproxy.microsoft.io/pypi/simple/"
+# UV_INDEX_URL = "https://packagefeedproxy.microsoft.io/pypi/simple/"
+# npm_config_registry = "https://packagefeedproxy.microsoft.io/npm/"
+
+# ==============================================================================
+# Resolution
+# ==============================================================================
+
+[tasks."a-detect"]
```
katriendg, line 32: **Medium:** The hyphenated task families cannot be discovered or invoked through the repository's required semantic grouping patterns such as `mise run cluster:**`. Please rename these to colon-separated groups (for example `detect:platform`, `cluster:create`, and `offload:deploy`) and update dependencies and documentation atomically.

gpu-offload/runtime/remoter/rmtclass.py
```diff
@@ -0,0 +1,296 @@
...
+    allowallfunctions(cls, isserver)
+    remoter.addremotedclass(cls)
+    return cls
+
+
+def remotedclass(taskname=None, params={}):
+    def decorator(cls):
+        return createRemotedClass(cls, taskname, params)
+
+    return decorator
```
katriendg: **Medium:** The default dictionary is shared across decorator calls, and `createRemotedClass()` mutates it. Copy the caller's configuration per decoration so state cannot leak between classes.

```suggestion
def remotedclass(taskname=None, params=None):
    def decorator(cls):
        return createRemotedClass(cls, taskname, dict(params or {}))

    return decorator
```

gpu-offload/scripts/install-controller.sh
```diff
@@ -0,0 +1,23 @@
...
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+eval "$(scripts/detect-platform.sh --export)"
+
+# build-controller-image.sh always tags the image localhost/xavier-mutate:local,
+# and load-images.sh imports that exact archive/tag straight into the node's
+# containerd for every runtime (k3s and kind alike) -- it is never pushed to the
+# host-local registry under any other reference. Referencing it as
+# $registry_host/xavier-mutate:local would ask the cluster to pull an image that
+# was never published there, so this always uses the imported tag directly.
+helm --kube-context "$GPU_OFFLOAD_KUBE_CONTEXT" upgrade --install gpu-offload helm/gpu-offload \
```
katriendg, line 14: **Medium:** This cluster-mutating entry point bypasses the repository shell contract: shared `common.sh`, argument/help parsing, `--config-preview`, `require_tools`, and the deployment summary are absent. Please apply that contract here and to the other new setup scripts so operators retain the standard dry-run and configuration review path.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+    changed |= set_env_var_if_not_exists(container, "CONFIGFROMKUBE", "true")
+    changed |= set_env_var_if_not_exists(container, "SERVERLABEL", server_label)
+    changed |= set_env_var_if_not_exists(container, "XAVIER_CONTAINER", "true")
+    changed |= set_env_var_if_not_exists(container, "STAGE_NAME", "client")
+    if perclient_label is not None:
+        changed |= set_env_var_if_not_exists(container, "PERCLIENTSERVERLABEL", perclient_label)
+    return changed
+
+
+def DoMutate(obj: dict[str, Any], objOrig: dict[str, Any] | None = None, *, strict: bool = False) -> bool:
```
katriendg, line 439: **Medium:** This new package exposes camel-case and concatenated public APIs (`DoMutate`, `getserverlabel`, `getparam`, and related helpers), contrary to the repository's Python naming convention. Because this is a new API with no compatibility requirement, please rename these to snake_case and update exports, imports, and tests together.

gpu-offload/runtime/remoter/rmtconfigkube.py
```diff
@@ -0,0 +1,337 @@
...
+                if os.environ.get("NODE_NAME", None) and pod.spec and pod.spec.node_name == os.environ["NODE_NAME"]:
+                    # also check if REMOTERSOCK envvar exists on other pod and if it does use unix sockets
+                    socket_path = get_pod_socket_path(pod)
+                    if socket_path:
+                        loc = f"unix:{socket_path}"
+                if add:
+                    locations[serverlabel].add(loc)
+                    print("Added location for serverlabel", serverlabel, ":", loc)
+                elif remove:
+                    locations[serverlabel].discard(loc)
```
katriendg, line 111: **Medium:** A previously advertised endpoint is removed only on deletion. A `MODIFIED` event that changes the pod to NotReady, Failed, or a new IP leaves the stale location selectable until deletion. Please track entries by pod UID and remove or replace the prior endpoint on every non-ready transition and IP change.

gpu-offload/XAVIER-PORTING.md
```diff
@@ -0,0 +1,239 @@
...
+| `env`                | list    | Implemented | Environment variables      |
+| `noserverdeployment` | boolean | Implemented | Skip deployment creation   |
+| `remoteablecm`       | string  | Implemented | ConfigMap name             |
+| `remoteableconts`    | list    | Implemented | Container filter list      |
+
+These are documented in [remote-spec-schema.md](./specifications/remote-spec-schema.md)
+
+## Validation and Tests
+
+Current test coverage:
```
katriendg, line 203: **Low:** The “Next Steps” still request committed lockfiles, frozen-lock tests, and Podman scripts that this PR already adds. Please remove completed items and retain only genuinely deferred CI cross-build and end-to-end work, linked to durable follow-up issues. Also confirm whether this internal porting-history document belongs in the public product tree.

gpu-offload/controller/mutate.py
```diff
@@ -0,0 +1,1313 @@
...
+    def do_POST(self) -> None:
+        if urlsplit(self.path).path != "/mutate":
+            self._send_json(HTTPStatus.NOT_FOUND, {"error": "not found"})
+            return
+        try:
+            content_length = int(self.headers.get("Content-Length", "0"))
+        except ValueError:
+            self._send_json(HTTPStatus.BAD_REQUEST, {"error": "Content-Length must be an integer"})
+            return
+        body = self.rfile.read(content_length)
```
katriendg, line 1117: **Medium:** The webhook trusts an arbitrary non-negative `Content-Length` and performs an unbounded blocking read in an unbounded-thread server. A reachable pod can exhaust memory or retain handler threads; with fail-closed admission this becomes workload-creation denial of service. Reject missing/negative/oversized lengths before reading, cap reads to `limit + 1`, configure read timeouts and bounded concurrency, and test oversized, truncated, and slow bodies.

gpu-offload/controller/tests/test_mutate.py
```diff
@@ -0,0 +1,828 @@
...
+import json
+import os
+import threading
+import types
+
+import pytest
+from kubernetes import client
+
+
+def _load_mutate_module():
```
katriendg, line 16: **Medium:** This large new suite leaves helpers, fakes, and tests untyped and uses action-only test names and `os.path`, bypassing the repository's Python test conventions. Please add full annotations, use `pathlib.Path`, and group/rename tests by behavior (`test_given_<context>_when_<action>_then_<result>`) while the package is still new.

.github/workflows/pytest-gpu-offload.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Setup uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v10.0.1
+
+      - name: Install dependencies
+        working-directory: ${{ matrix.path }}
+        run: uv sync --locked --extra test
+
+      - name: Run pytest
+        working-directory: ${{ matrix.path }}
+        run: uv run pytest -o addopts="" tests -v
```
katriendg, line 44: **Medium:** These tests run without coverage collection or upload, and `codecov.yml` has no GPU-offload flags or paths. A green repository Codecov result therefore says nothing about the new controller/runtime domain. Please collect and upload separate controller/runtime reports and add matching Codecov components/statuses.

gpu-offload/scripts/preflight-check.sh
```diff
@@ -0,0 +1,19 @@
...
+kubectl version --client
+helm version
+# Podman builds the workload images on every platform. Its rootless and runtime
+# configuration only matters for kind, where Podman also hosts the node.
+podman --version
+if [ "$GPU_OFFLOAD_RUNTIME" = "kind" ]; then
+  kind version
+  podman info --format '{{.Host.Security.Rootless}} {{.Host.OCIRuntime.Name}}'
+fi
+helm template gpu-offload helm/gpu-offload \
```
katriendg, line 16: **Medium:** Portable validation and configuration generation remain embedded in Bash without fixture-based tests; this preflight renders only one chart. Please move portable parsing, discovery, policy checks, and generated configuration into a PowerShell module with Pester coverage for charts/values, every `remote.yaml`, build-file discovery, image digest policy, CI path triggers, Codecov, mise consistency, and registry output. Keep Linux host mutation in Bash. Pester should complement, not replace, kind/image/live-cluster smoke tests.

gpu-offload/runtime/pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+packages = ["remoter"]
+
+[tool.ruff]
+line-length = 120
+target-version = "py312"
+
+[tool.ruff.lint.per-file-ignores]
+# remoter is vendored from microsoft/xavier; remaining findings are upstream style
+# choices (unused locals, lambda assignment, == False) left unrefactored.
+"remoter/*.py" = ["E712", "E731", "F841"]
```
katriendg: **Medium:** This child project replaces rather than extends the root Ruff policy and then suppresses three rule families across the entire runtime package. Inherit the repository configuration and fix the reported instances; keep only narrowly justified line-level suppressions.

```suggestion
[tool.ruff]
extend = "../../pyproject.toml"
```

gpu-offload/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+For a zero-to-verification local walkthrough, start with [docs/README.md](./docs/README.md). It covers the CPU-only, WSL2 NVIDIA, and bare-metal NVIDIA paths, all built on the same [first-run example](./examples/first-run/README.md).
+
+1. Install the control plane:
+
+   ```bash
+   helm install gpu-offload ./helm/gpu-offload \
+     --set image.registry=<your-registry>.azurecr.io
+   ```
+
```
katriendg, line 35: **Medium:** The root quick start sets only the registry, while the chart defaults leave the controller tag and digest empty. A fresh registry therefore has no reproducible controller image to deploy. Please document the build/mirror prerequisite and require an explicit reviewed tag or, preferably, `mutate.image.digest=sha256:...`, followed by rollout verification.

gpu-offload/docs/TODO.md
```diff
@@ -0,0 +1,129 @@
...
+* [x] Start every task name with a letter, never a digit or symbol.
+* [x] Order the groups so the alphanumeric sort matches the human path: setup, deploy, verify, develop, then teardown.
+* [x] Update the READMEs and walkthroughs that reference the old task names.
+
+### Mise Task Acceptance Criteria
+
+* [x] `mise.toml` contains no multi-line shell bodies.
+* [x] `mise tasks` lists the groups in the order a contributor executes them.
+* [x] Each script is independently executable and passes `shellcheck`.
+* [ ] Every documented command matches a task that exists.
```
katriendg, line 35: **Medium:** This checked-in ledger contains release-relevant implementation and validation gaps, but several states contradict landed code/evidence and there is no durable owner or blocking disposition. Please reconcile completed/partial items and move unresolved work to owned issues with explicit acceptance criteria; `TODO.md` should not be the sole tracking system for known defects.

katriendg (commented):
Added inline review comments for the current head.

katriendg:
Thanks for the substantial GPU-offload integration and for addressing earlier review rounds by Alexandre. Also adding some inline comments. The added pytest workflow and submodule removal are useful completed responses. The proposed follow-up split should be narrowed: hardware-only GPU qualification can be tracked separately, but CPU smoke, chart/config validation, image build discovery, Codecov wiring, and a validated fail-closed rollback path belong in this PR because they validate or govern the newly introduced merge surface. TODO.md should not remain the sole durable owner of known implementation defects; unresolved release-relevant items need explicit issue ownership and blocking/non-blocking disposition. Open question is even if a file named `TODO.md` should be there? Leaving one critical and some High findings, I believe those are key to address before merge? And then for Medium findings letting you decide which ones to tackle or add to backlog issues to track.

The customer-facing documentation should also establish the value before exposing the implementation machinery. State plainly that this is an optional deployment profile for running policies that exceed a robot's onboard compute, power, thermal, or battery budget on a nearby site GPU while robot I/O and control remain local. Add a decision guide covering when to use onboard inference, this GPU-offload profile, a conventional inference service, or cloud inference, together with explicit non-goals for disconnected and hard real-time operation. Clarify the intended audience and prerequisites so users without Kubernetes operations experience do not mistake this specialized capability for the default inference path.

The naming also needs an explicit decision. The PR title and documentation use “Xavier,” but the implementation does not appear to require NVIDIA Jetson Xavier hardware; the term instead refers to another project), wonder if we keep that and if this will become clear for users. If “Xavier” remains in the title or user-facing terminology, confirm that this ambiguity is intentional and add a prominent clarification that it is not an NVIDIA Jetson Xavier requirement. Otherwise, prefer “GPU Offload” in the title and customer-facing navigation, and keep Xavier in provenance or implementation-history material.

gpu-offload/controller/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+name = "gpu-offload-controller"
+version = "0.0.0"
+description = "Hardened Xavier-style admission mutate controller (gpu-offload)"
+requires-python = ">=3.12"
+
+dependencies = [
```
katriendg, line 7: **High:** This is one of three new independent uv projects, but none of their directories is added to `.github/dependabot.yml`. The root uv entry does not maintain these nested manifests, so Dependabot cannot raise version or security-update PRs for the controller, runtime, or SO-101 lockfiles. Please add a uv configuration covering `/gpu-offload/controller`, `/gpu-offload/runtime`, and `/gpu-offload/examples/so101-real-hardware`, and add a check that every committed `uv.lock` directory has Dependabot ownership. **If this is not implemented in this pass, please create a backlog issue with an owner and acceptance criteria and link it from this PR.**

.github/workflows/dependency-review.yml
```diff
@@ -42,4 +43,5 @@ jobs:
             GHSA-7qhf-v65m-g5f3,
             GHSA-46r5-x6jq-v8g6,
             GHSA-fh64-r2vc-xvhr,
-            GHSA-h35f-9h28-mq5c
+            GHSA-h35f-9h28-mq5c,
+            GHSA-xrqw-3rrv-vx5w
```
katriendg, line 47: **High:** Adding this GHSA to the shared `allow-ghsas` list suppresses it repository-wide, not only for the SO-101 lock that currently cannot resolve the fixed transformers release. It can therefore hide the same advisory in another manifest and has no enforced expiry or automated removal path. Please isolate or path-scope this evidence-backed exception and attach an owner and removal condition for the first compatible fixed LeRobot/transformers resolution. **If this is not implemented in this pass, please create a backlog issue with an owner, expiry/removal criteria, and a link from this PR.**

gpu-offload/controller/Containerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.12-slim
```
katriendg: **Medium:** The new GitHub Actions are correctly full-SHA pinned, but the new OCI surfaces are not equivalently governed. This and other Python/CUDA/device-plugin references are tag-only; the pinning scanner explicitly excludes Dockerfile `FROM` lines, scans YAML images only in existing workflow/infrastructure paths, and base-image discovery searches `*Dockerfile*` rather than `Containerfile`. The required pinning check therefore passes while these references remain mutable and undiscovered. Please digest-pin the images and extend discovery, scanning, and update ownership across GPU-offload Dockerfiles, Containerfiles, shell-generated manifests, and Helm values. **If this is not implemented in this pass, please create a backlog issue with an owner and acceptance criteria and link it from this PR.**

katriendg (commented):
Added three follow-up inline comments on dependency update coverage and pinning enforcement.

.github/workflows/pr-validation.yml
dwhiteddsoft: Fixed in 8b97d4b6. Added a `gpu_offload_e2e` path filter covering the Helm chart, host scripts, local registry, and examples (including committed `remote.yaml` manifests and Containerfiles) — the paths this gate previously ignored.

Rather than build new coverage from scratch, this wires the existing `gpu-offload/examples/first-run` example into CI: it's a purpose-built example that already proves admission (webhook mutates the client, generates the server Deployment) and a real remote call (`check-execution.sh` asserts `executed_by` names the generated server pod, not the client) end to end on a CPU-only `kind` cluster. `detect-platform.sh` resolves to `platform=cpu` on any GPU-less runner, and every GPU-only step in the pipeline self-skips, so `mise run f-10-setup && mise run f-20-verify` was already a complete CI-safe smoke run — it just wasn't wired in. New workflow: `.github/workflows/gpu-offload-kind-smoke.yml`, gated in `pr-validation.yml` and required via `pr-validation-summary`.

gpu-offload/runtime/pyproject.toml
dwhiteddsoft: Done — `gpu-offload/runtime/pyproject.toml` now extends the root Ruff config as suggested. The originally-suppressed `E712`/`E731`/`F841` instances (7 total) are fixed directly rather than suppressed. Inheriting root's rule set surfaced ~75 additional findings in the vendored `remoter/` package (not covered by the old per-file-ignore); those are now narrowly scoped with per-line `# noqa: <CODE>` comments (added via `ruff --add-noqa`) instead of a blanket file-glob ignore. `ruff check`/`ruff format --check` pass clean and the runtime test suite (23 tests) passes. Commit: e1799b1f3e1865561e2bf55b4c7d50753ced3e73.

gpu-offload/controller/Containerfile
dwhiteddsoft: Fixed in 290f0eec, implemented in full rather than a backlog issue.

- Digest-pinned every tag-only image under `gpu-offload/`: the controller, first-run, ur10e-single, and so101 Python base images, and the CUDA base image in `cluster-gpu-check.sh`. `configure-k3s-nvidia.sh`'s device-plugin image now takes a paired `--plugin-digest` flag alongside `--plugin-version`, with a guard against silently pinning an overridden version to the old digest.
- Extended `Test-DependencyPinning.ps1` with a new `dockerfile-base-image` type (`Get-DockerfileFromViolations`), handling multi-stage aliases, `scratch`, and bare `${VAR}` bases; added `**/*.sh` to the existing `docker` type so the shell-generated manifests in `cluster-gpu-check.sh`/`configure-k3s-nvidia.sh` are scanned too. Both are wired into the actual CI default type list, not just the script's own default — confirmed OpenSSF Scorecard never runs on `pull_request`, so it wasn't actually covering this.
- Fixed `discover-base-images.sh`'s glob to also match `Containerfile` (it was `*Dockerfile*`-only), and its digest-extraction regex, which was truncating the ARG-templated so101 `FROM` line into a garbage partial match once pinned.
- Added `dependabot` `docker`-ecosystem entries for the gpu-offload directories so these pins get kept current.

Verified: 266/266 Pester tests pass (including new coverage for the added function), zero pinning violations under the new type or anywhere in `gpu-offload/`, and every digest independently re-resolved against its registry before committing.

gpu-offload/controller/mutate.py
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/controller/mutate.py
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/runtime/remoter/remoter.py
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/helm/gpu-offload/templates/mutating-webhook.yaml
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/scripts/preflight-check.sh
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/README.md
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/docs/TODO.md
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

gpu-offload/controller/pyproject.toml
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

.github/workflows/dependency-review.yml
dwhiteddsoft: This is us intaking a Microsoft research project. We will file this as an issue to be addressed but wanted to get this into the repo.

agreaves-ms (approved):

dwhiteddsoft:
@katriendg and @algattik , thank you for the feedback and the assistance in making this better. I have done my best to address all the issues raised. Some wil have to be put off for a review with the MS Research team. I will merge this and address the remainder of the issues in follow on work denoted [here](https://github.com/microsoft/physical-ai-toolchain/issues/1627)

